PR #60, opened 2014-12-27
Fix dns use-after-free bug

bassosimone:
Spotted with valgrind while developing the socks5 feature. See d931724c0's comment for bug analysis and 71e12f6d5's comment for a description of the fix.

### DIFF
--- a/src/protocols/dns.cpp
+++ b/src/protocols/dns.cpp
@@ -329,7 +329,7 @@ Request::Request(std::string query, std::string address,
     if (libevent == NULL) {
         libevent = IghtGlobalLibevent::get();
     }
-    cancelled = std::make_shared<bool>();
+    cancelled = SharedPointer<bool>(new bool());
     *cancelled = false;
     RequestImpl::issue(query, address, std::move(func),
                        dnsb, libevent, cancelled);
@@ -429,7 +429,7 @@ Resolver::request(std::string query, std::string address,
     // or because of an error, or because the resolver is destroyed (this
     // is guaranteed by the destructor's impl).
     //
-    auto cancelled = std::make_shared<bool>();
+    auto cancelled = SharedPointer<bool>(new bool());
     *cancelled = false;
     RequestImpl::issue(query, address, std::move(func),
                        get_evdns_base(), libevent, cancelled);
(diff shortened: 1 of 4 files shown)


src/protocols/dns.cpp
```diff
@@ -194,8 +194,6 @@ namespace dns {
  * This is the internal object thanks to which Request is movable. Of
  * course, RequestImpl is not movable, since its address is passed to
  * one of the many evnds delayed requests functions.
- *
- * \remark You should not use this class directly, use Request instead.
```
hellais: Why was this removed? Is it the case that now people should use this class directly?

I think that should not be the case.

hellais:
When I run `make check` on OSX I get the following error:

```
test/ooni/tcp_test.cpp:56:26: error: chosen constructor is explicit in copy-initialization
    TCPTest tcp_test("", {});
```

src/protocols/dns.cpp
bassosimone: I think that specific line was misleading because RequestImpl was in fact directly used by the Resolver. On second thought, do we need to document RequestImpl? It's an internal class only defined inside dns.cpp.

bassosimone:
If you run `make check` in the master branch, do you get the same error? At the moment I've no access to any MacOS machine, but maybe I can force libc++ and see whether the error pops up!

hellais:
After reading http://stackoverflow.com/questions/17264067/chosen-constructor-is-explicit-in-copy-initialization-error-with-clang-4-2 it seems like the correct way to fix this is to pass as argument to the TCPTest constructor an instance of `ight::common::Settings()` instead of `{}`.

I will apply this fix to all the places where it appears.

bassosimone:
Ok, thank you!

bassosimone:
Oh, and yes, the proposed fix makes sense according to me.

I have developed this branch using clang and GNU's C++ library, so I had not seen your issue previously, but there have been cases (in other branches) in which initializing `ight::common::Settings` with {} was not possible because it was ambiguous.

For example, this [occurred to me](https://github.com/TheTorProject/libight/blob/feature/socks5/test/protocols/http.cpp#L122) when I wanted to pass an empty `Settings` as the argument of `std::make_shared<T>()` where `T` was `http::Stream`.

hellais:
I submitted a PR for this in `fix/tcp_tests_osx`. Once that is merged I will merge also these other branches.

bassosimone:
#65 was merged, so you can now proceed!

hellais:
Ok merging. Thanks!

PR #61, opened 2015-01-03
Fix http stream corrupted stack

bassosimone:
This pull request fixes a 'corrupted stack' error in http::Stream that leads to a crash. The crash reason is quite stupid (once I realized it); http::Stream shall not be allocated on the stack because its `this` is captured by a few lambdas. Therefore, after http::Stream is moved, the captured `this` is no longer valid.

Fix by using SharedPointer and allocating http::Stream on the heap.


hellais:
Merging, thanks!

PR #62, opened 2015-01-03
Add SharedPointer<T> smart pointer

bassosimone:
This pull request adds the `SharedPointer<T>` template smart pointer. This extends the standard `std::shared_ptr<T>` to check whether the pointee is `nullptr` each time the pointer is accessed. The way in which `SharedPointer<T>` is implemented is such that you can assign it the result of the standard `std::make_shared<T>()` function template.

I have already used the `SharedPointer<T>` template to fix a couple of bugs in pull requests #60 and #61.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ test/common/bufferevent
 test/common/delayed_call
 test/common/evbuffer
 test/common/poller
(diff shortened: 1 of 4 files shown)


hellais:
Merging, thanks!

PR #63, opened 2015-01-04
Start to fix move semantic

bassosimone:
I was using move semantic in an error prone way. Hand-writing move constructor and move-assignment operator leads to errors when one forgets to update them *and* a new field is added (see #61 and in particular 78fc5d122ec and 5c221b47278).

Also, when there are lambdas bound to an object's `this`, either the object must be created on the heap or the lambdas must be recreated after each move (see again #61, in particular 097d1b5a3a8a and 72ea0117539). In this case the problem is that making an object movable encourages one to create it on the stack, which is exactly what should not happen.

Thinking a bit more on move semantic, I've come to the conclusion that  we should rely on the compiler's move semantic for simpler objects. As regards complex objects, we should make them non copyable and non movable; we should allocate such complex objects on the heap; and we should manage such complex objects lifecycle using smart pointers (see #62).

This pull requests starts making sure that move semantic is not used in an error-prone way and replaces its usage with safer and simpler patterns. It covers the files below the `src/common` folder. I have patches for other parts of the tree.

### DIFF
--- a/src/common/check_connectivity.hpp
+++ b/src/common/check_connectivity.hpp
@@ -12,11 +12,15 @@
 // Helper class to check whether we have connectivity
 //
 
+#include "common/constraints.hpp"
+
 // Forward declarations
 struct event_base;
 struct evdns_base;
 
 namespace ight {
+
+using namespace ight::common::constraints;
 
 /*!
  * \brief Class used to check whether the network is up.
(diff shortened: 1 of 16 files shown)


hellais:
Seems reasonable. Merging!

PR #64, opened 2015-01-08
Pass ight::common::Settings() instead of {} to TCPTest constructor.

hellais:


bassosimone:
The regress test failed with the following errors:

> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> lt-dns is a Catch v1.0 b53 host application.
> Run with -? for options
> -------------------------------------------------------------------------------
> 
> A specific custom resolver works as expected
> -------------------------------------------------------------------------------
> test/protocols/dns.cpp:1066
> ...............................................................................
> test/protocols/dns.cpp:1109: FAILED:
> REQUIRE( response.get_evdns_status() == 0 )
> with expansion:
> 2 == 0
>

The error above is `#define DNS_ERR_SERVERFAILED 2` which means `The name server was unable to process this query due to a problem with the name server`.

Could be this a transient error? I've not seen it before when running tests on Travis CI (and I have compiled and tested your branch successfully on Linux).

I'm going to restart the Travis CI build again (maybe more than once) to see what happens.

> [warn] event_base_loop: reentrant invocation. Only one event_base_loop can run on each event_base at once.
> -------------------------------------------------------------------------------
> If the resolver dies the requests are aborted
> -------------------------------------------------------------------------------
> test/protocols/dns.cpp:1142
> ...............................................................................
> test/protocols/dns.cpp:1142: FAILED:
> due to unexpected exception with message:
> event_base_dispatch() failed
> [warn] event_base_loop: reentrant invocation. Only one event_base_loop can run on each event_base at once.
> -------------------------------------------------------------------------------
> A request to a nonexistent server times out
> -------------------------------------------------------------------------------
> test/protocols/dns.cpp:1181
> ...............................................................................
> test/protocols/dns.cpp:1181: FAILED:
> due to unexpected exception with message:
> event_base_dispatch() failed
> [warn] event_base_loop: reentrant invocation. Only one event_base_loop can run on each event_base at once.
> -------------------------------------------------------------------------------
> It is safe to cancel requests in flight
> -------------------------------------------------------------------------------
> test/protocols/dns.cpp:1221
> ...............................................................................
> test/protocols/dns.cpp:1221: FAILED:
> due to unexpected exception with message:
> event_base_dispatch() failed

The errors above are caused by an exception terminating libevent's loop. Since libevent is not written in C++, it is not exception aware, therefore for libevent we're still in the event loop event thought we've exited the expedite way using an exception. I guess the libight function that runs the event loop should do something such that, when we exit from the event loop due to an exception, we can later reeenter it. This is not useful in real, but helps with tests.

bassosimone:
I'm going to call the above a transient failure and merge this pull request.

Repeated tests on my machine as well as on Travis CI not reported again the error seen above. I've also added a ticket (#65) to describe the issue that one cannot call again the event loop after an exception terminated a previous event loop.

Thanks, @hellais, for fixing the compilation error on MacOS!

PR #66, opened 2015-01-18
Fix dns move semantic

bassosimone:
Remove hand-rolled move semantic. Use compiler-provided copy and move implementation for objects that don't contain pointers to C structures and/or lambdas. Otherwise, disable copy and move and, when copy and move are needed, allocate on the heap using SharedPointer.


hellais:
Ok, looks good. Thanks!

PR #67, opened 2015-01-18
Fix net move semantic

bassosimone:
This fixes the move semantic in `src/net`. If an object does not contain pointers to C structures allocated on the help and does not contain lambdas bound to this and does not contain callbacks, make it copyable and movable. Otherwise, make it non copyable and non movable and, when we need to copy and/or move it, allocate it on the heap and manage its lifecycle using a SharedPointer.


hellais:
After the conversation with @bassosimone I better understand what is going on here. This is reasonable. Merging!

PR #68, opened 2015-01-18
Simplify connection by removing its delayed close

bassosimone:
Before implementing socks5, I dedicated some love to connection to simplify it a bit. In perspective the changes in this pull request should allow to avoid the complete wrapping of IghtConnectionState using IghtConnection.


hellais:
Cool merging.

PR #69, opened 2015-01-18
Connection: no separately managed socket

bassosimone:
Do not manage the socket separately. I've checked that libevent is robust with respect to calling close(-1) when the filedescriptor is not set for a long time now.


bassosimone:
(Close by mistake, sorry)

PR #70, opened 2015-01-18
Make connection's API more C++-ish

bassosimone:
This pull request increases the C++-ish-ness of connection. This is functional to next pull request, that will introduce socks5 and will make connection a class derived from a generic transport.


hellais:
This looks good.

PR #71, opened 2015-01-18
Implement socks5

bassosimone:
This is done by introducing Transport, a new abstraction, from which both socks5 and the old connection derive. Then, when you create a transport, you can specify whether to use socks5 or not. Depending on your choice, the connection code or the socks5 code is picked up.


hellais:
I am going to merge this with the reserve that it should not by default set the SOCKS ip and port to be Tor.

Anyways awesome work! :+1:

PR #73, opened 2015-01-18
Refine SOCKS5 and Tor implementation

bassosimone:
- Do not automatically use Tor's socks5
- Automatically use Tor instead when httpo://... URL are used

### DIFF
--- a/src/protocols/http.hpp
+++ b/src/protocols/http.hpp
@@ -558,9 +558,16 @@ public:
         // Extend settings with address and port to connect to
         settings["port"] = serializer.port;
         settings["address"] = serializer.address;
-        if (serializer.schema == "httpo" && settings.find(
-                "socks5_proxy") == settings.end()) {
-            settings["socks5_proxy"] = "127.0.0.1:9050";  // over Tor
+        // If needed, extend settings with socks5 proxy info
+        if (serializer.schema == "httpo") {
+            // tor_socks_port takes precedence because it's more specific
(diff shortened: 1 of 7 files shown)


src/protocols/http.hpp
```diff
@@ -558,6 +558,9 @@ class Request : public NonCopyable, public NonMovable {
         // Extend settings with address and port to connect to
         settings["port"] = serializer.port;
         settings["address"] = serializer.address;
+        if (serializer.schema == "httpo") {
+            settings["socks5_proxy"] = "127.0.0.1:9050";  // over Tor
+        }
```
hellais: This should not override the proxy settings that are specified by the user. That is it should be possible to specify an alternate Tor SOCKS port and make sure such setting has priority over defaulting to 9050.

bassosimone: Agreed, see fd17a53b6...

bassosimone:
Hi @hellais,

I've created a setting called `tor_socks_port` that takes precedence over `socks5_proxy` and is used only when `httpo://...` URLs are used.

While there, I've also changed the code not to modify the settings object, but to make changes on a private copy of it, because it's surprising that you pass a settings object to a function and after that function returns the object is unchanged.

What do you think?

bassosimone:
Ok, with the previous commit I've improved the regress tests so that they do not fail and it's obvious that the behavior with respect to `tor_socks_port` and `socks5_proxy` is correct.

hellais:
Beautiful! Thanks!

PR #74, opened 2015-01-22
Add script by @alemela for building libight on iOS

hellais:


bassosimone:
Merged, thanks! I'm working to polish this script a bit, which will go in as another pull request.

PR #90, opened 2015-02-22
Make embedders job simple

bassosimone:
This pull request teaches the build system to install yaml-cpp and boost headers when libight uses the builtin yaml-cpp. It also modifies the build system to compile a shared yaml-cpp library rather than embedding such library into libight. This simplifies the job of people who wants to embed libight into an application. Assuming that no dependencies are installed, one can run `./configure; make; make install DESTDIR=/foo/bar` to install at `/foo/bar` the libraries and headers of libight and of all its dependencies. Then, one can just import `/foo/bar` from his/her build system.


bassosimone:
Closed because I rebased the pull request on top of other branches (see #103)

PR #91, opened 2015-02-26
Make logging more flexible

bassosimone:
This pull request (constructed on top of #89) makes the logger more flexible. Now, you can specify a `std::function<>` that receives each line to be logged.

Example usage taken from regress tests:

    std::string buffer;

    ight_set_logger([&buffer](const char *s) {
        buffer += s;
        buffer += "\n";
    });

    ight_set_verbose(1);

    ight_debug("Antani");
    ight_info("Foo");

    REQUIRE(buffer == "Antani\nFoo\n");


hellais:
Yes this iOS good :+1: :P

Issue #42, opened 2014-11-12
Use the ight namespace rather than the Ight prefix


bassosimone:
In C++ code we should use the `ight` namespace rather than the `Ight` prefix.

Firstly implemented in #41 and +1-ed by @hellais during such pull request's review process.

I wonder what should we do with C code.

bassosimone:
Speaking of C code, we have little if not C code left in libight. However, there are some C-style functions that are implemented in C++ (e.g., `ight_debug()`). For such functions, I'm following the convention that they use the `ight_` prefix.

hellais:
Yes you are right. I would suggest we start making the log type functions be a bit more c++ like by putting them in the `ight::log` namespace.

bassosimone:
Ok, I like the idea, @hellais . I was just bikeshedding^H^H^H wondering whether it should be `ight::common::log::debug()` or `ight::log::debug()` as you say.

In fact in most cases the namespace is the path to the file containing that piece of code, e.g., `ight::net::buffer::Buffer` is a class in `ight/net/buffer.hpp`.

So, for consistency the function should be `ight::common::log::debug()` but I'd rather have `ight::log::debug()` for simplicity of usage.

Taking this a little further, we should also have `ight::loop()` rather than `ight::common::poller::loop()` for the function that runs the event loop.

I propose that we favor simplicity and adopt short prefixes for functions expected to be used often. In particular, I'd like to have `ight::loop()` and `ight::log::debug()`. As a second best, we can have a `GlobalPoller` class, as we already have, and a `GlobalLogger` class, which we don't have. But I'd rather use convenience function when possible.

What do you think?

My current proposal is that we adopt short prefixes for functions that are expected to be used often. In particular, I'd like to have `ight::loop()` and `ight::log::debug()`.

bassosimone:
This has been addressed in #142.

PR #92, opened 2015-03-13
Fix namespaces usage

bassosimone:
Make sure that we always use the ight namespace rather than having names using the Ight prefix. This pull request addresses issue #42 (sadly, even though issue #42 is sorted out, we're still far from knowing the answer to the ultimate question of life, the universe and everything).


hellais:
Poetry :book:

PR #93, opened 2015-03-17
Increase libight headers consistency

bassosimone:
This pull request renames `foo.h` headers to `foo.hpp` and makes sure that namespaces are always used consistently, that all headers have a brief copyright statement, etc.

In a couple of cases, I've done this transformation: `void foo()` => `void ight_foo()`, which could possibly have been `void foo()` => `void ight::foo()` instead as discussed in #42. I plan to deal with applying the latter transformation with a subsequent pull request once we finish the discussion in #42.


bassosimone:

PR #94, opened 2015-03-17
Remove unused build ios.sh

bassosimone:
I'm calling this script unused because I've recently created [a repository dedicated to the task of compiling libight for iOS](https://github.com/bassosimone/libight-build-ios) that includes an improved version of this script.

This pull request lays on the shoulders of #93. To best see what was changed, therefore, one should only consider [the difference between it and #93](https://github.com/TheTorProject/libight/compare/08bf27e...277d342).


bassosimone:

PR #95, opened 2015-03-17
Reindent files that were using spaces

bassosimone:
I picked up a popular C++ formatting style and tweaked it to look like the style used across libight (i.e., four spaces to indent). Then, I throwed all the files using tabs at `clang-format`.

Since this pull request is based on #94, it's best to review it [in comparison to #94 alone and without considering space changes](https://github.com/TheTorProject/libight/compare/277d342...7c11ed9?w=1) rather than in comparison to master.


bassosimone:

PR #96, opened 2015-03-17
Fix use-after-free error in http::Stream

bassosimone:
Since this pull request is based on #95, it's best to [review the difference between it and #95](https://github.com/TheTorProject/libight/compare/7c11ed9...6af9ad2).


bassosimone:

PR #103, opened 2015-03-20
Make embedders job simple (rebased)

bassosimone:
This is "same as" #90 except that it has been forward ported on top of other branches to allow to prepare a pull request containing code to run tests in another thread, which in turn is useful to develop the iOS App.

This pull request is "on top of" #96 and this is [the comparison with respect to that](https://github.com/TheTorProject/libight/compare/6af9ad2...b03d0c5).


bassosimone:

PR #104, opened 2015-03-20
Start tests asynchronously in a background thread

bassosimone:
Hi @hellais and @lorenzoPrimi,

in this branch I tried to implement the feature that starts tests asynchronously in a background thread directly into libight, as suggested [in libight_ios pull request #2](https://github.com/TheTorProject/libight_iOS/pull/2).

I'm not sure the code in here is good to be merged also because it contains some memory leaks, but I think it's a good starting point to discuss what I have tried to do.

Note that for convenience I have also created [a branch pointing to this branch in the repository that I typically use to build libight on iOS](https://github.com/bassosimone/libight-build-ios/pull/2) so that you can rebuild a libight for iOS including the code in this pull request. Unfortunately, from there I cannot recompile libight for iOS because I need to install too much stuff on this MacOS and have a crappy Internet connection at the moment.

What do you think of the code in here? (This pull request is based on #103 and [here's a useful link to only see what changed with respect to #103](https://github.com/TheTorProject/libight/compare/b03d0c5...618f2a4).)


src/common/async.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+void Async::break_loop() {
+    //state->poller->break_loop();  // Idempotent
+    ight_break_loop();
+    state->interrupted = true;
+}
+
+void Async::restart_loop() {
+    state->interrupted = true;
+}
```
bassosimone: By the way, this function's implementation is clearly wrong. It doesn't matter as far as using this code is concerned, but it should be fixed before this pull request is ready to merge into master.

bassosimone:

PR #105, opened 2015-03-20
Start tests asynchronously in a background thread (wrt master)

bassosimone:
This is same as #104 except that this pull request is based on master rather than on #103:

> Hi @hellais and @lorenzoPrimi,
>
> in this branch I tried to implement the feature that starts tests asynchronously in a background thread directly into libight, as suggested [in libight_ios pull request #2](https://github.com/TheTorProject/libight_iOS/pull/2).
>
> I'm not sure the code in here is good to be merged also because it contains some memory leaks, but I think it's a good starting point to discuss what I have tried to do.
>
> Note that for convenience I have also created [a branch pointing to this branch in the repository that I typically use to build libight on iOS](https://github.com/bassosimone/libight-build-ios/pull/2) so that you can rebuild a libight for iOS including the code in this pull request. Unfortunately, from there I cannot recompile libight for iOS because I need to install too much stuff on this MacOS and have a crappy Internet connection at the moment.
>
> What do you think of the code in here? (This pull request is based on #103 and [here's a useful link to only see what changed with respect to #103](https://github.com/TheTorProject/libight/compare/b03d0c5...618f2a4).)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ stamp-h1
 .DS_Store
 
 # Regression tests
+test/common/async
 test/common/bufferevent
 test/common/delayed_call
 test/common/evbuffer
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ before_install:
   - sudo apt-get install -qq gcc-4.8 g++-4.8
   - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 90
 install: autoreconf -i
-script: ./configure && make && make check
+script: ./configure && make && make check-am
 compiler:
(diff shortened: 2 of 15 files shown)


bassosimone:
Hmm, there was a segfault when this branch was tested by travis-ci. I'm closing the pull request to investigate further.

bassosimone:
I've run travis some more times and experienced a couple of segfaults. My hypothesis is that this set of patches cannot live without the use after free fix for http::Stream. In fact, I've not experienced travis-ci (or local) crashes when this code is compiled after the http::Stream user after free fix. For this reason I've prepared a pull request that includes both, #106.

PR #106, opened 2015-03-21
Patches needed to move forward iOS application development

bassosimone:
This pull request contains the patches needed to move forward iOS application development:

1) bugfix of an use after free bug without which the code to run tests in a background thread (see 3) experienced occasional crashes

2) code needed by the libight-build-ios repository to generate a directory containing libight and all its dependencies libraries and headers without which we cannot recompile a libight for iOS including the code described in 3

3) code to run tests in a background thread that is needed to move forward the iOS app development

Regarding 3, it still needs some love (for example, there are leaks that I am investigating) but before making it perfect I'd say let's merge it and use it in the App to see whether it fullfills our needs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ stamp-h1
 .DS_Store
 
 # Regression tests
+test/common/async
 test/common/bufferevent
 test/common/delayed_call
 test/common/evbuffer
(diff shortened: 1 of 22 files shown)


hellais:
This looks very good. I have also tested it on OSX and it works well. I am not going to test the building of the iOS library and seeing if it integrates well, but from the looks of the code it will greatly facilitate the job of an iOS developer.

My only concern is if the C++ threading library will work well on iOS (vs using the objective-c one), but we shall see.

bassosimone:
Thanks!

PR #109, opened 2015-03-26
More patches for libight iOS integration

bassosimone:
This pull request allows one to redirect the log messages generated by a test to a test-specific function and it also contains small tweaks and fixes to the code to run tests in a background thread.

All functions invoked by tests have been converted to the default logger, except the functions that live inside `src/common/utils.cpp` for which some extra refactoring is also needed.

My intent is to convert also this function the next time I'll commit some time to libight. If, at that point, this pull request is merged, I'll open a new one, otherwise I'll push my commit on top of this one.


src/common/async.cpp
```diff
@@ -42,20 +63,23 @@ struct AsyncState {
...
-    ight_debug("loop thread entered");
+
+    EvThreadSingleton::ensure();
+
+    ight_debug("async: loop thread entered");
     for (;;) {
 
         LOCKED(
-            ight_debug("loop thread locked");
+            ight_debug("async: loop thread locked");
```
hellais: Should you also not be using logger->debug() here?

test/common/async.cpp
```diff
@@ -17,20 +17,23 @@
...
 
+using namespace ight::ooni::dns_injection;
 using namespace ight::ooni::http_invalid_request_line;
+using namespace ight::ooni::tcp_connect;
 
 TEST_CASE("The async engine works as expected") {
-    evthread_use_pthreads();
-    ight_set_verbose(1);
+
+    //ight_set_verbose(1);
```
hellais: Why is this commented out?

test/common/async.cpp
```diff
@@ -60,6 +67,36 @@ TEST_CASE("The async engine works as expected") {
...
+        auto test = SharedPointer<DNSInjection>(
+            new DNSInjection("test/fixtures/hosts.txt", Settings{
+                {"nameserver", "8.8.8.8:53"},
+            })
+        );
+        test->set_log_verbose(1);
+        test->set_log_function([](const char *s) {
+            (void) fprintf(stderr, "test #3: %s\n", s);
+        });
+        ight_debug("test created: %llu", test->identifier());
```
hellais: Also here you should be calling log->debug() I believe

hellais:
Would it be possible to also add to the unittest an example on how to use the custom logger to obtain the log output only for the specific test instance fired?

Perhaps there should be a unittest that fires multiple nettests and it check to make sure it is able to obtain ONLY the log line for one specific instance.

src/common/async.cpp
bassosimone: In a certain sense, that's what this line of code does: `ight_debug()` is now syntactic sugar for DefaultLogger::get()->debug().

Here in async.cpp I am using the default logger because async code is not bound to any specific test (especially there in the thread that runs in the background).

test/common/async.cpp
bassosimone: Yes, it should be removed. The reason to have this line of code is silly: it makes the default logger verbose, which is helpful when you want to see the noise made by the code in the backfround thread, which -as explained above- uses the default logger.

test/common/async.cpp
bassosimone: True.

bassosimone:
Thanks for the review, @hellais! I will add all the nits into a single commit which I will push on top of the pull request.

bassosimone:
Speaking of redirecting the log messages of a single specific test to somewhere, I guess the following code (taken from [test/common/async.cpp line 52](https://github.com/TheTorProject/libight/blob/3a3074f9226a5e25d4f6e816b7e1b697631f60fc/test/common/async.cpp#L52)) could be the blueprint of this:

```C++
        auto test = SharedPointer<HTTPInvalidRequestLine>(
            new HTTPInvalidRequestLine(Settings{
                {"backend", "http://nexa.polito.it/"},
            })
        );
        test->set_log_verbose(1);
        test->set_log_function([](const char *s) {
            //(void) fprintf(stderr, "test #1: %s\n", s);
            // code to handle log message of this specific test here
        });
        ight_debug("test created: %llu", test->identifier());
        async.run_test(test);
```

I know that code of different tests go to different loggers because I see them prefixed with different `test #N` in the output of `test/common/async.cpp` but I agree that this is not a robust regress test in the long run.

So, I guess I'll either modify this test or add another test that saves the logs of each experiment to different vectors, verify that those vectors are nonempty, and commit this on top of the current pull request as well!

hellais:
Ok awesome great!

And don't forget to rebase with the current master as this branch does not merge cleanly.

hellais:
After speaking with @bassosimone and @lorenzoprimi we concluded to merge this and defer the improvements discussed in this PR to future work.

PR #110, opened 2015-03-30
Fix crash when socks5 code is compiled using gcc

bassosimone:
This pull request fixes a crash caused by used after free in socks5 code when the code was compiled using gcc. Interestingly, everything was OK when the code was compiled with clang. Detailed discussion and plan to mitigate this issue in 8c992d7 commit message.

In addition to this, I've also commited changes that, if applied earlier, would have allowed us to spot this bug. Namely, now we use the default compiler on Linux (as opposed on clang) which leads to use gcc in most cases. Now, we install Tor on travis-ci so that we test socks5 code also on travis-ci with both gcc and clang on Linux.

Finally, while testing all of this I noticed that an HTTP test took more than one minute to complete and I modified it to take less than four seconds.


hellais:
Good. Merging.

PR #114, opened 2015-04-16
Android-NDK-integration changes

bassosimone:
With the following changes merged, I am able to cross-compile libight for Android for ARM, MIPS and android-86 (see [libight-build-android](https://github.com/TheTorProject/libight-build-android)).

After this change, **remember to run** `git submodule sync` because I have changed the URL of boost/core repository to point to our personal fork.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,74 +1,74 @@
 [core]
-  ignore = dirty
+	ignore = dirty
 [submodule "src/ext/http-parser"]
 	path = src/ext/http-parser
 	url = https://github.com/joyent/http-parser
-  ignore = dirty
+	ignore = dirty
 [submodule "src/ext/libevent"]
 	path = src/ext/libevent
 	url = https://github.com/bassosimone/libevent
-  ignore = dirty
+	ignore = dirty
 [submodule "src/ext/Catch"]
 	path = src/ext/Catch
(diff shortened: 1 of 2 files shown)


bassosimone:
Extra info on this pull request: my fix for `boost/core` is to disable boost's demangling functionality for every platform because it is not available when cross compiling on Android for MIPS and x86. This is to say that the fix could be more precise.

However, since boost is only used by yaml-cpp and since yaml-cpp seems to work as before, I'm not sure we need to be more precise (unless we want to submit a pull request to `boost/core`).

What do you think @hellais?

hellais:
I thinking trying to get an acceptable patch to make boost/core work for Android MIPS and x86 cross compilation is ideal.

I will soon take a look at the code to better understand the matter.

hellais:
Ok let's go for the fork approach for the time being, but let's try to see if we can get a patch into boost core.

bassosimone:
Ok, @hellais. I've done some further work to [properly fix this boost core issue](https://github.com/bassosimone/libight-boost-core/commit/574d96f61af1c06dd087340c256c74fc2899ce88) and created [a minimal test case repository](https://github.com/bassosimone/boost-core-gabicxx-fix) that show the issue and demonstrates that the proposed fix work. What do you think of this solution?

Issue #49, opened 2014-12-09
Implement TCP Connect test based on libight


hellais:
This involves implementing the following test based off of libight: https://gitweb.torproject.org/ooni-probe.git/tree/ooni/nettests/blocking/tcp_connect.py
https://gitweb.torproject.org/ooni/spec.git/tree/test-specs/ts-008-tcpconnect.md

with support for reporting to local file in YAML format.

PR #115, opened 2015-04-21
Use the system-wide DNS resolver for the connectivity check

alessandro40:
As @bassosimone suggested, the Google public DNS might be blocked
in some countries. So, it is more robust to use the default system
resolver to query for the IP address of nexa.polito.it.

### DIFF
--- a/src/common/check_connectivity.cpp
+++ b/src/common/check_connectivity.cpp
@@ -65,10 +65,10 @@ Network::Network(void)
         throw std::bad_alloc();
     }
 
-    if (evdns_base_resolve_ipv4(dnsbase, "nexa.polito.it", DNS_QUERY_NO_SEARCH,
+    if (evdns_base_resolve_ipv4(dnsbase, "ebay.com", DNS_QUERY_NO_SEARCH,
                                 dns_callback, this) == NULL) {
         cleanup();
-        throw std::runtime_error("cannot resolve 'nexa.polito.it'");
+        throw std::runtime_error("cannot resolve 'ebay.com'");
     }
 
     if (event_base_dispatch(evbase) != 0) {
--- a/src/common/check_connectivity.cpp
+++ b/src/common/check_connectivity.cpp
@@ -41,12 +41,20 @@ Network::dns_callback(int result, char type, int count, int ttl,
     (void) ttl;
     (void) addresses;
 
-    //
-    // TODO: The following check is good for the unit test but, to be
-    // precise, there are other error conditions that can indicate that
-    // we have connectivity (e.g., DNS_ERR_NOTEXIST).
-    //
-    that->is_up = (result == DNS_ERR_NONE);
+    switch (result) {
+    case DNS_ERR_NONE:
+    case DNS_ERR_FORMAT:
+    case DNS_ERR_SERVERFAILED:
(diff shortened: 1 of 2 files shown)


hellais:
I don't think it's a good idea to always query the same domain though.

Perhaps we could query for a non existent domain and expect to get back a NXDOMAIN.

alessandro40:
The problem of the latter solution is that [some ISPs answer to queries for non existent domain with valid IP addresses](http://en.wikipedia.org/wiki/DNS_hijacking#Manipulation_by_ISPs), to display their own redirect page. So, the test query could receive either NXDOMAIN or a valid response.

hellais:
@alessandro40 that is true, but it would still be an indication of the network being available.

NXDOMAIN or a valid response means you have internet, though it may be a captive portal internet.

If you don't want to go this route, I would still advise to use a service that a lot of people access and not the nexa website since that is a very simple fingerprint to match against to spot users.

Some good candidates are:

* analytics|adsense.google.com

* cloudflare.com

* amazon.com

* ebay.com

bassosimone:
We could relax [the check on line #49](https://github.com/TheTorProject/libight/blob/master/src/common/check_connectivity.cpp#L49) to have a better definition of "we have networking" that also includes several errors returned by the name server. In dns.cpp I've already listed [all the error values that indicate that a DNS server replied to us](https://github.com/TheTorProject/libight/blob/master/src/protocols/dns.cpp#L35).

Speaking of captive portals, the `check_connectivity.cpp` module is currently used by regress tests requiring Internet access, not just the degree of "connectivity" provided by a captive portal. Yet, I don't think this is a big problem :-).

Regarding resolving nonexistent domains, I assume @hellais proposes to resolve domains at random, otherwise the domain would always be the same albeit not existent and I don't see the different with respect to nexa.polito.it. This leads me to ask whether resolving random domains is something that could qualify as suspicious user behavior.

alessandro40:
Considering @bassosimone and @hellais' observations, I think the best option is to use of of the popular domains proposed, but considering that analytics|adsense.google.com [seems to be blocked in China](https://en.greatfire.org/search/blocked).

alessandro40:
Furthermore, since we are not interested if the specific domain can be resolved, I agree on the fact that we can accept as a positive answer any reply from the DNS server.

src/common/check_connectivity.cpp
```diff
@@ -60,19 +68,15 @@ Network::Network(void)
         throw std::bad_alloc();
     }
 
-    if ((dnsbase = evdns_base_new(evbase, 0)) == NULL) {
+    if ((dnsbase = evdns_base_new(evbase, 1)) == NULL) {
```
bassosimone: Why here did you change 0 to 1?

alessandro40: According to the libevent docs, the second argument of ```evdns_base_new()``` must be 1 to use the system DNS resolver.

bassosimone: Ah, yes, sorry. This is correct and was the original objective of the PR. I'll merge this as soon as I can test it, hopefully by end of week.

bassosimone:
Thanks, merged!

PR #117, opened 2015-04-28
Implement Android traceroute

bassosimone:
This pull request implements traceroute on Android. The code is derived from the same purpose code used by the [Portolan](http://portolan.iet.unipi.it/) project developed at University of Pisa

I started to work to merge this code during Tor winter dev meeting. I finished tonight, after two days spent in Pisa working to this pull request with @ValerioLuconi.

Tested on Linux Ubuntu 14.10 as well as on Travis-ci. A previous version of this code compiled on Android on March, 5 2015 (and I don't see obvious reasons why this code should not). A nearly-finished version of this code was tested with IPv6.

PR #118, opened 2015-05-21
Add android traceroute code

bassosimone:
This is same as #117 except that I rebased the branch and squashed commits. I'd rather used #117 but apparently I cannot easily reopen it since I rebased.

---

This code was contributed by developers of Portolan, a network measurement tool developed at University of Pisa.

Many many thanks!

I started merging this code to libight during the Tor winter dev meeting in Valencia, after which I was able to compile it.

Then I worked on this together with @ValerioLuconi during a two day hackfest held in Pisa (27-28 Apr).

Finally I also worked on this on May 21, at Nexa.

This code compiles and runs on Linux, an early version of it was tested it both using IPv6 (at CNR), this version only on IPv4.

The original history of this contribution to libight counted 60+ commits that I have squashed to tidy things up.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,3 @@
 ---
 BasedOnStyle: LLVM
-IndentWidth: 4
 ---
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,7 @@ test/ooni/tcp_test
 test/protocols/dns
 test/protocols/http
 test/report/file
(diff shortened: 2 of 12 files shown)


hellais:
This should be all good and since it's very modular and compartimentalised it should not pose a risk to the overall stability of the library.

Merging :+1:

PR #132, opened 2015-06-15
Fix building on MacOSX

bassosimone:
Add one missing header required by MacOS and use defines that apparently are defined by both MacOS and Linux (saying this since compiling succeeded on both systems).


bassosimone:
Compiles both locally and on travis-ci, going ahead with merge.

PR #133, opened 2015-06-18
async: adjust tests to expose bug, then fix it

bassosimone:
This pull request adjusts async tests to expose a bug (leading to a crash) spotted while working on the iOS application and then fixes the bug (basically, the issue was that I overwrote the thread object before either detaching or joining it).


hellais:
This look like a hotfix. I think we can merge this safely.

bassosimone:
"I found it hard; it's hard to find" :+1:

hellais:
Oh well, whatever, nevermind

Issue #138, opened 2015-06-23
Memory leak in async.cpp


bassosimone:
[Moved from measurement-kit/measurement-kit-project-management#8]

Preparing measurement-kit/measurement-kit#133, I noticed this memory leak that occurs in async.cpp code (this run in particular sits on top of measurement-kit/measurement-kit#133).

Since there are no leaks when running individual OONI tests and since this is a memory leak and since we use many (perhaps too many) shared pointers, I am inclined to think that this is circular reference issue. Anyway, this is what Valgrind says about it:

```text
test #1: http: BODY
test #1: http: received body chunk...
test #1: http: BODY
test #1: http: received body chunk...
test #1: http: END
test #1: http: we have reached end of response
test #1: net_test: tearing down
test #1: net_test: written entry
test #1: net_test: reached end of input
async: test stopped
test complete: 120671888
async: bottom of loop thread
async: loop thread locked
async: empty
async: thread detached
all tests completed
async: exiting from thread
===============================================================================
test cases: 1 | 1 passed
assertions: - none -

==4019== 
==4019== HEAP SUMMARY:
==4019==     in use at exit: 2,864,310 bytes in 14,724 blocks
==4019==   total heap usage: 330,961 allocs, 316,237 frees, 36,759,700 bytes allocated
==4019== 
==4019== 1 bytes in 1 blocks are possibly lost in loss record 2 of 1,119
==4019==    at 0x4C2C100: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==4019==    by 0x40C6D1: _M_clone (functional:1878)
==4019==    by 0x40C6D1: std::_Function_base::_Base_manager<run_http_invalid_request_line(ight::common::async::Async&)::{lambda(char const*)#1}>::_M_manager(std::_Any_data&, std::_Function_base::_Base_manager<run_http_invalid_request_line(ight::common::async::Async&)::{lambda(char const*)#1}> const&, std::_Manager_operation) (functional:1914)
==4019==    by 0x41FBB3: std::function<void (char const*)>::function(std::function<void (char const*)> const&) (functional:2410)
==4019==    by 0x41FC2E: operator= (functional:2243)
==4019==    by 0x41FC2E: set_logger (log.hpp:69)
==4019==    by 0x41FC2E: ight::common::net_test::NetTest::set_log_function(std::function<void (char const*)>) (net_test.hpp:39)
[...]

PR #141, opened 2015-07-24
Register test complete callback along with test

### LINKED ISSUE #139: Even when using async "test complete" must be a callback of the test
As discussed with @lorenzoPrimi during the review of measurement-kit/measurement-kit-app-ios#14, it would be handy if the "test ended" callback was a callback of the test rather than a callback of the Async object. This way, in fact, it would be possible to bind the lambda closure to the self object, while it's more convoluted to do the same when the "test ended" is a callback of Async, because when you set the callback the ObjectiveC part of the test typically does not exist yet.

bassosimone:
This should solve the problem described in issue #139, even though the solution implemented here is different from the one suggested in issue's description (but that's irrelevant).

I've tested this on Linux, using valgrind to make sure that except for the leak reported in #138 there are no further memory errors.

Closes #139.

### DIFF
--- a/include/ight/common/async.hpp
+++ b/include/ight/common/async.hpp
@@ -34,8 +34,11 @@ class Async {
     /// Constructor with specified poller
     Async(SharedPointer<Poller> p);
 
-    /// Run the specified network test
-    void run_test(SharedPointer<NetTest> test);
+    /// Run the specified network test and call a callback when done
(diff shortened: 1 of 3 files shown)


bassosimone:
Tested and approved by @lorenzoPrimi, thanks!

PR #142, opened 2015-07-29
Big rename from libight to measurement_kit

bassosimone:
- rename the namespaces and of course move header files

- rename pretty much everything, including variables in the build system, commented out code and documentation

- while there

    - get rid of functions prefixed by `ight_` and instead put them into the `measurement_kit` namespace

    - make sure we always use two levels of namespaces only, so that we have measurement_kit::sublibrary::ClassName

    - move dns and http into their own "sublibraries"

    - create easier-to-use per sublibrary headers, e.g. `<measurement_kit/common.hpp>`, to simplify the task of importing sublibraries for other devs


hellais:
Lgtm!

hellais:
Long live MK!

PR #153, opened 2015-08-29
Teach portolan traceroute about reply packets.

bassosimone:
<strike>This needs some **more testing before it can be merged**. However I'm opening a pull request to bind it to the upcoming milestone and know this is something to work on.</strike>

This has now been refined and moderately tested. Note that this is needed by *Portolan*.

### DIFF
--- a/include/measurement_kit/traceroute/android.hpp
+++ b/include/measurement_kit/traceroute/android.hpp
@@ -100,6 +100,22 @@ private:
   ProbeResult on_socket_readable();
 
   /// Returns the source address of the error message.
+  /// \param s IPv4 socket address
+  /// \return source address of the error message
+  static std::string get_source_addr(const sockaddr_in *s);
+
+  /// Returns the source address of the error message.
+  /// \param s IPv6 socket address
+  /// \return source address of the error message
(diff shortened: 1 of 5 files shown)


bassosimone:
Closing for now: will reopen when this has been tested extensively.

bassosimone:
Hi @ValerioLuconi, can you please review this diff and tell me what you think? I like to merge this so that we can move forward Portolan integration! Thanks.

bassosimone:
@ValerioLuconi confirmed that this diff works as expected from both CNR premises (where it behaves like scamper) and from home.

bassosimone:
I've checked why regress tests fail. In both cases (clang and gcc) the code fails with error `EHOSTUNREACH` triggered during `sendto()`. This error is a transient network error that we can safely ignore (i.e. this error does not tell us that code in this pull request is wrong, it just informs us of a transient network situation). However, this also reminds me that traceroute code should not throw an exception when `sendto()` fails; rather this error should be routed to `on_error()` so that the programmer can write the proper handling code. Since this change is not trivial, I will do this in another pull request.

PR #155, opened 2015-08-31
Improve net/buffer

bassosimone:
This pull request aims to simplify and to make more robust net/buffer code. More details in the individual commit messages. This is what was needed to prepare this part for version 0.1.

Issue #111, opened 2015-03-30
Make std::function usage robust with respect to self-destruction of their closure


bassosimone:
This problem was first exposed by fix 8c992d7 of #110. When a `std::function` assigns to itself and the code is compiled using gcc, the closure is immediately destroyed. If the closure is used to keep safe the object that owns the std::function, this leads to use after free with possible crash.

The following minimal example ([derived a longer example of this behavior](https://github.com/bassosimone/cxx-experiments/commit/b2a4d3a2)) shows the problem:

```C++
#include <functional>
#include <memory>

struct Foo {
    std::function<void()> func;
};

static Foo *make_foo() {
    std::shared_ptr<Foo> foobar{new Foo};
    foobar->func = [foobar]() { foobar->func = nullptr; };
    return foobar.get();
}

int main() {
    make_foo()->func();
}
```

(`make_foo()` only returns a pointer to `foobar` which is however not destroyed at end of `make_foo()` scope because it is kept alive by the capture list of the lambda assigned to `foobar->func`.)

Compiled with `g++ -Wall -std=c++11 minimal.cpp -o minimal` and executed with `valgrind ./minimal, this leads to the following memory errors:

    ==5873== Memcheck, a memory error detector
    ==5873== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
    ==5873== Using Valgrind-3.10.0.SVN and LibVEX; rerun with -h for copyright info
    ==5873== Command: ./minimal
    ==5873== 
    ==5873== Invalid free() / delete / delete[] / realloc()
    ==5873==    at 0x4C2C2BC: operator delete(void*) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
    ==5873==    by 0x40106B: std::_Function_base::_Base_manager<make_foo()::{lambda()#1}>::_M_destroy(std::_Any_data&, std::integral_constant<bool, false>) (in /tmp/minimal)
    ==5873==    by 0x400F25: std::_Function_base::_Base_manager<make_foo()::{lambda()#1}>::_M_manager(std::_Any_data&, std::_Function_base::_Base_manager<make_foo()::{lambda()#1}> const&, std::_Manager_operation) (in /tmp/minimal)
    ==5873==    by 0x4012A4: std::function<void ()>::operator=(decltype(nullptr)) (in /tmp/minimal)
[...]

PR #156, opened 2015-08-31
Improve net/transport code

bassosimone:
This pull request returns to the net/transport code and prepares it for the upcoming release. It either simplifies code where excessive complexity exists, or it implements missing features.

More details in the individual commit messages. More can be done to improve this area of measurement-kit. But given the pending release and since I did not want to fuzz the patch too much, this pull request only contains what I consider to be the most urgent fixes.


include/measurement_kit/http/http.hpp
```diff
@@ -328,7 +327,7 @@ class Stream {
      */
     void on_connect(std::function<void(void)>&& fn) {
         connect_handler = fn;
-        connection->on_connect([this]() {
+        connection.on_connect([this]() {
```
hellais: Just to understand the rational. You think the memory usage reduction is not worth complicating the code? If so, what is the benefit of not using the shared pointer?

include/measurement_kit/net/connection.hpp
```diff
@@ -32,20 +34,18 @@ namespace net {
...
     Bufferevent bev;
     dns::Request dns_request;
     unsigned int connecting = 0;
-    char *address = NULL;
-    char *port = NULL;
-    StringVector *addrlist = NULL;
-    char *family = NULL;
-    StringVector *pflist = NULL;
+    std::string address;
+    std::string port;
```
hellais: Should port not be an int?

hellais:
I like :+1: 

A cursory glance does not lead me to suspect this has any issues in merging. Will review it with more detail and testing when I do #155.

As with that if for some reason I fail to do so by a reasonable time, proceed with merge.

bassosimone:
Thanks!

include/measurement_kit/net/connection.hpp
bassosimone: At that point in the code it could be an integer as well as a string. In fact, later on the port value is passed to an overloaded function that could either receive the port as an integer or as a string.

Historically here there is a string because libneubot was designed also to be called from Python `ctypes` and moving checks in C++ simplified the glue code between Python and C++.

include/measurement_kit/http/http.hpp
bassosimone: The benefit of not using a shared pointer is simpler code (hence simpler review and most importantly simpler debugging). On the contrary the benefit of using a shared pointer is that the pointee is alive as long as the shared pointer is alive. At a certain point this looked promising because it could allow one to emulate the semantic of languages like JavaScript. In particular I explored the possibility of using lambda closures to keep objects alive.

However, after trying this strategy out, I become convinced it was not so good.The code was becoming more complex to read. It was not very reliable to store objects inside the lambda closure (strange implementation issues, leaks because there is no cycle-aware garbage collector, etc.). As an example, see the issues reported in #134 and #111.

So I decided the wisest thing to do was to remove complexity (and probably it would have been better to apply KISS since the beginning). This payed off because I finally understood, for example, the reason of #134 (the fix for which is in a branch that will soon become a pull request).

bassosimone: Reading more carefully the header file, since `connection` is a transport, nothing really changed in this case since before. That is, `connection` used to be `SharedPointer<Transport>`. Now `Transport` (the abstract thing) is `TransportInterface` and `Transport` contains a `SharedPointer<TransportInterface>`. So, at the end of the day, here what really changed is that using transport is easier since the template is hidden.

bassosimone:
Merging!

PR #157, opened 2015-09-02
README.md: improve description

bassosimone:
- define capabilities and goals more clearly than before

- mention implemented tests

- mention facilities allowing you to implement your tests

- mention future plans including more ooni tests, oonib interaction, performance tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,50 @@
 
 [![Build Status](https://travis-ci.org/measurement-kit/measurement-kit.svg?branch=master)](https://travis-ci.org/measurement-kit/measurement-kit)
 
-MeasurementKit is an experimental library that provides common functionalities
-useful to implement open measurement tools on mobile platforms.
+MeasurementKit is a library that implements open measurement methodologies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,8 +2,50 @@
 
 [![Build Status](https://travis-ci.org/measurement-kit/measurement-kit.svg?branch=master)](https://travis-ci.org/measurement-kit/measurement-kit)
 
-MeasurementKit is an experimental library that provides common functionalities
-useful to implement open measurement tools on mobile platforms.
+MeasurementKit is a library that implements open measurement methodologies
```
hellais: I would specify we are talking about network measurements.

Perhaps it's also good to clarify that users of the library may be either those that want the results of the measurements to be used in the application logic of some arbitrary piece of software or it may be of interested to researchers or people that have specific network measurement needs.

hellais:
I will merge this since it's good, but perhaps it's worth thinking about expanding the overall description a bit more, but can be done in another PR.

Thanks a lot!

:+1:

PR #160, opened 2015-09-03
Improve async interface and implementation

bassosimone:
This pull request improves the interface and implementation of async. Among other things, this fixes issue #138 (i.e. after this pull request the memory leak in `./test/common/async` disappears).

Documentation also added. More details in individual commit messages.

### DIFF
--- a/src/common/async.cpp
+++ b/src/common/async.cpp
@@ -4,11 +4,15 @@
 
 #include <measurement_kit/common/async.hpp>
 #include <measurement_kit/common/logger.hpp>
+#include <measurement_kit/common/net_test.hpp>
 #include <measurement_kit/common/poller.hpp>
+#include <measurement_kit/common/pointer.hpp>
 
 #include <mutex>
 #include <map>
+#include <stdexcept>
 #include <thread>
+#include <utility>
 
 #include <event2/thread.h>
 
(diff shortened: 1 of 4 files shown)


src/common/async.cpp
```diff
@@ -4,156 +4,172 @@
...
-  std::function<void(SharedPointer<NetTest>)> fn) {
-    LOCKED(
-        debug("async: test inserted");
-        auto ret = state->ready.insert(std::make_pair(test, fn));
-        if (!ret.second) {
-            throw std::runtime_error("async: element already inserted");
-        }
+void Async::run_test(NetTestVar test, std::function<void(NetTestVar)> fn) {
+    LOCKED({
+        NetTest *ptr = test.operator->(); // get() does not check for NULL
```
bassosimone: Actually since 64c9992 `SharedPointer::get()` checks for `nullptr` so we can modify the above to use `get()`.

bassosimone:
I just added a new commit (4128435eb7) that implements thread-safe handling of logs. Basically now, when you call `pump()` you are notified in your thread context of both the tests that terminated (as it used to be) and of the logs that were generated for each test. This should further simplify writing Apps embedding measurement-kit, since now no lambda function is ever called from another thread.

bassosimone:
@lorenzoPrimi suggests that this diff is not what a iOS developer expects, because typically a iOS developer expects callbacks to be called by background threads.

Closing this pull request. I will soon open a new one that only fetches the goodies implemented here and avoids all the pump() machinery per the above comment.

PR #170, opened 2015-09-05
Route QueryImpl() exception to the callback

bassosimone:
Route exceptions raised by `QueryImpl::QueryImpl()` to the callback of `QueryImpl::issue()` so that users in practice do not see the exception, but rather the callback is always called.

This is already documented as the behavior of `Query` in documentation.


bassosimone:
Superseded by #175. Closing this one.

PR #175, opened 2015-09-07
Cleanup DNS API

### LINKED ISSUE #45: Clean up the API of the DNS library
* Make the errors consistent with OONI errors

bassosimone:
This pull request finishes the cleanup of the DNS API:

1. route early errors when issuing the query through the callback to report the error (as opposed to raising an exception)

2. implement support for PTR

3. provide error as first argument of DNS callback for consistency with HTTP code

4. use defines for query classes and types (as opposed to using strings)

5. document everything

Supersedes #170 (which contains only item 1. above).

Closes #45.


anadahz:
Seems OK to me!

include/measurement_kit/dns/query.hpp
```diff
@@ -34,8 +37,8 @@ class Query {
     Query() {}
 
     /// Start an async DNS request.
-    Query(std::string query, std::string address,
-          std::function<void(Response)> func,
+    Query(QueryClass dns_class, QueryType dns_type, std::string name,
```
hellais: I am not 100% convinced of the plan to use a struct to define the query type and class, but I will think a bit more about this.
Generally this is something I have found a bit annoying with other libraries since it means I need to know where to find the query types and names.
Perhaps we can overcome this by making an alternative constructor that takes the query name as a string.

include/measurement_kit/dns/response.hpp
```diff
@@ -52,9 +52,6 @@ class Response {
     /// Get the integer status code returned by evdns.
     int get_evdns_status() { return code; }
 
-    /// Map the status code returned by evdns to a OONI failure string.
-    std::string get_failure() { return map_failure_(code); }
```
hellais: This is a very important feature to have. Why was it removed?

include/measurement_kit/dns/response.hpp
```diff
@@ -64,9 +61,6 @@ class Response {
 
     /// Return the evdns type (e.g. DNS_IPv4_A)
     char get_type() { return type; }
-
-    /// Static function to map evdns status codes to OONI failures.
-    static std::string map_failure_(int code);
```
hellais: Same as above.

src/dns/error.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include <measurement_kit/common/error.hpp>
+#include <measurement_kit/dns/error.hpp>
+
+#include <event2/dns.h>
+
+namespace measurement_kit {
+namespace dns {
+
+common::Error dns_error(int code) {
```
hellais: I understand now why you removed it from the above. Still perhaps it makes sense to have the function inside of the namespace of the class, though it's also a bit bikeshedding.

I will think more about this too.

src/dns/query_impl.hpp
```diff
@@ -99,55 +110,77 @@ class QueryImpl {
...
+            std::string s;
+            if ((s = measurement_kit::unreverse_ipv4(name)) != "") {
+                dns_type = QueryType::REVERSE_A;
+                name = s;
+            } else if ((s = measurement_kit::unreverse_ipv6(name)) != "") {
+                dns_type = QueryType::REVERSE_AAAA;
+                name = s;
+            } else
+                throw InvalidNameForPTRError();
+        }
```
hellais: Note to self: think of how this is done. In particular PTR lookups require you to reverse the IP address and I am not 100% sure it should be done automatically as in above.

hellais:
I apologise for not reviewing this sooner and will probably not have time to do so until next week.
I took note of some things that by just skimming quickly the code came to mind, but I would ideally like to take a closer look at such an important PR.

If it's urgent and I am blocking on something please go ahead and merge it since also @anadahz signed off on it already.

Anyways good job and in general it seems like the DNS class is starting to look good.

include/measurement_kit/dns/query.hpp
bassosimone: Yes, I think having constructors using string is a good idea. I'll add a diff for that.

src/dns/error.cpp
bassosimone: It was removed because I migrated the code from `[](dns::Response resp) {...}` to `[](Error error, dns::Response resp) {...}`. This migration was done for consistency with other places in the code, and because I think it is cleaner to separate error and response.

Because of this change it was natural to move the conversion to OONI error in the Error class. Since also other parts of MK will need to map errors to OONI errors, furthermore, it seems to me more uniform that one always calls `error.as_ooni_error()` to access the OONI error bound to a specific MK error.

bassosimone:
I've added an additional commit that implements specifying query class and type as strings rather than as enums. Additionally, I've edited most uses of the library to use strings, to communicate that this is the preferred / simplest way to use the library. Also the documentation was updated accordingly.

include/measurement_kit/dns/query.hpp
bassosimone: Done (see below).

bassosimone:
I am now going to merge this to prepare the first alpha release of measurement-kit. @hellais, since you wanted to check the implementation of the feature that reverses DNS addresses, I'll add a blocker issue before v1.0.0 (semver) is out explicitly mentioned that this must be done before releasing.

PR #184, opened 2015-10-02
Add support for JSON and GeoIP

bassosimone:
We use jansson to provide JSON support. The choice of jansson was discussed looooong time ago with @hellais.

We use libmaxmindb to provide GeoIP support (this is the library to parse the new format GeoIP version).


bassosimone:
Checks failing because travis-ci does not like the recently added traceroute regress test. Have tested this locally trying to compile both using the built-in and the system-wide jansson and libmaxminddb.

PR #189, opened 2015-10-06
Modify libevent to compile for all Android archs

bassosimone:
This pull request has been updated and now only contains:

- changes in Makefile.am to reduce the compile time to recompile libevent for Android and iOS (rationale: we're integrating libevent, we're not developing it)

- the [arc4random_addrandom fix v2](https://github.com/measurement-kit/libevent/commit/8b275d967d7ffd95d5cc12069aef35669126c6d9) fix, needed to compile on Android under most architectures

- a revert for e7ea40bd9ad so that all Android architectures are built

To test this diff, I suggest to take the low-level approach of recompiling everything:

```
make distclean
git clean -dfx
cd mobile/android
# read README.md
./scripts/build.sh /path/to/ndk 21
```

and then do the following for each architecture:

- check `build/$arch/config.h` and see whether `arc4random` and `arc4random_addrandom` are supported

- disassemble `jni/$arch/libevent.a` (names change unfortunately between build and arch, see `./scripts/build_target.sh` for the mapping) and check `evutil_secure_rng_add_bytes()`:

    - if both `arc4random` and `arc4random_addrandom` are supported, you should see that the function calls `_arc4random_addrandom` (this is what happens on MacOS)

    - if only `arc4random` is supported, you should see the few assembly instructions that make an empty function (this is what should happen on OpenBSD and on selected Android architectures)

    - if both are not supported, you should see the inline implementation of libevent's `arc4random_addrandom` that at a certain point should call `arc4_stir` (this is what happens on Linux)

### DIFF
--- a/scripts/update-deps.sh
+++ b/scripts/update-deps.sh
@@ -25,7 +25,7 @@ get boostorg/type_traits boost-1.56.0 boost/type_traits
 get boostorg/typeof boost-1.56.0 boost/typeof
 get boostorg/utility boost-1.56.0 boost/utility
 get joyent/http-parser v2.1-47-g1b31580 http-parser
-get measurement-kit/libevent release-2.0.22-stable-12-g8aad8b7 libevent
+get measurement-kit/libevent release-2.0.22-stable-18-gf1feb10 libevent
 get bassosimone/yaml-cpp master yaml-cpp
 get akheron/jansson master jansson
 get measurement-kit/libmaxminddb master libmaxminddb
--- a/mobile/android/scripts/build.sh
+++ b/mobile/android/scripts/build.sh
@@ -9,7 +9,9 @@ if [ $# -eq 3 ]; then
 elif [ $# -eq 2 ]; then
     NDK_DIR=$1
     API=$2
-    for arch in arm-linux-androideabi mipsel-linux-android x86; do
+    for arch in aarch64-linux-android arm-linux-androideabi \
+      arm-linux-androideabi-v7a mipsel-linux-android mips64el-linux-android  \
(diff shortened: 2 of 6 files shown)


scripts/update-deps.sh
```diff
@@ -25,5 +25,5 @@ get boostorg/type_traits boost-1.56.0 boost/type_traits
 get boostorg/typeof boost-1.56.0 boost/typeof
 get boostorg/utility boost-1.56.0 boost/utility
 get joyent/http-parser v2.1-47-g1b31580 http-parser
-get bassosimone/libevent release-2.0.21-stable-52-g86b586a libevent
+get measurement-kit/measurement-kit-libevent release-2.0.22-stable-15-ged01cec libevent
```
bassosimone: @DavideAllavena could you please update the name of the repository here? It was renamed and now it's called just libevent. Thanks!

bassosimone:
@DavideAllavena fyi: I have rebased the branch and added more clear instructions regarding how to properly test this. Thank you for your patience and help!

DavideAllavena:
Checked and disassembled binaries compiled for android-21:

* `aarch64-linux-android`, `arm-linux-androideabi`, `mips64el-linux-android`, `x86_64`:
  * `evutil_secure_rng_add_bytes()` has few assembly instructions
  *  only `arc4random` is supported

* `mipsel-linux-android`, `x86_64`: 
  * `evutil_secure_rng_add_bytes()` calls `_arc4random_addrandom`
  *  `arc4random` and `arc4random_addrandom` are supported

Everything as expected.

bassosimone:
Good, thanks!

PR #200, opened 2015-10-19
Introduce MeasurementKit JNI API

bassosimone:
This pull request, which is not yet complete, starts sketching out MeasurementKit JNI API.

- - -

*Update (2015-11-01)*: this is now [used in the prototype app](https://github.com/measurement-kit/measurement-kit-app-android/pull/1). There are still couple of things that I need to figure out, such as:

1. how to reliably retrieve the device's DNS server

2. how to handle exceptions (many exceptions are currently suppressed)

3. whether the archive containing jniLibs and Java files is the optimal solution to deploy (this is in line with what is done by open-peer) or whether we can do better


bassosimone:
Closing this pull request. I've experimented with a simpler approach in bassosimone/measurement-kit-jni-libs.

PR #202, opened 2015-10-26
send_probe send errorto error_cb_ and no more throws exceptions.

### LINKED ISSUE #196: traceroute: route send_probe() errors to on_error()
As of #153 merged, the regress test fails because of `EHOSTUNREACH`. This is because `send_probe()` triggers exception in case of send failure. A better approach would be to route errors to the `on_error()` callback instead. This should also avoid breaking the regress tests (for some reason doing the traceroute to OpenDNS with an invalid DNS packet from travis-ci fails).

The simplest way to implement this is to rename `send_probe()` as `send_probe_()` and then define `send_probe()` as a new wrapper function that calls `send_probe_()` forwarding parameters and wraps such invocation inside a `try-catch` block, forwarding the error to `on_error()`, pretty much like it's already done in `event_callback()`.

DavideAllavena:
This pull request is based on issue #196.

Code is compiling and make check passes.

Closes #196.

### DIFF
--- a/src/traceroute/android.cpp
+++ b/src/traceroute/android.cpp
@@ -107,6 +107,7 @@ AndroidProber::AndroidProber(bool a, int port, event_base *c)
 
 void AndroidProber::send_probe(std::string addr, int port, int ttl,
                                std::string payload, double timeout) {
+
   int ipproto, ip_ttl, family;
   sockaddr_storage ss;
   socklen_t sslen;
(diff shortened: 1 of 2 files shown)


include/measurement_kit/traceroute/android.hpp
```diff
@@ -59,7 +59,6 @@ class AndroidProber : public measurement_kit::common::NonCopyable,
   /// \param use_ipv4 Whether to use IPv4
   /// \param port The port to bind
   /// \param evbase Event base to use (optional)
-  /// \throws Exception on error
```
bassosimone: Can you document that, in case of error, the error function is called?

src/traceroute/android.cpp
```diff
@@ -115,13 +116,16 @@ void AndroidProber::send_probe(std::string addr, int port, int ttl,
   measurement_kit::debug("send_probe(%s, %d, %d, %lu)", addr.c_str(), port, ttl,
              payload.length());
 
-  if (sockfd_ < 0)
-    throw std::runtime_error("Programmer error"); // already close()d
```
bassosimone: Can you integrate the above comment in the exception message? It is useful to know which kind of programmer error was triggered.

include/measurement_kit/traceroute/android.hpp
DavideAllavena: Comment is added

src/traceroute/android.cpp
DavideAllavena: Comment readded

include/measurement_kit/traceroute/android.hpp
bassosimone: This is actually wrong: the documentation change is only relative to `send-probe()`. As of now, the constructor of `AndroidProber` still raises exceptions.

src/traceroute/android.cpp
bassosimone: Thanks!

bassosimone:
Looks good to me, could you please squash everything into the first commit? Then, I'll merge!

bassosimone:
Awesome work @DavideAllavena :+1:! Thank you!

bassosimone:
(P.S. I am particularly happy that this is merged because it repaired travis-ci build!)

PR #203, opened 2015-10-26
traceroute: make constructor lazy 

### LINKED ISSUE #198: traceroute: make constructor lazy
Currently the constructor throws exceptions on error. Since we don't need to Prober constructed until we `send_probe()`, we can change the constructor to only store constructor parameters, and then we can modify `send_probe()` to construct the object the first time it is invoked. This way, we can also route errors occurring during the construction phase to `on_error()`, this making the whole implementation of the traceroute fully async friendly.

DavideAllavena:
AndroidProber constructor body is empty. Errors are passed to error_cb_ by init function.

With gcc 5.2.0  this warning are shown:

```
src/traceroute/interface.cpp: In member function ‘measurement_kit::traceroute::ProbeResultMeaning measurement_kit::traceroute::ProbeResult::get_meaning()’:
src/traceroute/interface.cpp:77:72: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘measurement_kit::traceroute::ProbeResultMeaning’ [-Wformat=]
                            icmp_type, icmp_code, PRM_::GOT_REPLY_PACKET);
                                                                        ^
src/traceroute/interface.cpp:84:40: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘measurement_kit::traceroute::ProbeResultMeaning’ [-Wformat=]
                              m->meaning);
                                        ^
src/traceroute/interface.cpp:89:48: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘measurement_kit::traceroute::ProbeResultMeaning’ [-Wformat=]
                          icmp_code, PRM_::OTHER);
                                                ^
```

Closes #198.

### DIFF
--- a/src/traceroute/android.cpp
+++ b/src/traceroute/android.cpp
@@ -51,14 +51,21 @@ namespace measurement_kit {
 namespace traceroute {
 
 AndroidProber::AndroidProber(bool a, int port, event_base *c)
-    : use_ipv4_(a), evbase_(c) {
+    : use_ipv4_(a), evbase_(c), port_(port) {}
+
+void  AndroidProber::init() {
+
+  if (sockfd_ >= 0 && evp_ != nullptr)
+    return;
+  else if (sockfd_ >= 0 || evp_ != nullptr)
+    throw std::runtime_error("Objects not properly initialized.");
 
   sockaddr_storage ss;
   socklen_t sslen;
   int level_sock, opt_recverr, level_proto, opt_recvttl, family;
   const int val = 1;
 
-  measurement_kit::debug("AndroidProber(%d, %d, %p) => %p", use_ipv4_, port,
+  measurement_kit::debug("AndroidProber(%d, %d, %p) => %p", use_ipv4_, port_,
              (void *)evbase_, (void *)this);
 
   if (use_ipv4_) {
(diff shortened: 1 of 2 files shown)


src/traceroute/android.cpp
```diff
@@ -113,6 +119,8 @@ void AndroidProber::send_probe(std::string addr, int port, int ttl,
   socklen_t sslen;
   timeval tv;
 
+  init();
```
bassosimone: This is not correct: you don't want to `init()` every time you fire off a probe. What you want is to emulate the constructor semantic, but opportunistically; i.e., you want to run `init()` only once when it's needed to do so. An alternative -- and perhaps more robust approach -- is that you make `init()` idempotent, meaning that it only initialize stuff if it's needed otherwise it just returns immediately.

bassosimone:
Also, could you report the warning in another issue?

src/traceroute/android.cpp
DavideAllavena: Pushed a commit that should fix this.
I made init() idempotent.

bassosimone: Ok, thank you!

src/traceroute/android.cpp
```diff
@@ -78,50 +85,60 @@ AndroidProber::AndroidProber(bool a, int port, event_base *c)
...
+    error_cb_(std::runtime_error("bind() failed"));
+    return;
   }
 
   // Note: since here we use `this`, object cannot be copied/moved
   if ((evp_ = event_new(evbase_, sockfd_, EV_READ, event_callback, this)) ==
       NULL) {
     cleanup();
-    throw std::runtime_error("event_new() failed");
+    error_cb_(std::runtime_error("event_new() failed"));
```
bassosimone: Could you please also add a return statement here for robustness? We may need to add code below that in the future, and it's more future-proof to have a return statement already in place.

include/measurement_kit/traceroute/interface.hpp
```diff
@@ -86,12 +86,12 @@ class ProberInterface {
...
   /// Send a traceroute probe
+  /// In case of error this function emits error-event
   /// \param addr Destination address
   /// \param port Destination port
   /// \param ttl Time to live
   /// \param payload packet payload
   /// \param timeout Timeout for this probe
-  /// \throws Exception on error
   virtual void send_probe(std::string addr, int port, int ttl,
                           std::string payload, double timeout) = 0;
```
bassosimone: Have you rebased to master?! This hunk was already present in #202.

bassosimone: Also, could you change the documentation of `AndroidProber::AndroidProber()` to remove the statement saying that the constructor throws on error, since that is not true anymore?

src/traceroute/android.cpp
DavideAllavena: I agree. Added!

bassosimone:
Thank you! Merging!

PR #215, opened 2015-11-17
Add instructions to use jansson and libmaxminddb

### LINKED ISSUE #208: ios: fix build script wrt jansson and libmaxminddb
The iOS build script should explicitly tell configure to use the builtin jansson and libmaxminddb, to avoid using system-wide libraries for another architecture, if they are installed.

This should mirror [what is already done for boost etc](https://github.com/measurement-kit/measurement-kit/blob/master/mobile/ios/scripts/build_arch.sh#L35). To see what flags are defined, I suggest to read the output of `./configure`.

AntonioLangiu:
Like in [issue-208](https://github.com/measurement-kit/measurement-kit/issues/208).

To make sure that configure use the built-in version instead of the version in the system.

It works. ./configure prints no more this output:

    configure: WARNING: No jansson found: will use the builtin jansson
    configure: WARNING: No libmaxminddb found: will use the builtin libmaxminddb

Closes #208.


bassosimone:
This is good, thanks!

PR #217, opened 2015-11-18
More consistent configure flags

### LINKED ISSUE #211: build system: normalize --with names
IMO if the project is named boost, the flag should be `--with-boost`. If the project is named libevent, the flag should be `--with-libevent`. That is, `--with-` followed by the name of the project on github. This makes it super easy for someone to guess the name of the dependency. Currently, I guess the flags for `yaml-cpp` and `boost` are wrong, because they prefix `lib` to the name.

bassosimone:
Closes #211.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,10 +113,10 @@ at specific directories using the following flags:
 - `--with-libevent=PREFIX` that tells `./configure` to use the
 libevent library and headers installed at PREFIX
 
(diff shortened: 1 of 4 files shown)


bassosimone:
Tested by me and @AntonioLangiu building measurement-kit for iOS and Android.

PR #219, opened 2015-11-19
Use common::Error rather than std::runtime_error.

### LINKED ISSUE #207: traceroute: use Error rather than std::runtime_error for errors
Similarly to what [already done for net](https://github.com/measurement-kit/measurement-kit/blob/master/include/measurement_kit/net/error.hpp):

- define a header for containing errors for traceroute

- define inside this header all the possible traceroute errors

- modify interface and implementation of `on_error()` to receive as first parameter a child of `common::Error` rather than `std::runtime_error()`,

DavideAllavena:
closes #207


bassosimone:
Good work, thanks!

PR #220, opened 2015-11-19
Issue 206

### LINKED ISSUE #206: traceroute: use virtual and override consistently
- when a method is firstly declared as virtual, the `virtual` keyword must be added, e.g.

```C++
    virtual void foo() = 0;
```

- derived methods shall be marked only with `override` instead, e.g.

```C++
    void foo() override;
```

- I'd also remove `final` since it's overkill at this stage (and probably forever)

DavideAllavena:
closes #206


bassosimone:
This looks good, merging, thanks!

PR #231, opened 2015-11-22
Feature/code coverage

### LINKED ISSUE #23: Find a way to generate code coverage reports
It would be nice to have reports on what is the current code coverage of libight. The unittesting framework we chose (Catch) does not seem to support this natively, but it looks like it is something that can be done with `clang --coverage`.

Doing a little bit of research I came up with these links that may be useful:

* http://stackoverflow.com/questions/19881928/code-coverage-in-clang
* http://clang-developers.42468.n3.nabble.com/Code-coverage-on-clang-td4033066.html

Alternatively expcov also seems promising:

* http://www.expcov.com/

AntonioLangiu:
Enable code coverage with [coveralls](https://coveralls.io/github/measurement-kit/measurement-kit).

Closes #23


bassosimone:
Merging, thank you!

PR #241, opened 2015-11-24
Buffer::readline returns Maybe<std::string>.

### LINKED ISSUE #235: Buffer::readline() should return Maybe<std::string>

bassosimone:
Closes #235.


bassosimone:
Merging!

PR #244, opened 2015-11-24
Allowed creation of empty maybe.

### LINKED ISSUE #242: allow creation of empty maybe
- define a new error: "maybe not initialized error"

- define a default constructor for maybe that sets the error as the above error

- add test case that an empty maybe is:

    - false

    - with as_error() equal to such error

    - with as_value() that throws such error

DavideAllavena:
`Maybe<type> maybe`  now can be called.

MaybeNotInitializedError() is called when accessing the variable without initialization.

Closes #242

### DIFF
--- a/include/measurement_kit/common/error.hpp
+++ b/include/measurement_kit/common/error.hpp
@@ -53,6 +53,12 @@ class GenericError : public Error {
     GenericError() : Error(1, "unknown_failure 1") {} ///< Default constructor
 };
 
+/// Maybe not initialized error
+class MaybeNotInitializedError : public Error {
+  public:
+    MaybeNotInitializedError() : Error(2, "unknown_failure 2") {}
+};
+
 } // namespace common
 } // namespace measurement_kit
 #endif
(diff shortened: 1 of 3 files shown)


include/measurement_kit/common/error.hpp
```diff
@@ -53,6 +53,12 @@ class GenericError : public Error {
     GenericError() : Error(1, "unknown_failure 1") {} ///< Default constructor
 };
 
+/// Maybe not initialized error
+class MaybeNotInitializedError : public Error {
+  public:
+    MaybeNotInitializedError() : Error(2, "Maybe_template_not_initialized 2") {}
```
bassosimone: This should be `unknown_failure 2`

include/measurement_kit/common/maybe.hpp
```diff
@@ -13,6 +13,11 @@ namespace common {
 /// Maybe contains a value of type T
 template <typename T> class Maybe {
   public:
+    /// Empty constructor
+    Maybe() {
```
bassosimone: Could you use the ` : error_( ... )` syntax here please?

bassosimone:
Excellent, thanks!

PR #245, opened 2015-11-24
Teach OONI tests to run themselves

### LINKED ISSUE #243: Increase abstraction of ooni tests api
I propose to do little more work to increase the abstraction of the ooni tests api. In particular, the following illustrates what I'd like to see:

```C++
#include <measurement_kit/ooni.hpp>
#include <list>
#include <string>

void run_sync(std::list<std::string> logs) {
    ooni::DnsInjectionTest()
        .set_nameserver("8.8.8.8:53")
        .set_input_file_path("test/fixtures/hosts.txt")
        .set_verbose()
        .on_log([=](const char *s) { logs.push_back(s); })
        .run();
};

void run_async(std::list<std::string> logs, std::function<void()> callback) {
    ooni::DnsInjectionTest()
        .set_nameserver("8.8.8.8:53")
        .set_input_file_path("test/fixtures/hosts.txt")
        .set_verbose()
        .on_log([=](const char *s) { logs.push_back(s); })
        .run([=]() { callback(); });
};
```

In the first case, the test is run synchronously instantiating a custom `Poller` object and running its event loop until the test completes. Of course, for greater customization, one should be able to provide a pointer to a custom poller that he/she allocated.

In the second case, the test is run asynchronously using the default `Async` object. Of course, for greater customization, one should be able to provide a pointer to a custom async that he/she allocated.

bassosimone:
Tested locally also with Valgrind to make sure this new set of patches does not add memory errors.

This is not enough to close #243, because improvements are needed to allow one to provide a specific poller or async to the code. However, to do so we would need to change the underlying implementation too much, and I don't want to do this before making a release.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,9 +79,12 @@ test/net/buffer
 test/net/connection
 test/net/connection_check_fds
 test/ooni/dns_injection
+test/ooni/dns_injection_test
 test/ooni/net_test
 test/ooni/http_invalid_request_line
+test/ooni/http_invalid_request_line_test
 test/ooni/tcp_connect
+test/ooni/tcp_connect_test
 test/ooni/tcp_test
 test/dns/defines
 test/dns/error
--- a/README.md
+++ b/README.md
@@ -186,34 +186,53 @@ Then type `pod install` and open `.xcworkspace` file (beware not to open the
 
 ## How to use MeasurementKit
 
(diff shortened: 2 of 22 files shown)


DavideAllavena:
According to valgrind there are some memory leak due to ld.

I'm fine in merging

PR #251, opened 2015-11-27
Fix coverage reports

AntonioLangiu:


bassosimone:
Thanks @AntonioLangiu!

PR #253, opened 2015-11-27
fix for http-invalid-request-line

### LINKED ISSUE #101: Fix ooni backend addresses used by regress tests
In an email exchange @hellais suggested:

- to use 213.138.109.232 as backend for http_invalid_request_line

- to use one or more of the [URLs published by Citizen Lab](https://github.com/citizenlab/test-lists/blob/master/csv/global.csv) for tcp_connect

- to use 197.156.64.3 as backend for dns_injection

bassosimone:
This diff is the starting point to fix http-invalid-request-line test when the backend used is an echo server rather than a properly working HTTP server.

So far, I've changed the backend address according to what is indicated in #101. Then, I've added code to route errors in the read and write callbacks of the stream. I think this is a better approach than the other branch that I opened, in which I was rewriting too much of HTTP too close to the release.

The diff works (meaning that routing exceptions prevent the exception thrown by the http-parser to terminate the test). However, I think more is needed to have something mergeable.

Specifically:

1. two new errors should be added to `include/measurement_kit/common/error.hpp`: one should represent a runtime exception received when reading, the other should represent a runtime exception received when writing

2. the code in this pull request should be modified to throw the two exceptions defined in step 1 respectively in the reading and in the writing case (this removing the need to XXX the code)

3. a new file should be added to the `include/measurement_kit/http` folder and should be called `error.hpp` for symmetry with other parts of measurement-kit

4. inside this file we shall move the errors currently defined in `src/http/response_parser.hpp` and furthermore such errors should inherit from `measurement_kit::common::Error` rather than from `std::runtime_error`, i.e. this `error.hpp` file should be similar to `net/error.hpp` or `dns/error.hpp`

5. we should make sure that all the errors that were defined by the parsing code are actually used and unused errors should be removed from `http/error.hpp`

6. we should use Valgrind to make sure that there are no memory errors introduced by the code that routes the exceptions (http code is still fragile); to this end it would probably suffice to test with Valgrind the http-invalid-request-line test because it triggers that behavior, but to be safe I'd also check other http-related regress tests

7. make sure that (and possibly add regress tests) the http-invalid-request-line test reports success if the response is exactly what was sent (this brings the question of how to get the content of the buffer that caused the http parser to freak out, which can probably be done through the http stream and the related transport) and failure otherwise

Closes #101.

### DIFF
--- a/test/ooni/http_invalid_request_line_test.cpp
+++ b/test/ooni/http_invalid_request_line_test.cpp
@@ -24,6 +24,15 @@ TEST_CASE("Synchronous http-invalid-request-line test") {
     for (auto &s : *logs) std::cout << s << "\n";
 }
 
+TEST_CASE("Synchronous http-invalid-request-line test with HTTP backend") {
+    Var<std::list<std::string>> logs(new std::list<std::string>);
+    ooni::HttpInvalidRequestLineTest()
+        .set_backend("http://data.neubot.org/") // Let's troll Davide!
+        .on_log([=](const char *s) { logs->push_back(s); })
+        .run();
+    for (auto &s : *logs) std::cout << s << "\n";
+}
+
 TEST_CASE("Asynchronous http-invalid-request-line test") {
     Var<std::list<std::string>> logs(new std::list<std::string>);
     bool done = false;
--- a/test/ooni/http_invalid_request_line_test.cpp
+++ b/test/ooni/http_invalid_request_line_test.cpp
@@ -18,7 +18,7 @@ using namespace measurement_kit;
 TEST_CASE("Synchronous http-invalid-request-line test") {
     Var<std::list<std::string>> logs(new std::list<std::string>);
     ooni::HttpInvalidRequestLineTest()
-        .set_backend("http://nexa.polito.it/")
+        .set_backend("http://213.138.109.232/")
         .on_log([=](const char *s) { logs->push_back(s); })
         .run();
     for (auto &s : *logs) std::cout << s << "\n";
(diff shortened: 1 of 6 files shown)


AntonioLangiu:
I have implemented 1,3 and 4. I'm not sure what do you mean with point 2. Can you explain it better please?

bassosimone:
@AntonioLangiu I think you already did that, since now you throw the new errors in the parser when something unexpected occurs.

bassosimone:
I just discussed with @hellais about point 7 above and we agreed that for this release we can get away without fully implementing point 7 (i.e. we don't need at this stage to access the offending buffer and compare it with what we sent). So, I'd say, let's check the cost of implementing the complete check and if it's too much let's delay the complete-check implementation after release of 0.1.0.

AntonioLangiu:
Ok, I tested with valgrind all the tests included in this folders:

* test/ooni
* test/http
* test/dns
* test/net

There are no error in all the tests.

 I have to check the point 5 and then we should be done.

bassosimone:
Ok, point 5. is now done by @AntonioLangiu. I will continue from there checking whether the test works correctly both when there is an echo backend and when the backend responds with HTTP (that is, point 7 indicated above).

test/ooni/http_invalid_request_line_test.cpp
```diff
@@ -18,7 +18,16 @@ using namespace measurement_kit;
...
+        .set_backend("http://213.138.109.232/")
+        .on_log([=](const char *s) { logs->push_back(s); })
+        .run();
+    for (auto &s : *logs) std::cout << s << "\n";
+}
+
+TEST_CASE("Synchronous http-invalid-request-line test with HTTP backend") {
+    Var<std::list<std::string>> logs(new std::list<std::string>);
+    ooni::HttpInvalidRequestLineTest()
+        .set_backend("http://data.neubot.org/") // Let's troll Davide!
```
bassosimone: cc: @DavideAllavena

bassosimone:
Alright, with `b725554` it is now confirmed that the http-invalid-request-line test is robust enough to work both with an `echo` backend (in such case it reports a parser error) and with a real HTTP server as backend (in such case, depending on the request pattern, it returns either 501 or 200).

I'd say we're done here and the diff I see looks good to me. I'll leave the final review of this diff and the merge decision to @AntonioLangiu, since I was the one who last touched this pull request.

AntonioLangiu:
Ok, looks good to me :+1:

PR #255, opened 2015-11-27
Repair coveralls and improve build process

bassosimone:
Start to repair coveralls and improve the build process. Work done so far:

- travis: split build command on multiple lines

- configure.ac: recognize clang++

- yaml-cpp: zap -Wno-missing-prototypes flag

- run quiet configure with summary at the end

- revert "Teach travis to send one time the coverage reports"  (712602c5da)

- reduce verbosity of all tests

Work to do before we can merge this pull request:

- apparently coveralls segfaults for gcc builds because of a version mismatch between the compiler (v4.8) and some other unspecified utility (v4.6): find a fix for that

- find a way to teach coveralls to produce less output than this, because this is too much

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,17 @@ before_install:
   - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
   - sudo apt-get update -qq
   - sudo apt-get install -qq -y --force-yes gcc-4.8 g++-4.8
+  - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 90
   - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 90
+  - sudo update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-4.8 90
   - sudo apt-get install tor
   - pip install --user cpp-coveralls
 
 install: autoreconf -i
 
+# Only measure coverage with clang for now because I cannot manage to
+# convince coveralls to find all sources when using gcc-4.8 and gcov-4.8
+# (For more context see measurement-kit/measurement-kit#255)
 script:
   - if [ "$CXX" = "clang++" ]; then
       ./configure -q CFLAGS=--coverage CXXFLAGS=--coverage LDFLAGS=--coverage;
@@ -21,7 +26,7 @@ script:
 
 after_success:
   - if [ "$CXX" = "clang++" ]; then
-      coveralls --exclude src/ext;
+      coveralls --gcov /usr/bin/gcov-4.6 --exclude src/ext;
     fi
 
 compiler:
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,10 @@ script:
   - make -j2 V=0
   - make -j2 check-am V=0
 
+# Silence gcov standard output since it produces way too much output
 after_success:
   - if [ "$CXX" = "clang++" ]; then
-      coveralls --gcov /usr/bin/gcov-4.6 --exclude src/ext;
+      coveralls --gcov /usr/bin/gcov-4.6 --exclude src/ext > /dev/null;
     fi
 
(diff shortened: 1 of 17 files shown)


AntonioLangiu:
The problem is that we force to install the version 4.8 of gcc but gcov remains at version 4.6.

I printed this on travis 
    $ gcov --version
    gcov (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3

We have to create a simlink to the correct version of gcov as we made with g++.

bassosimone:
@AntonioLangiu I suggest to see how neovim does that. IIRC they use several different combinations of compiles and gcovs and I'm quite sure they have a neat solution for that.

AntonioLangiu:
From what I can see [here](https://coveralls.io/builds/4298364) neovim has also some problem while building with a different compiler. They get code coverage just with gcc 4.6 and clang.

Our problem is with gcc 4.8. I think we have this version problem when we execute the tests and not when we call `cpp-coveralls`.

@bassosimone Can you please give me a link of the IIRC project because I didn't found it. (I found one but It doesn't have code coverage implemented)

bassosimone:
@AntonioLangiu I am not sure they only use gcc 4.6 because there is a build with gcov-5 which indicates that gcc-5 was used to build. And, in fact, [travis.yml#L63](https://github.com/neovim/neovim/blob/master/.travis.yml#L63) configures gcc-5 and gcov-5. Looking at how they implemented the whole thing:

- they set an environment variable for GCOV in travis.yml (see the link above)

- then they call coveralls using `--gcov` and `$(which $GCOV)` to get the correct path to gcov from the bare name of the executable (see [.ci/call_success.sh#L10](https://github.com/neovim/neovim/blob/master/.ci/after_success.sh#L10))

This is similar to what you wrote independently in 9297d933c8bd9, except that you pass to coveralls the bare path to gcov. This means that we're clearly heading towards the right direction.

To increase robustness and guarantee correctness, I think we should:

1) use an environment variable, as neovim does, so that we can select different gcov implementations for different compilers (for example we can use clang's own gcov for clang -- even though currently we are not seeing any breakage with clang using the system's gcov, I still think it is better to aim to correctness)

2) use the same trick of `$(which)` that neovim does, because it makes the code more robust with respect to future changes that may occur in the travis-ci environment.

bassosimone:
(As for IIRC: I meant "if I recall correctly" which is sometimes shortened as IIRC -- sorry, too much slang on this side!)

AntonioLangiu:
Ok, I will try to implement a build matrix to create this environment variable.

bassosimone:
@AntonioLangiu regarding reducing the output, I've noticed that we're [also measuring coverage of standard library headers](https://travis-ci.org/measurement-kit/measurement-kit/jobs/94011515#L963) which is clearly not interesting to us. We'll probably limit the number of produced lines by the coverage-measuring phase if we manage to tell it to focus only on measurement-kit code.

bassosimone:
As stated in f5cbe1e commit message, too much time was already spent on this issue that is not a real blocker for the release. In fact, we are now able to take coverage measurements with clang, and that shall suffice for now. We will be more precise in the future, trying to nail down the reason why we cannot take coverage measurements with gcc-4.8. And also it would be nice to have a build matrix. With time, all of this can be done. But, as we say to /dev/null:

bassosimone:
Added one extra commit to route gcov's output to /dev/null such that we stay within the maximum number of lines accepted by travis and we can inspect the whole log from inside travis,

bassosimone:
Hi @AntonioLangiu, I think I've managed to sort out several issues here. It was not possible to go as far as neovim is (and I'll open an issue for the remaining things to do), but I think that what we have here (coverage with clang without too much output on the stdout) should be enough for a first public release.

What do you think?

AntonioLangiu:
I think is ok for now. We will fix the coverage with GCC in the next releases.

PR #274, opened 2015-12-03
Reduce compiled library size on android

### LINKED ISSUE #266: Shrink the size of android developer's kit

bassosimone:
This should be merged after #273. It changes the way in which we compile Android code to squeeze quite significantly the compiled libraries at the price of losing debug symbols.

Closes #266.

### DIFF
--- a/mobile/android/scripts/build_target.sh
+++ b/mobile/android/scripts/build_target.sh
@@ -80,13 +80,22 @@ export LDFLAGS="${LDFLAGS} -L${SYSROOT}/usr/lib${LIB_SUFFIX} -L${ANDROID_TOOLCHA
     test -f Makefile && make clean
     echo "Configure with --host=${ARCH} and toolchain ${ANDROID_TOOLCHAIN}"
     test -x ${ROOTDIR}/../../configure || (cd ${ROOTDIR}/../.. && autoreconf -i)
-    ${ROOTDIR}/../../configure --host=${ARCH} --with-sysroot=${SYSROOT} \
+    ${ROOTDIR}/../../configure -q --host=${ARCH} --with-sysroot=${SYSROOT} \
       --with-libevent=builtin --with-yaml-cpp=builtin \
       --with-boost=builtin --disable-shared --libdir=/ \
       --includedir=/include --with-libmaxminddb=builtin \
       --with-jansson=builtin
     make V=0
     echo "Installing library in ${BASEDIR}/build/${ANDROID_TOOLCHAIN}"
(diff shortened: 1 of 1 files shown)


bassosimone:
I've become more and more pissed off by `./configure` verbosity and hence I've recently started silencing it everywhere possible. If you think this is not a good thing to do, I'll revert that specific part of the commit.

mobile/android/scripts/build_target.sh
```diff
@@ -80,13 +80,19 @@ export LDFLAGS="${LDFLAGS} -L${SYSROOT}/usr/lib${LIB_SUFFIX} -L${ANDROID_TOOLCHA
...
     echo "Installing library in ${BASEDIR}/build/${ANDROID_TOOLCHAIN}"
+    make install-strip DESTDIR=${ROOTDIR}/jni/${DESTDIR_NAME}
+    rm -rf ${ROOTDIR}/jni/${DESTDIR_NAME}/include
     make install-data-am DESTDIR=${ROOTDIR}/jni/${DESTDIR_NAME}
-    make install-exec DESTDIR=${ROOTDIR}/jni/${DESTDIR_NAME}
+    rm -rf ${ROOTDIR}/jni/${DESTDIR_NAME}/pkgconfig
+    rm -rf ${ROOTDIR}/jni/${DESTDIR_NAME}/usr
+    rm -rf ${ROOTDIR}/jni/${DESTDIR_NAME}/*.la
+    rm -rf ${ROOTDIR}/jni/${DESTDIR_NAME}/libevent_core.a
+    rm -rf ${ROOTDIR}/jni/${DESTDIR_NAME}/libevent_extra.a
```
bassosimone: I forgot to explain this. Here the idea is:

1) do `install-strip` because otherwise I need to figure out the proper strip command for each architecture and at the first glance this was not straightforward (I noticed multiple strip commands for some architectures)

2) since step 1 installs *all* headers, including the ones installed by dependencies, next I remove all headers and run again the specific subcommand need to install headers only

3) then I remove extra stuff that is installed (pkgconfig, usr), libtool files (not needed), and two libevent libraries that we don't use

Sorry for not specifying this in the commit message. @DavideAllavena do you want me to document this in the source code?

DavideAllavena:
Now this PR is OK.
I think we can merge

mobile/android/scripts/build_target.sh
bassosimone: Alright, I pushed another commit that clarified why the code is written like that.

bassosimone:
Ok, I've merged! Thanks @DavideAllavena for the review!

PR #275, opened 2015-12-03
Create tarballs of the Android build

### LINKED ISSUE #268: Publish v0.1.0-beta.1 binaries for Android (where?)

bassosimone:
This should be merged after #274. It creates compressed tarballs of the build. I have plans to publish these tarballs along with the release so that Apps can integrate them.

Closes #268.

### DIFF
--- a/mobile/android/.gitignore
+++ b/mobile/android/.gitignore
@@ -1,0 +1,1 @@
+measurement_kit-jni-*.tar.bz2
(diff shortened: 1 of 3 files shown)


bassosimone:
Added labels such as "help wanted" and "question" because, before merging this, it's necessary to check whether this is the most optimal way to distribute the compiled code.

I previously assumed using compressed tarballs was the best way, but now I've started to realize that other formats (such as `.aar`) may instead be more appropriate.

bassosimone:
Alright! I have studied a bit more and better understood how to proceed. We need to provide the compiled libraries (as in this pull request) for people that wants to create their own JNI API based on measurement-kit.

But we will also build our own JNI API and the proper way to distribute this is actually to package a .aar file (see for example [how sodium-jni does that](https://github.com/JackWink/sodium-jni)) and distribute it using [jcentral](https://bintray.com/bintray/jcenter). The latter is the default repository where recent Android applications look for dependencies.

In light of what I said above, I now believe it is good to proceed with the solution proposed in this pull request to generate the binaries needed, e.g., to build portolan-jni.

AntonioLangiu:
It looks good to me!

bassosimone:
Ok, thank you @AntonioLangiu, I will proceed to merge this once travis terminates!

PR #278, opened 2015-12-03
Rename library and improve android build process

bassosimone:
- change the name of the library from libmeasurement_kit.la to libmeasurement-kit.la because the name of the repository is measurement-kit and we should only use measurement_kit when the name should match a name that is usable by C++ (i.e. in the include folder to match namespaces)

- update Android build instructions to provide the right level of information (no need to tell too much about the helper scripts when one is supposed to use the wrapper, yet mention them so that one that wants to hack knows what to read)

- create (manually rather than programmatically) a single archive with all the libraries since the libraries are small and a single archive is easier to manager with github releases

Is this OK? Any objections?


AntonioLangiu:
@bassosimone I think you work is good!  I've also fixed the iOS build script and the cocoapod file.
Can you check it?

bassosimone:
@AntonioLangiu thanks, will rebase and merge then! (As regards checking your changes, sure!)

bassosimone:
@AntonioLangiu ah, alright, maybe I will check your changes first and then merge :)

bassosimone:
My change 'libmeasurement_kit.la' => 'libmeasurement-kit.la' prompted a long discussion (#bikeshedding?) with @AntonioLangiu that I'll try to summarize below:

1. using the dash rather than the underscore in the library name is good because it's less surprising to the user that the library name matches more closely the repository name

2. but then we should also rename 'include/measurement_kit' to 'include/measurement-kit' for consistency with the above point

3. but we cannot change the namespace to 'measurement-kit' for consistency because that's not valid C++ (we can keep the underscore, but that's another source of inconsistency)

4. but it's annoying in any case to have to type every time 'measurement_kit' and to write less we should probably just use 'mk' as namespace (IIRC @hellais was in favour of this) -- I'm writing this here because I am already quite pissed off of writing 'measurement_kit' (and also @AntonioLangiu is)

5. while here, let's say that it is also annoying to have the 'common' sub-namespace that basically needs to be written every time, therefore it's probably better to scrap that (other namespaces have the function to segment sub-functionality, i.e. I believe that 'dns::Response' makes sense while 'common::Error' is probably unnecessary complexity)

6. to further reduce confusion, maybe we can also prefix lib to the repository (i.e. libmeasurement-kit) to make the name of the main product of this repository equal to the name of the repository

7. while there, I'd also remove the dash from the library name, because that's often used to split parts of the package name (i.e. variants or the version) and so I'd probably rename the repos and the library as libmeasurementkit.

What do you think @hellais?

bassosimone:
Alright, I think that after failing a first saving throw against bikeshedding now I am again able to discern what is important and what is not important. I'm hence going to close this as (auto?) bikeshedding.

Issue #267, opened 2015-12-02
Test v0.1.0-beta.1 with Portolan small application


bassosimone:
The small application is currently closed source but I have the sources and I can take care of this part of the work (cc: @ValerioLuconi such that you are informed!)

bassosimone:
Done with Nexus 5 running Android 6.

PR #281, opened 2015-12-04
Improve Android build process

bassosimone:
This is same as #278 except that bikeshedding is removed :-)

- - -

- simplify README.md

- document creation of the tarball

- simplify make_archive.sh

- do not call make_archive.sh from build_target.sh

The rationale of these changes is that:

- the compiled libraries are small (around 7 MiB so we can put everything into a single compressed tarball)

- if we do that, it's simpler to sign the tarball and also given how github releases work it's simpler to download a release

### DIFF
--- a/mobile/android/README.md
+++ b/mobile/android/README.md
@@ -25,64 +25,53 @@ Just type the following:
 
     $ brew install android-ndk
 
-## Creating a custom toolchain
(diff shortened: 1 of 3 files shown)


bassosimone:
This is needed to move forward with #267, so I merged it.

PR #282, opened 2015-12-04
Rename and simplify namespaces (issue #284)

### LINKED ISSUE #284: Rename and simplify namespaces
This issue is based on feedback provided by @AntonioLangiu and me after using the library for quite some time now. What we basically noticed is that:

1) the `measurement_kit` namespace is too long to write and, if used in code, would lead to very long lines; therefore, in most places one does `using namespace measurement_kit`. That's not optimal (@AntonioLangiu convincingly says "imagine if the standard library was under the `standard::` namespace, would you still use the namespace name inline, or would you instead `use namespace standard` everytime?"). To say this in a general way, a short namespace name allows one to use that namespace (as in `std::string`) without pulling names from such namespace into the global namespace.

2) the `measurement_kit::common` namespace is annoying because basic objects are in such namespace. Quoting again @AntonioLangiu: "even to declare there is an error you have to write `common::Error` and that's super annoying and again you are tempted to write `using namespace common`". In a general way, stuff in common is probably so common that either its namespace name should be something like `cmn` or it should be no namespace for that stuff.

At the cost of breaking symmetry here and there, the most obvious way to increase usability of the library is the following:

a) rename the outer namespace of the library to be `mk`: this is very short and thus people is likely to write stuff such as `mk::dns::resolve(...)`

b) remove the `common` namespace such that to declare an error one just needs to write `mk::Error` (if outside the library) and `Error` (if inside the library)

There is already an open pull request, #282, that deals with this issue.

bassosimone:
- Use `mk` as namespace rather than `measurement_kit`. On this there is agreement between me, @hellais (who originally prodded the idea), and @AntonioLangiu.

- Zap the common namespace. Prodded by @AntonioLangiu with me in favour and @hellais agnostic with respect to this. Removing the common namespace is good because stuff in such namespace is nearly always pulled and having the namespace forces you either to do `using namespace...` or to write (too many) times`common::<something>`.

Closes #284.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,10 +195,9 @@ of the same thread that called *run*).
 
 ```C++
 #include <measurement_kit/ooni.hpp>
-using namespace measurement_kit;
 
 // Run sync test
-ooni::HttpInvalidRequestLineTest()
+mk::ooni::HttpInvalidRequestLineTest()
     .set_backend("http://127.0.0.1/")
     .set_verbose()
     .on_log([](const char *s) {
@@ -216,7 +215,7 @@ the callback passed as argument to *run* is invoked when the test completed.
 
 ```C++
 // Run async test
-ooni::HttpInvalidRequestLineTest()
+mk::ooni::HttpInvalidRequestLineTest()
     .set_backend("http://127.0.0.1/")
     .set_verbose()
     .on_log([](const char *s) {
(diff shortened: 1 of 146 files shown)


bassosimone:
This is not ready to be merged yet. Before moving forward I'd like to remove `using namespace xx` from more places in the documentation and in the unit test.

bassosimone:
I've updated the documentation to always list namespaces rather than use `using namespace`. It is not worth the effort to do the same with the unit tests, since they are `.cpp` only files where it is OK to use the most handy `using statement` given the circumstances (i.e. the libs you use).

bassosimone:
Going to merge this, since there was rough consensus with respect to all the changes implemented in this pull request and since this is just a mechanical set of changes.

PR #291, opened 2015-12-07
Make the class connection non copyable and non movable

### LINKED ISSUE #280: The class connection should be not copyable and not movable
As wrote [here](https://github.com/measurement-kit/measurement-kit/blob/master/src/net/connection.cpp#L99) the class `connection` should be not copyable and not movable but I didn't' find an assignment in the class to do this (neither in the base class it extends).

AntonioLangiu:
Fix little bug in the class Connection. 

This closes #280

### DIFF
--- a/src/net/connection.hpp
+++ b/src/net/connection.hpp
@@ -31,7 +31,7 @@
 namespace mk {
 namespace net {
 
-class Connection : public Emitter {
+class Connection : public Emitter, public NonMovable, public NonCopyable {
   private:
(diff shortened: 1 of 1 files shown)


bassosimone:
@AntonioLangiu please run `git config --global user.email $your_email` because at the moment I continue to get patches from you with an invalid email

bassosimone:
Going to merge! *Bravo* @AntonioLangiu for spotting this!

PR #294, opened 2015-12-09
Teach tests to get boolean value for set_verbose

### LINKED ISSUE #293: Fix set_verbose() in new OONI API
[New OONI API](https://github.com/measurement-kit/measurement-kit/tree/master/include/measurement_kit/ooni): in many cases it is useful to always call set_verbose() and pass it a boolean specifying the verbosity level. This is useful, for example, when you write a command line app based on MK.

I'd also add that, if not argument is specified, the default value of the argument is true.

AntonioLangiu:
Teach ooni new API to get boolean value for set_verbose(); 

This closes #293


bassosimone:
Thank you! Merging! :+1:

PR #295, opened 2015-12-09
mobile/android/jni: move .gitignore down one level

bassosimone:
Much better because otherwise we put the .gitignore that ignores everything in jni inside the tarball, and this is annoying when we compile jniLibs because the .gitignore is extracted into a folder than in the jniLibs repository contains non-ignored stuff.


bassosimone:
Self-merging this hotfix.

PR #296, opened 2015-12-10
Issue 276

### LINKED ISSUE #276: Make sure we do not throw std::runtime_error
Since we have introduced [specific errors for http problems](https://github.com/measurement-kit/measurement-kit/blob/master/include/measurement_kit/http/error.hpp), in the implementation of  [request_serializer](https://github.com/measurement-kit/measurement-kit/blob/master/src/http/request_serializer.cpp#L31)  we should use those errors instead of `std::runtime_error`

AntonioLangiu:
I've introduced the new errors like @bassosimone said in #276 
The new errors are:  

- UrlParserError
- MissingUrlSchemaError
- MissingUrlHostError

Closes #276.

### DIFF
--- a/include/measurement_kit/http/error.hpp
+++ b/include/measurement_kit/http/error.hpp
@@ -22,6 +22,25 @@ class ParserError : public Error {
     ParserError() : Error(3001, "unknown_error 3001") {};
 };
 
+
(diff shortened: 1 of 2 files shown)


bassosimone:
Thanks, merging!

PR #302, opened 2015-12-14
Update Podspec 

AntonioLangiu:
Update the podspec file. Changes:

- Add all authors
- Improve version (and tag) control
- Remove branch specification because is not needed 
- Change prepare command, since now is possible to build from every folder

Thanks to @bassosimone for the given tips.


bassosimone:
Thanks!

PR #317, opened 2015-12-16
Allow to specify report file path

bassosimone:
Helps to implement the mobile application, especially Android. <strike>.**Before this is merged**, I'd like to **add a couple of regress tests** to make sure it does not break anything.</strike> [Now regress tests have been added!]


bassosimone:
Going to self-merge this hotfix, given that regress tests have now been added!

Issue #191, opened 2015-10-07
memory leak in test/net/connection.cpp:78


bassosimone:
```
==30704== 
==30704== HEAP SUMMARY:
==30704==     in use at exit: 824 bytes in 4 blocks
==30704==   total heap usage: 695 allocs, 691 frees, 46,139 bytes allocated
==30704== 
==30704== 824 (496 direct, 328 indirect) bytes in 1 blocks are definitely lost in loss record 4 of 4
==30704==    at 0x4C2DC90: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==30704==    by 0x44FF10: bufferevent_socket_new (bufferevent_sock.c:328)
==30704==    by 0x43D4C3: operator() (functional:2439)
==30704==    by 0x43D4C3: make (bufferevent.hpp:39)
==30704==    by 0x43D4C3: measurement_kit::net::Connection::Connection(char const*, char const*, char const*, measurement_kit::common::Poller*, measurement_kit::common::Logger*, int) (connection.cpp:73)
==30704==    by 0x4111BB: Connection (connection.hpp:70)
==30704==    by 0x4111BB: ____C_A_T_C_H____T_E_S_T____73() (connection.cpp:78)
==30704==    by 0x4355A4: invoke (catch.hpp:5507)
==30704==    by 0x4355A4: invoke (catch.hpp:6389)
==30704==    by 0x4355A4: runCurrentTest (catch.hpp:5131)
==30704==    by 0x4355A4: runTest (catch.hpp:5001)
==30704==    by 0x4355A4: Catch::Runner::runTests() (catch.hpp:5275)
==30704==    by 0x436201: Catch::Session::run() (catch.hpp:5395)
==30704==    by 0x405D8E: run (catch.hpp:5378)
==30704==    by 0x405D8E: main (catch.hpp:8819)
==30704== 
==30704== LEAK SUMMARY:
==30704==    definitely lost: 496 bytes in 1 blocks
==30704==    indirectly lost: 328 bytes in 3 blocks
==30704==      possibly lost: 0 bytes in 0 blocks
==30704==    still reachable: 0 bytes in 0 blocks
==30704==         suppressed: 0 bytes in 0 blocks
==30704== 
==30704== For counts of detected and suppressed errors, rerun with: -v
==30704== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

bassosimone:
[...]

PR #319, opened 2015-12-20
Run regress tests through Valgrind

### LINKED ISSUE #99: Automatically run regress tests through Valgrind
It should be useful to run regress tests through Valgrind. To this end, we should create a script that compiles libight using `./configure --disable-shared` and than run each regress test script using `valgrind --error-exitcode=1` so that, if there is a memory error or a leak, `make check` fails.

AntonioLangiu:
Using the variable `TESTS_ENVIRONMENT = valgrind` in Makefile.am we are now able to run regress tests through Valgrind. 

To enable it you should call ./configure with --enable-valgrind and --disable-shared.

To merge this PR we need first to fix the memory leak problems of 3 tests:

* [test/common/delayed_call](https://travis-ci.org/measurement-kit/measurement-kit/jobs/97945929#L1185)
* [test/net/connection](https://github.com/measurement-kit/measurement-kit/issues/191)
* [test/http/stream](https://travis-ci.org/measurement-kit/measurement-kit/jobs/97945929#L1509)

This closes #99

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ before_install:
   - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 90
   - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 90
   - sudo update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-4.8 90
-  - sudo apt-get install tor
+  - sudo apt-get install tor valgrind
   - pip install --user cpp-coveralls
 
 install: autoreconf -i
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,17 +16,9 @@ install: autoreconf -i
 # convince coveralls to find all sources when using gcc-4.8 and gcov-4.8
 # (For more context see measurement-kit/measurement-kit#255)
 script:
-  - if [ "$CXX" = "clang++" ]; then
-      ./configure -q --with-libevent=builtin --with-yaml-cpp=builtin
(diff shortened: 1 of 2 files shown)


bassosimone:
Are those errors also reproducible locally?

bassosimone:
What I find annoying of [test/http/stream](https://travis-ci.org/measurement-kit/measurement-kit/jobs/97945929#L1509) is that line numbers are not visible in the stacktrace; however, from the logs it seems that the failing test is on [line 60 of test/http/stream.cpp](https://github.com/measurement-kit/measurement-kit/blob/master/test/http/stream.cpp#L60).

bassosimone:
Looking at test/http/stream.cpp around line 60, I noticed that [stream saves a reference to itself inside a lambda that it owns.](https://github.com/measurement-kit/measurement-kit/blob/master/test/http/stream.cpp#L77). This, as we have seen, creates a reference loop and should prevent the destruction of the stream. If I'm right, adding `stream->on_connect(nullptr);` at end of test could plug the leak.

bassosimone:
Uhm, second reading the code it seems that [emit_error(NoError())](https://github.com/measurement-kit/measurement-kit/blob/master/test/http/stream.cpp#L88) does not cause `stream->on_end()` to be called. We should check whether this is the case. Also, we should reflect on whether this test is still relevant after recent changes made in HTTP code (the pattern of deleting stuff when an error happens is not good and more modern measurement-kit code actually does not use it).

AntonioLangiu:
I will open a new PR to be more clean and readable.

PR #334, opened 2016-01-26
http_parser initializer functions used instead of memset.

### LINKED ISSUE #292: http: use initializers from http-parser
When I merged http-parser v2.6.0, I noticed that they introduced initializer functions for the http-parser and the url-parser objects. Those should be used, rather than memset(), for correctness.

DavideAllavena:
`http_parser_init` and `http_parser_settings_init` are used instead of `memset`.

Closes #292


bassosimone:
Thanks! For this pull request to be complete, we shall also update the usage of [http_parser_url](https://github.com/measurement-kit/measurement-kit/blob/master/src/http/request_serializer.cpp#L23) to use the new initializer function provided by http-parser.

PR #349, opened 2016-02-01
traceroute: allow to use custom logger

### LINKED ISSUE #297: traceroute: pass custom logger to AndroidProber
The `AndroidProber()` constructor should be modified to receive a custom logger as argument, set by default to the default logger. This would enable to redirect the output of a traceroute scan to some place where it could then be processed separately from other logs.

DavideAllavena:
Custom logger added. Option 1 chosen.

Closes #297


bassosimone:
Thanks!

PR #360, opened 2016-02-04
Consolidate HTTP implementation

bassosimone:
The interface exposed by our HTTP implementation is tiny: just around 100 lines, so move have everything in the same file.

Join all bits of documentation to have a single markdown file.


DavideAllavena:
I checked the PR. I think it is ok. Building is ok even on my system.

bassosimone:
ACK, thank you! I will take care of merging this soon.

PR #362, opened 2016-02-07
Issue 355: increase abstraction of HTTP code

### LINKED ISSUE #355: Increase abstraction of HTTP code
We should add a new, more abstract piece of API for HTTP that allows one to issue HTTP requests without worrying about the life cycle of `http::Client`. The prototype for the new function shall be the following (slightly different from `http::Client::request()` because it has more default arguments):

```C++
void http::request(Settings settings,
    http::RequestCallback callback,
    http::Headers headers = {},
    std::string body = "",
    Logger *logger = Logger::global(),
    Var<http::Client> client = http::Client::global()
);
```

Steps to do this:

- [ ] implement `http::Client::global()` to retrieve the global `Var<http::Client>`
- [ ] implement `http::request()`

bassosimone:
Closes #355.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,6 +114,7 @@ __venv__
 example/ooni/dns_injection
 example/ooni/http_invalid_request_line
 example/ooni/tcp_connect
+example/http/request
 
 # files being edited
 *.swp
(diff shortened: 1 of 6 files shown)


DavideAllavena:
PR is OK. Checked and tested. Waiting for travis to finish to merge.

PR #363, opened 2016-02-07
Issue 333: implement basic mlabns client

### LINKED ISSUE #333: Implement basic mlabns client
We can improve it later, but I think we can start writing a basic client using:

- the already existing http only client (https could be added later)

- yaml-cpp to parse the response (we could switch to use jansson later)

bassosimone:
Closes #333.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,6 +111,7 @@ site
 __venv__
 
 # example
+example/mlabns/query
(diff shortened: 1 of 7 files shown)


DavideAllavena:
PR is OK. Checked and tested.

bassosimone:
Thank you!

PR #365, opened 2016-02-08
Start work to download dependencies on demand

bassosimone:


bassosimone:
Closing because #366 seems to be a better (as in less code, less coupling) approach.

PR #366, opened 2016-02-09
Use M4 macros and a build script for dependencies

bassosimone:
This allows to download, compile, and install dependencies through `./build/dependency all`.

Otherwise, the `./configure` script will stop when a dependency is missing and explain you how to install that dependency. Instructions explain how to use `apt-get`, `brew`, and `./build/dependency`.

The latter downloads, patches, configures, compiles, and installs a dependency inside the `./builtin` directory. To teach `./configure` to spot dependency `foo` installed like this, one shall pass the `--with-foo=builtin` flag to the `./configure` script.

This is an enhancement because we need to track less code (i.e., faster pulls) and because it disentangles building the core of MK from its dependencies. I tried also other approaches (see, e.g., #365; the end result however was always that I needed to couple tightly dependencies and MK and that more code than in this pull request needed to be added).

Furthermore, the `configure.ac` file has been refactored moving all the hand-rolled tests inside `m4/measurement_kit.m4` for increased readability, testability, and maintainability.

Contains also updates in the code to cross compile stuff for ios and android.

Code in this pull request is a super stripped-down version of [bassosimone/mkpm-ng](https://github.com/bassosimone/mkpm-ng).

### DIFF
--- a/build/android/scripts/build_target.sh
+++ b/build/android/scripts/build_target.sh
@@ -71,30 +71,38 @@ export RANLIB=${TOOL_PATH}-ranlib
 export STRIP=${TOOL_PATH}-strip
 
 export CPPFLAGS="${CPPFLAGS} --sysroot=${SYSROOT} -I${SYSROOT}/usr/include -I${ANDROID_TOOLCHAIN}/include"
-export LDFLAGS="${LDFLAGS} -L${SYSROOT}/usr/lib${LIB_SUFFIX} -L${ANDROID_TOOLCHAIN}/lib -lc++_static -latomic -lm"
+export LDFLAGS="${LDFLAGS} --sysroot=${SYSROOT} -L${SYSROOT}/usr/lib${LIB_SUFFIX} -L${ANDROID_TOOLCHAIN}/lib"
+export LIBS="${LIBS} -lc++_static -latomic -lm"
+
+export pkg_configure_flags="--host=${ARCH} --with-sysroot=${SYSROOT} --disable-shared"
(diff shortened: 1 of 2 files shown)


bassosimone:
Putting on hold again. Things that do not convince me:

- [ ] is it OK to use -fPIC compiled code in http-parser?
- [ ] probably things should work if CMake is invoked twice (once for static and once for shared)
- [ ] the pull request would have been simpler to reivew without that ton of removals

bassosimone:
Now that I have solved the above mentioned things that did not convince me, the following needs to be done before entering again in "ready for review" state:

- [ ] make sure that everything still compiles under OSX
- [ ] make sure that everything still compiles under ubuntu
- [ ] make sure that everything still compiles under travis
- [ ] make sure (with help from another branch) that we can still cross compile for ios
- [ ] make sure (with help from another branch) that we can still cross compile for android

Once all the above are ticket, I'll move this again into "ready for review" state.

bassosimone:
Added commit to adapt cross compile strip. It actually needs some more love, specifically:

- [ ] re-add and test code designed to minimize archive size for Android
- [ ] same as above, but for iOS

Until this is done the pull request is again on hold. Meanwhile, closing #368, because its changes have now been incorporated into this pull request.

bassosimone:
Hey, @DavideAllavena, could you check whether d54ecf8 is enough, according to what we were discussing today before I had to dash, to satisfy the needs of power users?

bassosimone:
Ok, now I think we can move again to "ready for review" state!

bassosimone:
Moved this to milestone 0.2.0. Basically, there are some little bits of goodies in this branch that I'd like to merge into master before next release because they contain improvements.

bassosimone:
Closing this pull request as too much time has passed and it would add little value to do the effort to merge the small changes in here into master for such a little gain.

PR #367, opened 2016-02-09
Removal of "unused parameter" warning in traceroute

### LINKED ISSUE #358: traceroute: fix `unused parameter: logger` warning
```
  CXX      src/traceroute/android.lo
src/traceroute/android.cpp:68:71: warning: unused parameter 'logger' [-Wunused-parameter]
 AndroidProber::AndroidProber(bool a, int port, event_base *c, Logger *logger)
```

This basically means that, even though the code is ready to use a custom logger, the default logger is always used because the custom logger parameter passed to the constructor is not used to initialize the logger attribute. Hence I've reopened issue #297.

Closes #297.

DavideAllavena:
This PR removes the warning appearing during compilation appeared after merge  of 71064b1fc04.

Closes  #358.


bassosimone:
Thanks, will merge it soon!

PR #372, opened 2016-02-12
Version number updated from 0.1.0 to 0.2.0-alpha

### LINKED ISSUE #359: define procedure to update all version numbers throughout the code
I've just noticed `<measurement_kit/common/version.hpp>` is out of date. Perhaps there are also other outdated version numbers around. Should track all of them and find a way to fix them all at once.

DavideAllavena:
This PR updates version number.

The patch can be used to locate version number in the project (al least where i found it). 

An exaustive process is still not defined. An update-script or a single file with the version maybe?

Closes #359.


bassosimone:
I think we should also check whether Makefile.am and/or configure.ac needs an update

bassosimone:
Removing the tag `question` since I think we've discussed this in person at length, and we've settled to the conclusion that we do not make ABI guarantees (i.e., `-version-info 0`) and we embed the release version number inside the library name (i.e. `-release 0.1.3`).

PR #374, opened 2016-02-12
Consolidate dns implementation.

DavideAllavena:
Similarly to what has happend to the http implentation all the
headers belonging to che dns implementation are grouped in a file.

Same has happend fos the DNS documentation.


doc/api/dns.md
```diff
@@ -1,5 +1,5 @@
 # NAME
-Resolver -- DNS resolver
+mk::dns -- MeasurementKit HTTP library
```
bassosimone: Beware that this is the DNS!

doc/api/dns.md
```diff
@@ -7,31 +7,13 @@ MeasurementKit (libmeasurement_kit, -lmeasurement_kit).
...
-
-// Issue an async DNS query
-reso.query(
-        "IN",                                             // Domain of the query
-        "AAAA",                                           // Type of query
-        "nexa.polito.it",                                 // Name to resolve
-        [](mk::Error error, mk::dns::Response response) { // Callback
-            if (error) throw error;
-            // handle successful response
-        });
```
bassosimone: I'd not remove this part of the synopsis. See how HTTP documentation now looks like.

bassosimone:
Thank you, @DavideAllavena! I'll assign myself this pull request and make sure it is merged soon

doc/api/dns.md
```diff
@@ -73,14 +76,59 @@ The difference between `REVERSE_A`, `REVERSE_AAAA` and `PTR` is that
...
+- `ShutdownError`:  evdns library was shut down
+- `CancelError`:  user cancelled query
+- `NoDataError`:  no data in the response
+
+
+The `Response` class holds the result of an async DNS `Query`. You do not
+typically instantiate a `Response` yourself, rather you receive an instance
+of `Response` in response to a DNS query.
+
+Given a DNS response data can be retrived as it follows:
```
bassosimone: I think this part can be moved into the synopsis

doc/api/dns.md
```diff
@@ -73,14 +76,59 @@ The difference between `REVERSE_A`, `REVERSE_AAAA` and `PTR` is that
 an IPv6 address, while for `PTR` you need to construct yourself the
```
bassosimone: Nit: just a little above here, I'd amend the parenthesis talking about `QueryTypeId::A` such that the end of it reads `QueryTypeId::A, QueryTypeId::AAAA`. I'd do this to suggest that for each query as a string there is a corresponding type that is represented by a class.

doc/api/dns/response.md
```diff
@@ -1,65 +0,0 @@
...
-}
-```
-
-# DESCRIPTION
-
-The `Response` class holds the result of an async DNS `Query`. You do not
-typically instantiate a `Response` yourself, rather you receive an instance
-of `Response` in response to a DNS query.
-
-# BUGS
```
bassosimone: I think this section should be moved in the new, consolidated document

doc/api/dns/response.md
```diff
@@ -1,65 +0,0 @@
...
-
-Since `evnds` does not report whether the response was authoritative, the
-`is_authoritative()` method always returns `"unknown"`.
-
-# SEE ALSO
```
bassosimone: Also this section

bassosimone:
Overall this looks very good to me. I think the following small changes need to be added to the documentation and then we're ready to merge!

- [ ] move usage example of response in the synopsis
- [ ] add little detail in parenthesis to indicate query-type classes names
- [ ] move `BUGS` and `SEE ALSO` of response.md to new file

doc/api/dns.md
DavideAllavena: I did this taking [`doc/api/http.md`](https://github.com/measurement-kit/measurement-kit/blob/master/doc/api/http.md) as a template. Shoul we change also that file?

bassosimone: Ah, right. Well probably in that case it fits where it is because the structure is explicitly listed (a thing that we cannot move in the synopsis) while here the focus is on usage (since the synopsis shows usage, I was proposing to move that usage-related part into the synopsis).

bassosimone:
Thanks, merging!

PR #370, opened 2016-02-11
build system: enable coverage in a better way

bassosimone:
This speeds up the development process because one does not need to update all the `include.am` files anymore. Now the list of things to compile (including tests and examples) is generated by `autogen.sh` which becomes the first step to run (instead of `autoreconf -i`).

Travis started failing after I implemented this. After losing some time with the existing travis config file to no avail, I decided to import a better travis config from [bassosimone/mkok-experimental-tor](https://github.com/bassosimone/mkok-experimental-tor).

PR #380, opened 2016-02-18
Improvements to build system

bassosimone:
- implement each build step in a separate file
- allow dependencies to override specific steps
- use above feature for boost and catch


bassosimone:
Moved this to milestone 0.2.0. Basically, there are some little bits of goodies in this branch that I'd like to merge into master before next release because they contain improvements.

bassosimone:
This is like #370 but it also splits the `./build/dependency` script in subfiles. This is not particularly interesting and so I think we can close this pull request.

PR #381, opened 2016-02-19
Tor poller and controller

bassosimone:
This pull request contains the following enhancements:
- builds tor (actually mkok-onion-event, our modified tor) as an MK dependency
- creates a "tor" poller that behaves similarly to an ordinary poller (with few exceptions)
- adds code to talk with tor for common tasks that we'll need


bassosimone:
This has become too old, at this point, to be useful. Closing.

PR #383, opened 2016-02-19
Download dependencies in autogen.sh

bassosimone:
@AntonioLangiu is testing this, and I'm keeping track of this progress here:

- [x] test on linux
- [x] test on travis
- [x] test on macosx
- [x] cross compile for android and test app
- [x] cross compile for ios and test app


AntonioLangiu:
I checked the compilation for Os X and iOS, and it works.

PR #385, opened 2016-02-19
Implement Settings class as map<std::string, genericType>

AntonioLangiu:


bassosimone:
Thanks! I will take care of this either tomorrow or during next week!

bassosimone:
I think it would be interesting to also have automatic conversion from settings to any other type. I am also wondering whether it wouldn't be more rational to just import into base [nlohmann/json](https://github.com/nlohmann/json) and use that for settings. What do you guys think? @AntonioLangiu @hellais

bassosimone:
I just rebased the branch and now I'm going to review.

bassosimone:
I discussed whether to use a json library for this or not with @hellais when he was here in Turin, and we concluded it was not wise to depend onto an external library for such functionality.

Speaking of moving forward, @AntonioLangiu, I'd like to improve the correctness of the code as far as the conversion from string to numbers is concerned. What I'd like to see it implemented here is a mechanism by which, if you attempt to convert to integer but we cannot, an exception is raised. To do this task, for symmetry with Python, we could for example raise a `ValueError`. Same should happen for converting to double and to float (and to any other non string, scalar, primitive type).

To perform error checking, I'd take advantage of the `failbit` etc features of a stringstream, which should be enough for us to understand whether conversion worked or not (it would be wise, of course, to add tests with wrong input formats to make sure it works).

include/measurement_kit/common/error.hpp
```diff
@@ -58,5 +58,11 @@ class MaybeNotInitializedError : public Error {
     MaybeNotInitializedError() : Error(2, "unknown_failure 2") {}
 };
 
+// Value error
```
bassosimone: Here it should be `/// Value error` (i.e. three slashes, such that the comment is picked up by doxygen)

include/measurement_kit/common/error.hpp
```diff
@@ -58,5 +58,11 @@ class MaybeNotInitializedError : public Error {
...
+class ValueError : public Error {
+    public:
+        ValueError() : Error(3, "unknown_failure 3") {}
+};
```
bassosimone: Could you indent this correctly using `clang-format -i` please?

include/measurement_kit/common/settings.hpp
```diff
@@ -6,10 +6,46 @@
...
+        _store = ss.str();
+    }
+
+    template <typename T>
+    T as () const {
+        std::stringstream ss;
+        ss.str(_store);
+        T t;
+        ss >> t;
+        if (!ss.eof()) {
```
bassosimone: Could you add a comment above this line explaining the rationale of checking both `eof()` and `fail()`?

include/measurement_kit/common/settings.hpp
```diff
@@ -6,10 +6,46 @@
...
+        return _store != "";
+    }
+ private:
+    std::string _store;
+};
+
+typedef std::map<std::string, Item> Settings;
 
 }
 #endif
```
bassosimone: Could you reformat this file using `clang-format` please?

test/common/settings.cpp
```diff
@@ -0,0 +1,118 @@
...
+    settings["key"] = 6;
+    REQUIRE(settings["key"].as<std::string>() == "6");
+}
+
+TEST_CASE(
+    "The settings class throws an exception when converting int (10) to bool") {
+    Settings settings;
+    settings["key"] = 10;
+    REQUIRE_THROWS_AS(settings["key"].as<bool>(), ValueError);
+}
```
bassosimone: This is suprising: I would have expected the conversion to succeed and return `true`.

test/common/settings.cpp
```diff
@@ -0,0 +1,118 @@
...
+    settings["key"] = "1";
+    REQUIRE(settings["key"].as<bool>() == true);
+}
+
+TEST_CASE("The settings class throws an exception when converting from string"
+          "to bool with an uncorrect value") {
+    Settings settings;
+    settings["key"] = "10";
+    REQUIRE_THROWS_AS(settings["key"].as<bool>(), ValueError);
+}
```
bassosimone: Also here I think that the conversion to bool is strange.

src/dns/resolver.cpp
```diff
@@ -88,7 +88,7 @@ evdns_base *Resolver::get_evdns_base(void) {
     // By default we don't randomize the query's case
     std::string randomiz{"0"};
     if (settings.find("randomize_case") != settings.end()) {
-        randomiz = settings["randomize_case"];
+        randomiz = settings["randomize_case"].as<std::string>();
```
bassosimone: In this case, I think it would be nice for the conversion to happen automatically. Here `randomiz` is clearly a string, so I'd expect that, if you attempt to assign `Item` to string it would either work or throw exception. Does it make sense?

src/http/request.hpp
```diff
@@ -98,7 +98,7 @@ class Request : public NonCopyable, public NonMovable {
             // tor_socks_port takes precedence because it's more specific
             if (settings.find("tor_socks_port") != settings.end()) {
                 std::string proxy = "127.0.0.1:";
-                proxy += settings["tor_socks_port"];
+                proxy += settings["tor_socks_port"].as<std::string>();
```
bassosimone: Also here I think we can rely on type deduction and allow to assign `Item` to `std::string`.

src/net/socks5.cpp
```diff
@@ -10,14 +10,15 @@ namespace net {
...
+           settings["socks5_address"].as<std::string>().c_str(),
+           settings["socks5_port"].as<std::string>().c_str(), lp, poller),
+      proxy_address(settings["socks5_address"].as<std::string>()),
+      proxy_port(settings["socks5_port"].as<std::string>()) {
 
     logger->debug("socks5: connecting to Tor at %s:%s",
-                  settings["socks5_address"].c_str(),
-                  settings["socks5_port"].c_str());
+                  settings["socks5_address"].as<std::string>().c_str(),
+                  settings["socks5_port"].as<std::string>().c_str());
```
bassosimone: The use case whether one wants to convert a `Item` to the corresponding C string is quite common. Since we're in C++ and in C++ it's custom not to automatically convert `std::string` to `const char *`, I'd say it would be good to have a `Item::c_str()` method that just takes the `c_str()` of the internal string and returns it.

Does that makes sense?

bassosimone:
Could you also please update the documentation (or create documentation if it is missing)?

I also think that we should reflect a bit more about the behavior of `as<>()`. More specifically, I believe it is fine to have `as<>()` throw `ValueError()` on error.

But, in addition to that, I am wondering also about a special way of accessing the content of `Item` that does not throw but rather signals the error in some way. On top of my head I'd say "let's use `Maybe` here, except that it is true that the current implementation of `Maybe` is not so practical. In any case here I'd like to have a go-like functionality where you easily return `err, value` from a function and are easily able to manage `err` and `value` as you wish. Which is similar to what `Maybe` does except that `Maybe` is *for sure* annoying to use. Another approach could be `std::tie<>` but it's not much better.

bassosimone:
@AntonioLangiu, I've tried telling the `GenericType` class (now renamed `SettingsEntry` because it's probably too bombastic to pretend it is actually a generic type) to inherit publicly from `std::string`. With the code structured this way, in most cases we don't need to apply changes, because `SettingsEntry` could be used in any place where a `std::string` is required (for this to work we must be careful to avoid object slicing). I am not sure I am satisfied with the end result, though.

test/common/settings.cpp
bassosimone: Ok, this is less surprising than I thought. Also `boost::lexical_cast<T>()` has the same limitation that you cannot convert `10` to `true`. So, my comment above could be disregarded.

test/common/settings.cpp
bassosimone: See above. My comment could be disregarded.

bassosimone:
Okay, now I'm more satisfied by this PR. I've seen that boost's lexical_cast has the same problem that we had (i.e. `10` is not `true` but rather a value error), so we were not doing that bad. Plus, I've figured out how to serialize boolean values as `true` and `false`, which removed all possible ambiguities.

bassosimone:
Uhm, I just noticed that "false" serialized as string does not correctly serialize as bool on Linux both using clang++ and g++ and have no idea why (on MacOS it worked with clang++).

I really don't understand why, so I decided to not apply the diff that added this feature to the current pull request and bless the behavior where "1" is `true`, "0" is `false` and other numbers are invalid.

I will merge this pull request as soon as travis terminates successfully.

PR #386, opened 2016-02-19
build: autogenerate list of files to build

bassosimone:
For now, do not autogenerate however the list of files to build under `src/ext` because that requires more work.

In fact, we need to avoid compiling test cases and other files and hence that is less trivial than the rest.


configure.ac
```diff
@@ -15,6 +15,8 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES])
 LT_INIT
 AC_CONFIG_HEADERS([config.h])
 
+AM_CONDITIONAL([BUILD_TESTS], [test "yes" = "yes"])  # for now, build always
```
AntonioLangiu: This is not necessary.

autogen.sh
```diff
@@ -2,6 +2,65 @@
...
+        if [ -d $1/$name ]; then
+            gen_headers $1/$name
+        fi
+    done
+}
+
+gen_sources() {
+    for name in `ls $2`; do
+        if [ ! -d $2/$name ]; then
+            if echo $name | grep -q '\.c[pp]*$'; then
```
AntonioLangiu: This regex can be more precise.

PR #390, opened 2016-02-21
async: radically simplify implementation

bassosimone:
The only visible change in headers is that it is optionally possible to pass to the async constructor the specific poller to use. This, in turn, will allow us to pass to the async object the tor poller (which will be added as part of a *subsequent* pull request).

Speaking of the implementation, the radically simpler changes stem from the observation that the older implementation was basically doing the following:

1. schedule functions registered to be called (either for starting tests or for dispatching the news that they completed) making sure lists of functions to be called are accessed in a thread safe way

2. dispatch I/O events by calling `event_base_loop()` in a non-blocking way

That's basically re-implementing part of the functionality of `event_base_dispatch()` out of libevent, because libevent's event loop is thread safe and has indeed the capability of registering functions to be called at a later time. The only problem, before the recent `Poller::call_soon()` patch, was that using such feature of scheduling callback was not that practical in MeasurementKit.

Hence, for what was said above, we can perform the following 1:1 changes to the implementation:

1. re-implement `loop_thread` as `poller->loop()` (i.e., a blocking I/O loop that is handled by libevent in a thread safe way and that also schedules registered callbacks)

2. directly start the thread inside the constructor and never actually exit it, unless explicitly asked by the user (this changes behavior with respect to previous implementation, we discuss this below)

3. re-implement `run_test()` as a chain of lambdas where the external lambda (the one to insert the "job" into async) and the most internal lambda (the one to extract it and call the final callback) are implemented through `poller->call_soon()`

4. re-implement `empty()` by checking for the value of an atomic counter that is incremented and decremented properly by `run_test()`.

Now, as regards the discussion of the changes, let's start to discuss 2. The previous implementation took advantage of the possibility of entering/exit libevent's loop to stop the background thread when it was not used. We are about to add support to the tor event loop, which will be wrapped by the poller class interface, and where it is not possible to exit and then reenter the loop. To keep things simple, I have hence decided to make the code in here equal for all kinds of poller. Of course, all of this matters if we are not running tests; because, if we're running tests, it goes without saying that we will have a background thread active and running them. So, if we're not running tests and we have the traditional libevent poller, we will have a thread blocked in `event_base_loop()`, that is in a blocking system call, until new event arrives. This is to say that we are only consuming some stack. On the contrary, with a tor poller, we will have a thread *mostly* blocked by a blocking system call, except that once per second the periodic function calls of tor are called. And, given that by default the tor poller runs with `DisableNetwork=1`, this again means very light and very periodic activity. For all this reasons, I've decided not to bother to optimize things further (given than now optimization is more complex). Should the need arise in future, I am totally open create - maybe - a fire-off-one-time test function using the libevent poller to just run one/a few tests (maybe this would be optimal for apps not using tor). In any case, I think this aspect should be created as an issue tagged as discussion.

As regards the correctness of the code, instead:

1. `poller->call_soon()` creates a `new std::function<void()>` on the heap, then passes such pointer as the `void *` argument to `event_base_once()`

2. given that `EvThreadSingleton::ensure()` is called when the thread is started, libevent is initialized to use threads, and `event_base_once()` should ensure that `event_base` internals are locked before inserting the new "once" event in queue

3. the pointer allocated on the heap in 1. will then be freed, after the associated lambda has been called

4. this happens in another thread, however, this was also the case before

In any case, to ensure that we're not making any mistake, we shall check with Valgrind that there are neither memory leaks nor memory errors in the code.

- [ ] create issue describing the always-running measurement-thread as a place where to discuss whether the fact that we use such thread is really an issue or not

- [x] make sure there are no issues running Valgrind (no memory errors and/or leaks)

- [x] make sure that travis completes correctly

- [x] have the code in here reviewed by another developer

- [x] move changes fixing leaks in a separate pull request so that they could be tested independently (as explained in more detail below)

- [ ] merge the pull request

### DIFF
--- a/include/measurement_kit/common/async.hpp
+++ b/include/measurement_kit/common/async.hpp
@@ -21,6 +21,8 @@ class Async {
     /// Default constructor
     Async(Poller *pol = Poller::global());
 
+    ~Async();
+
     /// Run the specified network test and call a callback when done
     /// \param func Callback called when test is done
     /// \warn The callback is called from a background thread
--- a/include/measurement_kit/common/async.hpp
+++ b/include/measurement_kit/common/async.hpp
@@ -5,6 +5,7 @@
 #ifndef MEASUREMENT_KIT_COMMON_ASYNC_HPP
 #define MEASUREMENT_KIT_COMMON_ASYNC_HPP
 
+#include <measurement_kit/common/poller.hpp>
 #include <measurement_kit/common/var.hpp>
 
 #include <functional>
@@ -18,7 +19,7 @@ struct AsyncState;
 class Async {
   public:
     /// Default constructor
-    Async();
+    Async(Poller *pol = Poller::global());
 
     /// Run the specified network test and call a callback when done
     /// \param func Callback called when test is done
--- a/src/common/async.cpp
+++ b/src/common/async.cpp
@@ -2,7 +2,6 @@
 // Measurement-kit is free software. See AUTHORS and LICENSE for more
 // information on the copying conditions.
 
-#include <event2/thread.h>
 #include <atomic>
 #include <measurement_kit/common/async.hpp>
 #include <measurement_kit/common/logger.hpp>
@@ -20,14 +19,6 @@ struct AsyncState {
     Poller *poller = Poller::global();
 };
 
-class EvThreadSingleton {
-  private:
-    EvThreadSingleton() { evthread_use_pthreads(); }
-
-  public:
-    static void ensure() { static EvThreadSingleton singleton; }
-};
-
 // Make sure that, no matter what kind of poller we're using and what kind of
 // events it periodically schedules, if any, there is at least one event
 static void keep_me_alive(Var<AsyncState> state) {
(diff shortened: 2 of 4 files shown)


bassosimone:
With valgrind we have no memory errors but all tests are leaked. Investigating, I've come to the conclusion that this has to do with the way in which lambda are used in `src/ooni`. In fact, the previous `async` implementation was specifically designed to avoid those lambda-related issues by avoiding to store `Var<NetTest>` in the lambda closure. I guess now it's come the time to solve the *real* issue.

bassosimone:
@AntonioLangiu, thanks for the great work you did in fixing leaks! I think most of this work shall be issued as part of other pull requests. Most fixes for OONI tests could go in the same commit, except the fix for `tcp_connect` which is more complex and should be explained on a commit of its own. I'd also put `a41d4e28` in a separate commit. It would also be nice to receive confirmation that there are no memory errors with valgrind in the context of such pull request in `./test/common/async`.

After that is merged, I'd rebase this pull request onto master and continue with the review process.

AntonioLangiu:
This branch should be rebased.

bassosimone:
Alright, I have rebased the branch. Reading again the diff I started to be pissed off by the fact that events scheduled before you enter the loop are lost, which is why call_later() is used in the regress tests. I am not sure which is the proper way of dealing with that, but I really dislike it.

bassosimone:
This pull request can now be closed because the approach pioneered here was implemented in #550 and #604. So, there's no need to keep this open, since all good ideas were cherry picked.

PR #391, opened 2016-02-22
Add dns usage example and fix warning

DavideAllavena:
<strike>This PR is an attempt to create a more abstract interface.  It is inspired by #362.

This pull request may not be complete. 
Do we need more examples? i.e. an example for each query()?
New function documentation is missing. Do we need to add it in this PR?</strike>

This PR adds an example usage of the new DNS API (as requested below) and, while there, it fixes a warning caused by not properly returning in case of error.

### DIFF
--- a/example/dns/query.cpp
+++ b/example/dns/query.cpp
@@ -48,27 +48,25 @@ int main(int argc, char **argv) {
     }
     std::string domain = argv[0];
 
-    Poller *poller = Poller::global();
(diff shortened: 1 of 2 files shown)


bassosimone:
@DavideAllavena the header/implementation part has now been superseded by the work done in #398. But the example part is very good. I'd be very happy to have this rebased onto #398 and accordingly updated such that we can merge in the example. :+1:

bassosimone:
Thanks, @DavideAllavena!

PR #395, opened 2016-02-24
Cleanup/http request

bassosimone:


bassosimone:
There is a segmentation fault in regress tests using g++. Currently investigating this (it is always nice that those kind of things pop up with one compiler and not with another :-)

AntonioLangiu:
I think it'is important to have a build matrix on Travis that checks such things for us.

bassosimone:
In general a build matrix is useful and I am totally in favor of it. But, in this case, it was enough to have g++ and clang++ to spot the issue. My comment was more of regret because I tested this on my Mac using clang++ and hence I did not spotted the segfault and opened the pull request.

bassosimone:
I've rebased and restarted the build. Interestingly it fails again (good that this thing is quite deterministic) and also interestingly cannot reproduce this bug using Vagrant and 14.04.

Will download 12.04 with Vagrant and try again next.

bassosimone:
#FastFact: issue is gone when switching to 14.04. Since the issue only appeared for g++ (and so, I imagine, GNU's libc++), I'm starting to wonder whether this had to do with a too young GNU C++ implementation in Ubuntu 12.04. It may have sense to apply b7b13ab775 to other recent failing pull requests (i.e. #398 and #396) and see whether this fixes them as well (maybe I'd run more than one test in a row to increase confidence that everything is good).

include/measurement_kit/http.hpp
```diff
@@ -131,6 +131,45 @@ Url parse_url(std::string url);
...
+/// \param body Optional HTTP request body.
+/// \param lp Optional logger.
+/// \param pol Optional poller.
+/// \param client Optional HTTP client.
+inline void GET(std::string url, RequestCallback cb,
+                Headers headers = {}, std::string body = "",
+                Settings settings = {}, Logger *lp = Logger::global(),
+                Poller *pol = Poller::global(),
+                Var<Client> client = Client::global()) {
+    settings["method"] = "GET";
```
hellais: I think that the `method` in the settings should not be overriden if it has been specified/passed.

Perhaps a check to ensure if it's set and if not set it by default to get is enough.

hellais: Ah no, nevermind I wasn't reading the class definition properly. This is ok. Ignore my previous comment.

hellais:
I have tested this and I like the new syntactic sugar. I am going to proceed in merging this.

PR #396, opened 2016-02-24
Feature/better connect

bassosimone:


hellais:
The failure here seems to be the same failure highlighted in PR #395

hellais:
I have been testing this on OS X 10.11 and I get the following errors:

```
We can override the default timeout
test/dns/resolver.cpp:174

test/dns/resolver.cpp:183: FAILED:
  REQUIRE( e == TimeoutError() )
with expansion:
  0 == 2007
```

I know that this one is actually not a problem with the unittest, but it's rather a problem with my current network. Since I am on vodafone they MITM DNS queries towards any IP address and will spoof the response so it's normal that it doesn't work.

The other one is:

```
Make sure that default get_output_path() is nonempty
test/ooni/dns_injection_test.cpp:53

test/ooni/dns_injection_test.cpp:53: FAILED:
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
```

This is a bit more scary and I will proceed to investigate into the root cause of it.

hellais:
Ok, so it seems like re-running the individual test doesn't reproduce the bug.

I tried cleaning up my repository directory to see if it had something to do with some temporary files being there, but nothing I still can't reproduce.

Given the fact that the changes introduced in these commits don't actually impact those sections of the code I am going to defer debugging this issue to the future and proceed with merging this.

Issue #392, opened 2016-02-23
dns: write a function to run dns test and takes needed parameters


bassosimone:
and move the object doing the resolution inside a private file

bassosimone:
This has now been merged in master

Issue #387, opened 2016-02-19
Nits in autogen.sh


bassosimone:
This is the "math co-processor" of #386:

- regular expressions could be more precise
- the AM_CONDITIONAL test is not necessary

bassosimone:
I've addressed the most important issues highlighted here in #410. Now closing.

PR #398, opened 2016-02-24
Feature/ng dns

hellais:
In this PR I address the changes we discussed with @bassosimone to the `mk::dns` API.

In particular it makes it no longer required to create the `Resolver` but DNS resolution is entirely handled via a `query` function.


hellais:
How is it possible that the unittest for the commit with just documentation changes fails, while the code one is :green_heart: :green_heart: :green_heart: ?

bassosimone:
Because among the tests there are integration tests that depend on network conditions and one of those has failed in the second case, I guess. (At a first glance, not totally clear to me what happens.)

bassosimone:
Closes #392 and #387.

bassosimone:
@hellais could you check whether applying b7b13ab7759 fixes the regress tests?

hellais:
Oh noes, now the clang build fails :(.

bassosimone:
Have tried this in Vagrant using 14.04 (which is what we're using on travis in this branch). It seems that the test that fails is [test/net/connection.cpp:61](https://github.com/measurement-kit/measurement-kit/blob/feature/ng-dns/test/net/connection.cpp#L61). Now trying to figure out why.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+                    answer.ipv6 = string;
+                    answer.type = QueryTypeId::AAAA;
+                }
+                logger->info("dns - adding '%s'", string);
+                answers.push_back(answer);
+            }
+
+        } else {
+            logger->warn("dns - too many addresses");
+            code = DNS_ERR_UNKNOWN;
```
AntonioLangiu: This line (and also the line 175) has no effect!

AntonioLangiu: This line (and also the line 175) has no effect!

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+    evdns_base *base;
+    event_base *evb = poller->get_event_base();
+
+    // XXX free variables when entering error path
+    if (settings.find("nameserver") != settings.end()) {
+        if ((base = evdns_base_new(evb, 0)) == nullptr) {
+            throw std::bad_alloc();
+        }
+        if (evdns_base_nameserver_ip_add(
+                base, settings["nameserver"].c_str()) != 0) {
```
bassosimone: Here you should `evdns_base_free(base, 1);` to free the allocated `evdns_base`.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+            throw std::runtime_error("Cannot set server address");
+        }
+    } else if ((base = evdns_base_new(evb, 1)) == nullptr) {
+        throw std::bad_alloc();
+    }
+
+    if (settings.find("attempts") != settings.end() &&
+        evdns_base_set_option(base, "attempts",
+                                    settings["attempts"].c_str()) != 0) {
```
bassosimone: Also here you should `evdns_base_free(base, 1);` to free the allocated `evdns_base`.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+        throw std::runtime_error("Cannot set 'attempts' option");
+    }
+    if (settings.find("timeout") != settings.end() &&
+        evdns_base_set_option(base, "timeout",
+                                    settings["timeout"].c_str()) != 0) {
```
bassosimone: Also here you should `evdns_base_free(base, 1);` to free the allocated `evdns_base`.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
-        dnsb = mk::get_global_evdns_base();
+    // By default we don't randomize the query's case
+    std::string randomiz{"0"};
+    if (settings.find("randomize_case") != settings.end()) {
+        randomiz = settings["randomize_case"];
     }
-    if (libs == nullptr) {
-        libs = Libs::global();
+    if (evdns_base_set_option(base, "randomize-case", randomiz.c_str()) !=
+        0) {
```
bassosimone: Also here you should `evdns_base_free(base, 1);` to free the allocated `evdns_base`.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+        // this is here only for robustness.
+        //
+        if (count >= 0 && count <= INT_MAX / size + 1) {
+
+            for (auto i = 0; i < count; ++i) {
+                // Note: address already in network byte order
+                if (inet_ntop(family, (char *)addresses + i * size,
+                              string, sizeof(string)) == NULL) {
+                    logger->warn("dns - unexpected inet_ntop failure");
+                    code = DNS_ERR_UNKNOWN;
```
bassosimone: @hellais setting `code` here as no effect. Since `inet_ntop()` can only fail in corner cases caused by programmer errors, here I'd just throw an exception to exit the program.

src/dns/query.cpp
bassosimone: Also in this case I'd throw an exception because again this is a corner case.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+            }
+
+        } else {
+            logger->warn("dns - too many addresses");
+            code = DNS_ERR_UNKNOWN;
+        }
+
+    } else {
+        logger->warn("dns - invalid response type");
+        code = DNS_ERR_UNKNOWN;
```
bassosimone: Also in this case I'd throw an exception because this cannot happen for real, unless we hit some serious bug.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+        case DNS_ERR_NODATA:
+            context->message.rtt = mk::time_now() - context->ticks;
+            break;
+        default:
+            context->message.rtt = 0.0;
+            break;
+    }
+
+    context->message.answers = build_answers_evdns(code, type, count, ttl,
+            addresses);
```
bassosimone: After my proposed changes above are implemented, this function throws. We may want to catch the exception here and route it, however, I'm not sure it's worth the effort since we always throw in edge cases where probably the most sane thing is to let the exception unwind all the stack and the program terminate.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+            Poller *poller) {
+
+    Message message;
+    Query query;
+
+    // XXX add proper exception handling for evbase creation
+    evdns_base *base = create_evdns_base(settings, poller);
+
+    if (dns_class != QueryClassId::IN) {
+        cb(UnsupportedClassError(), nullptr);
```
bassosimone: Here we should `evdns_base_free(base, 1)` to avoid leaking in the error path

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+void query(QueryClass dns_class, QueryType dns_type, std::string name,
+            std::function<void(Error, Message)> cb,
+            Settings settings,
+            Poller *poller) {
+
+    Message message;
+    Query query;
+
+    // XXX add proper exception handling for evbase creation
+    evdns_base *base = create_evdns_base(settings, poller);
```
bassosimone: Here we should catch exceptions (because they are mostly caused by user input) and callback. If we want to be precise, we should probably define different exceptions in `create_evdns_base()` and let the exception caused by a failing `evdns_base_new()` propagate because that's "out of memory" and it's not good to continue in that case.

src/dns/query.cpp
```diff
@@ -3,46 +3,306 @@
...
+    }
+
+    query.type = dns_type;
+    query.qclass = dns_class;
+    query.name = name;
+    message.queries.push_back(query);
+
+    //
+    // We explain above why we don't store the return value
+    // of the evdns_base_resolve_xxx() functions below
```
bassosimone: This comment references a comment that was lost in the refactoring. I'd reinstate the original comment, which was:

```
    //
    // Note: evdns_base_resolve_xxx() return a evdns_request
    // object that may be used to cancel the request. Yet, the
    // semantic of cancelling a request is such that evdns
    // could invoke the callback of a pending request if such
    // request was cancelled when its corresponding response
    // was already pending. This seems to be confirmed by
    // comments in libevent's evdns.c:
    //
    //     This does nothing if the request's callback is
    //     already running (pending_cb is set)
    //
    // For the above reason this class does not explicitly
    // cancel pending evdns requests and uses the `cancelled`
    // variable to keep track of cancelled requests.
    //
```

src/net/connection.cpp
```diff
@@ -169,23 +169,21 @@ void Connection::handle_resolve(Error error, char type,
...
+    for (auto answer: answers) {
+        std::string address;
+        if (answer.type == dns::QueryTypeId::A) {
+            address = answer.ipv4;
+            _family = "PF_INET";
+        } else if (answer.type == dns::QueryTypeId::AAAA) {
+            address = answer.ipv6;
+            _family = "PF_INET6";
+        } else {
+             abort();
```
bassosimone: Note to self: here it would be better to throw an exception (a kind of assertion that could be checked in the regress tests) rather than calling `abort()` (which instead is something that you cannot intercept).

src/net/connection.cpp
```diff
@@ -216,11 +214,11 @@ bool Connection::resolve_internal(char type) {
...
 
-    dns_request = dns::Query(
+    dns::query(
         dns::QueryClassId::IN, query,
-        address, [this](Error error, dns::Response resp) {
-            handle_resolve(error, resp.get_type(), resp.get_results());
-        }, logger, poller->get_evdns_base());
+        address, [this](Error error, dns::Message message) {
+            handle_resolve(error, message.answers);
+        }, {}, poller);
```
bassosimone: The previous semantic was that the DNS request was to be alive as long as the connection was alive. The new semantic is that the lifecycle of the connection is independent of the lifecycle of the query. This is bad because it may lead to use-after-free in certain cases. As discussed with @hellais when preparing this pull request, the idea was to change `connect()` API to avoid giving the user ownership of the connection before connect is complete shortly after this pull request was merged, so to avoid the issue that I'm discussing here.

bassosimone:
In addition to all the changes suggested above, it would be nice to have the branch rebased onto master and also to run `clang-format -i` on the files.

src/ooni/dns_injection_impl.hpp
```diff
@@ -37,9 +37,9 @@ class DNSInjectionImpl : public DNSTestImpl {
               std::function<void(report::Entry)> &&cb) {
         entry["injected"] = NULL;
         query("A", "IN", input, options["nameserver"],
-              [this, cb](dns::Response response) {
```
AntonioLangiu: @hellais you lost this commit while merging and this is why it doesn't compile

src/dns/query.cpp
bassosimone: Note to self: this is now fixed in the branch (do not understand why github does not tell me that and I still see the comment, but whatever)

src/dns/query.cpp
bassosimone: Fixed in branch.

src/dns/query.cpp
bassosimone: Fixed in branch.

src/dns/query.cpp
bassosimone: Fixed in branch.

src/dns/query.cpp
bassosimone: Also this comment is addressed. We've decided that here we will throw.

src/dns/query.cpp
bassosimone: This was fixed in branch.

src/net/connection.cpp
bassosimone: Okay: the corresponding failing test has been disabled.

src/net/connection.cpp
bassosimone: Since we're going to refactor this code anyway, I'd say we can defer this change.

src/ooni/dns_injection_impl.hpp
bassosimone: This has now been addressed by a hotfix I pushed to repair building another branch.

bassosimone:
Review complete! It is now okay to merge master into this branch (as I've just done) and then proceed with automatic merging (which I'll do as soon as travis-ci completes).

bassosimone:
Merging in 3... 2... 1...

PR #402, opened 2016-03-09
Document call_later()'s design pattern

bassosimone:
Hi @hellais, could you review this pull request in which I try to explain measurement-kit's pattern of using functions as opposed to objects in certain cases?


hellais:
I think a couple of reads of the text are necessary to fully grasp the rational behind these design decisions, but I think it's good.

I don't think there are improvements to be made to make it clearer. This is good to merge for me. :+1:

PR #405, opened 2016-03-14
Update poller implementation to prepare for #390

bassosimone:
This contains poller-related changes that are required by #390.

1. implement call_later()

2. make sure event-loop does not exit

3. deprecate loop_once()

The deprecation of loop_once() is because after #390 that functionality becomes unused.

### DIFF
--- a/src/common/poller.cpp
+++ b/src/common/poller.cpp
@@ -20,6 +20,11 @@ static void mk_call_soon_cb(evutil_socket_t, short, void *p) {
     delete cbp;
 }
 
+static void do_periodic(evutil_socket_t, short, void *ptr) {
+    mk::Poller *poller = static_cast<mk::Poller *>(ptr);
+    poller->handle_periodic_();
+}
+
 } // extern "C"
 
 namespace mk {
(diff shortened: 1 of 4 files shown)


src/common/poller.cpp
```diff
@@ -36,16 +38,58 @@ Poller::~Poller() {
...
         delete cbp;
         throw std::runtime_error("event_base_once() failed");
     }
 }
 
+void Poller::call_soon(std::function<void()> cb) {
+    call_later(-1.0, cb);
+}
+
+static void do_periodic(evutil_socket_t, short, void *ptr) {
```
AntonioLangiu: This should be declared as extern "C"

test/common/poller.cpp
```diff
@@ -153,3 +153,22 @@ TEST_CASE("poller.call_soon() works") {
     poller.call_soon([&poller]() { poller.break_loop(); });
     poller.loop();
 }
+
+TEST_CASE("poller.call_later() works") {
+    mk::Poller poller;
+    poller.call_later(3.14, [&poller]() { poller.break_loop(); });
+    poller.loop();
```
AntonioLangiu: Require that break_loop is called after at least 3 seconds

AntonioLangiu:
I have checked it with valgrind and it is ok :+1:
Ready to merge!

PR #406, opened 2016-03-15
http: add rewritten http parser

bassosimone:
This pull request, in particular, adds a new http parser. I tried to gently evolve the existing parser in the desired direction, but after I while I realized a rewrite was better.

Of course, this is *not* a rewrite from scratch, but rather refactoring.

### DIFF
--- a/src/http/response_parser.hpp
+++ b/src/http/response_parser.hpp
@@ -50,10 +50,16 @@ class ResponseParser : public NonCopyable, public NonMovable {
     ResponseParserImpl *impl = nullptr;
 };
 
-// Header parsing states (see do_header_internal)
+// Header parsing states (see do_header_internal) - used by old impl.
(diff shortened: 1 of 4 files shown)


src/http/response_parser.hpp
```diff
@@ -48,6 +50,163 @@ class ResponseParser : public NonCopyable, public NonMovable {
     ResponseParserImpl *impl = nullptr;
 };
 
+// Header parsing states (see do_header_internal)
+#define S_NOTHING 0
+#define S_FIELD 1
```
AntonioLangiu: I would prefer using enums instead of macros, so you have some compilation check and you can't give to the finite state machine a number as a state!

bassosimone: I agree! Please see 6879183c8!

PR #408, opened 2016-03-17
vagrant: add support for several ubuntus

bassosimone:


bassosimone:
Well, it turns out that effectively supporting precise64 is not simple. Furthermore, vivid64 is not exported anymore in atlas/hashicorp, so I am going to close this pull request as "wontmerge"

PR #409, opened 2016-03-18
travis: fix error checking

bassosimone:
Spotted by me and fixed by @AntonioLangiu. Move the if down otherwise the if steals the success/failure bit set by the build and the build does not actually fail.

In the future we should use external script for this. Relying on travis.yml is way too fragile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,9 @@ install:
           ARG="$ARG --disable-shared";
       fi  
 
+# Only measure coverage with clang for now because I cannot manage to
+# convince coveralls to find all sources when using gcc-4.8 and gcov-4.8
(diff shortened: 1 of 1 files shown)


bassosimone:
For now we have fixed the build (where fixed means that builds that should fail actually fail). What we have not fixed up yet is the coverage. This is an error we encountered before. The coveralls script [dies because it cannot stat common/var.hpp](https://travis-ci.org/measurement-kit/measurement-kit/jobs/116796451#L5678). This is really frustrating.

bassosimone:
Good. I've figured out how to restore coverage measurements, which was my priority for this pull request together with letting the code fail when it needs to. Now moving on.

PR #410, opened 2016-03-18
Fix/autogen.sh

bassosimone:


bassosimone:
Closes #387 (the level of nits specified in that issue is excessive and I'd say it's ok to just close that issue with the fix contained here).

PR #413, opened 2016-03-24
Add single dependency build script 

AntonioLangiu:
In this PR I'm integrating the work of @bassosimone in [this branch](https://github.com/measurement-kit/measurement-kit/tree/feature/build-improvements).

Pratically, now we are able to build a single dependency, with a script that downloads it from a git repository and builds it inside a defined folder.

To see more details check the script [./build/dependency](https://github.com/measurement-kit/measurement-kit/blob/feature/build/build/dependency)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ Makefile.in
 __venv__
 aclocal.m4
 autom4te.cache/
+builtin/
 compile
 config.guess
 config.h
@@ -107,3 +108,4 @@ test/ooni/tcp_connect_test
 test/ooni/tcp_test
 test/report/file
 test/traceroute/android
+third_party/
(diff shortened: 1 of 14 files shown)


AntonioLangiu:
I think this is ok to merge!

bassosimone:
Okay, thanks! I have rebased and fixed conflicts. While at it, I've prepended `LC_ALL=C` to the sort command to make sure sorting is always done using the same criterion.

Going to merge as soon as travis-ci completes.

bassosimone:
To provide additional context, let me add that this build script is necessary to decouple MK from the dependencies that typically are available on common systems (i.e. libevent).

This is a preparatory change to recompile libressl as a dependency for iOS and Android, which is something that we need to have in tree to fully implement SSL.

PR #428, opened 2016-04-01
Allow to use socks5 with new net::connect()

bassosimone:


bassosimone:
I think at this point this branch is too old to extract something valuable from it.

PR #435, opened 2016-04-01
net/filter.hpp: filter for bufferevent

bassosimone:
This would be extremely useful to implement socks5 as a filter so that next it would be possible to have a ssl bufferevent over a socks5 bufferevent over a plain normal socket.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ example/dns/query
 example/http/request
 example/mlabns/query
 example/net/connect
+example/net/filter
 example/net/listen
 example/ooni/dns_injection
 example/ooni/http_invalid_request_line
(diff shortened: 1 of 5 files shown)


bassosimone:
Before merging this pull request, it would be nice to understand how to make sure we always echo back all we received also when receiving EOF. (This seems to be, by the way, not only a problem of the filter but also a problem of `example/net/listen.cpp` echo implementation.)

- [ ] understand how to have echo working properly

bassosimone:
This was a nice experiment, but I'd rather decouple our implementation of `socks5` from libevent than going for more coupling between us and libevent. Not that I don't like libevent, but it seems to me there are other ways for improving our `socks5` that are equally easy to implement and more library independent.

Issue #400, opened 2016-02-24
Improving our NDT client


bassosimone:
- [x] understand the feasibility of wrapping NDT using a C++11 interface and then proceeding in small steps to evolve each phase of the test to be asynchronous
- [x] merge connect-many (#467)
- [x]  merge initial implementation of NDT client (#441)
- [x] teach the client to wait in queue
- [x] merge full fledged NDT client
- [x] save results into a json-ish structure
- [ ] write regress test verifying that we can read messages we serialize
- [ ] close connection if test version is too old
- [ ] introduce specific error for each failed test rather than "a test failed" error
- [ ] protect against functions calling themselves "forever" using a counter
- [ ] send more metadata to server (what specifically?)
- [ ] terminate download coroutine if the server seems to have gone wild and sends data "forever"
- [ ] implement the WebSockets transport
- [ ] make sure we can run WebSockets over TLS

bassosimone:
We're not going to continue improving the ndt client in MK. The future is ndt7 and we'll expose it to OONI apps using ooni/probe-engine.

PR #441, opened 2016-04-04
NDT client: first iteration

bassosimone:
This branch has now reached the "ready for review state". It contains the initial implementation of the NDT client, already good enough to start testing with it and already tested enough (coverage super high for all files of this pull request) to stay in master. Issue #400 highlights future improvements that could be implemented as separate pull requests once this branch has been merged.


src/ndt/run_impl.hpp
```diff
@@ -0,0 +1,122 @@
...
+    dump_settings(ctx->settings, "ndt", ctx->logger);
+
+    // The following code implements this sequence diagram:
+    // https://raw.githubusercontent.com/wiki/ndt-project/ndt/NDTProtocol.images/ndt_10.png
+
+    connect(ctx, [ctx](Error err) {
+        if (err) {
+            disconnect_and_callback(ctx, err);
+            return;
+        }
```
hellais: Maybe it would be cleaner to make this into a macro.

src/ndt/protocol_impl.hpp
```diff
@@ -0,0 +1,260 @@
...
+        }
+        ctx->logger->info("Authorized tests: %s", s.c_str());
+        ctx->granted_suite = split(s);
+        callback(NoError());
+    });
+}
+
+template <MK_MOCK_NAMESPACE(test_c2s, run), MK_MOCK_NAMESPACE(test_meta, run),
+          MK_MOCK_NAMESPACE(test_s2c, run)>
+void run_tests_impl(Var<Context> ctx, Callback<Error> callback) {
```
hellais: I am thinking that it may be useful for a user of the API to be able to run specific tests. My understanding of how this call works is that you look at the tests that are supported by the NDT server and run through them in the order by which they are returned from the server.

Wouldn't it be better instead to have the order, and the types of tests to run, be defined by a user of the `NdtTest` DSL?

Perhaps having a call to set the types of test to run.

src/ndt/protocol_impl.hpp
```diff
@@ -0,0 +1,260 @@
...
+    case TEST_C2S:
+        func = test_c2s_run;
+        break;
+    case TEST_META:
+        func = test_meta_run;
+        break;
+    case TEST_S2C:
+        func = test_s2c_run;
+        break;
+    }
```
hellais: Maybe it's good to add checks for all the supported NDT tests `TEST_MID`, `TEST_SFW ` (even if they are not yet implemented in NDT) and as suggested by the NDT design doc:

> The Client MUST drop the connection when it receives unknown test id, because such situation means an error on the Server's part

src/ndt/protocol_impl.hpp
hellais: Edit: 

I realise now that apparently

> The tests MUST be performed in the order received from the Server

I guess then the above suggestion would be to instead of setting `test_suite = TEST_STATUS | TEST_META | TEST_C2S | TEST_S2C` make it possible to select also a subset of all tests.

I am thinking this could be particularly useful in the case where one wants to run a download test, but not an upload test or vice-versa.

src/ndt/test_c2s_impl.hpp
```diff
@@ -0,0 +1,193 @@
...
+                        Var<double> previous(new double(begin));
+                        Var<size_t> count(new size_t(0));
+                        logger->debug("ndt: resume coroutine");
+                        logger->info("Starting upload");
+                        txp->set_timeout(timeout);
+                        txp->on_flush([=]() {
+                            double now = time_now();
+                            if (now - *previous > 0.5) {
+                                double x =
+                                    (*count * 8) / 1000 / (now - *previous);
```
hellais: Count is the result of calling size() on string, which is the size of the string in bytes so there is no need to multiply this by 8.

hellais: On second thought, I guess it depends what is the unit of measure you have below.

If below you mean to use kilobits, then it's OK like this, but I would probably change the reading of the info log call to make the clearer, such as (`Speed: %.2f kbit/s`).

src/ndt/run_impl.hpp
bassosimone: Yeah! Implemented in cb884a744a.

src/ndt/test_c2s_impl.hpp
bassosimone: I agree! Now using `kbit/s`, see e65cc9bf4544!

src/ndt/protocol_impl.hpp
bassosimone: Refactored the code in 77ef0835fe8b to simplify the flow and clarify that, on invalid test id, the connection with the server is closed. (Better to see [the diff without considering space changes](https://github.com/measurement-kit/measurement-kit/commit/77ef0835fe8b73def24b1061759a1767672cd26c?w=1), for clarity.)

bassosimone: Then also added 445a3778e5 because 77ef083 was broken.

src/ndt/protocol_impl.hpp
bassosimone: I agree. I have implemented the possibility to specify which test to run in d9bdd241.

bassosimone:
@hellais Thanks for all the comments! I think I have implemented all the recommended suggestions!

hellais:
Ok I have finished another pass of reviewing and testing of this branch and it looks good to me.

I think we can merge this into master!

🎉 👍 💯

hellais:
I finished another pass of review and I think this is good to be merged into master 👍 💯 

As discussed with @bassosimone adding support for obtaining the result of a NDT test will be added in a PR on top of this one.

bassosimone:
The build failed because a header needs to be changed to reflect #627. [Apparently github is currently stuck](https://twitter.com/githubstatus/status/734789351754846209); I've pushed the fix and a couple of reverts to trigger it, but the web interface (and their backend) has not updated yet.

bassosimone:
Thanks for the review, @hellais! Now merging!

PR #442, opened 2016-04-04
response_parser: remove auto-delete impl feature

bassosimone:
I reckon this was added to avoid use after frees, but now that it is
clear that one should defer destruction after end of test, I think that
this auto-delete feature is not needed anymore.  In fact, delayed
deletion is now performed by Async.

We'll see when testing if there are any memory issues (especially given
that now I've committed a diff to improve Valgrind diagnostics).

If there are no memory issues, then good because the duplicate
interface/impl for the parse is generating lots of complexity. Would be
nice in another diff to merge interface and implementation to obtain
just a single class implementing the respoinse parser.

I've added a comment reminding us to do so.


bassosimone:
Okay, so apparently there are [novel Valgrind related issues added by this PR](https://travis-ci.org/measurement-kit/measurement-kit/jobs/120668444#L5726). Damn.

bassosimone:
Now that #451 has been merged, I can get rid of this failed attempt at simplifying existing http code to be able to use it to write new http code. We have found another way.

PR #446, opened 2016-04-05
Use JSON instead of YAML for the data format of results

hellais:
* Write utility functions for computing time in UTC


src/ooni/ooni_test_impl.hpp
```diff
@@ -63,10 +80,15 @@ class OoniTestImpl : public mk::NetTest {
         probe_cc = "ZZ";
     }
 
-    void write_header() {
+    void set_test_start_time() {
+        if (test_start_time == nullptr) {
+        }
+    }
```
bassosimone: This function seems to be unused. Can I get rid of it before merging?

src/common/utils.cpp
```diff
@@ -40,6 +40,21 @@ double time_now(void) {
     return result;
 }
 
+void utc_time_now(struct tm **utc) {
+    time_t tv;
+    time(&tv);
+    *utc = gmtime(&tv);
```
bassosimone: Could you refactor this as

```C
*utc = gmtime(time(nullptr));
```

Rationale for this request is the following excerpt of text from the time(3) on Linux:

```man
BUGS
       Error returns from this system call are indistinguishable from
       successful reports that the time is a few seconds before the Epoch, so
       the C library wrapper function never sets errno as a result of this
       call.

       The tloc argument is obsolescent and should always be NULL in new code.
       When tloc is NULL, the call cannot fail.
```

src/common/utils.cpp
```diff
@@ -40,6 +40,21 @@ double time_now(void) {
...
+}
+
+std::string timestamp(struct tm *t) {
+    char result[30];
+    std::string ts;
+    if (strftime(result, 30, "%Y-%m-%d %H:%M:%S", t) == 0) {
```
bassosimone: In this line, could you `s/30/sizeof (result)/`?

src/ooni/dns_test_impl.hpp
```diff
@@ -27,45 +29,42 @@ class DNSTestImpl : public ooni::OoniTestImpl {
...
 
         dns::query(
             query_class, query_type, query_name,
             [=](Error error, dns::Message message) {
                 logger.debug("dns_test: got response!");
-                YAML::Node query_entry;
+                json query_entry;
+                query_entry["resolver_hostname"] = resolver_hostname;
+                query_entry["resolver_port"] = resolver_port;
+                query_entry["failure"] = NULL;
```
bassosimone: I recall that yaml-cpp converted `NULL` to zero rather than to `null`. I am not sure of what is done by this new json library. Could you check whether we end up with the desired value in the output? Otherwise, it's probably reasonable to expect that, to get `null` on output, one shall use `nullptr` in C++.

bassosimone:
I really really like this diff! I've just highlighted a couple of small fixes that I'd like to see applied before merging! As for the travis failure, nevermind, I merged the master branch and this triggered travis, then I stopped travis to give priority to another build.

src/ooni/ooni_test_impl.hpp
hellais: Yeah sorry, I forgot to trash this code after having refactored it.

src/ooni/dns_test_impl.hpp
hellais: Will look into this.

src/common/utils.cpp
hellais: Yeah.

src/common/utils.cpp
hellais: Seems reasonable.

hellais: It actually get's even worse with gmtime.

Reading the docs for gmtime(3) I see:

```
BUGS
     Except for difftime(), mktime(), and the _r() variants of the other functions, these functions leaves their result
     in an internal static object and return a pointer to that object.  Subsequent calls to these function will modify
     the same object.

     The C Standard provides no mechanism for a program to modify its current local timezone setting, and the
     POSIX-standard method is not reentrant.  (However, thread-safe implementations are provided in the POSIX threaded
     environment.)

     The tm_zone field of a returned tm structure points to a static array of characters, which will also be overwritten
     by any subsequent calls (as well as by subsequent calls to tzset(3) and tzsetwall(3)).

     Use of the external variable tzname is discouraged; the tm_zone entry in the tm structure is preferred.
```

Basically calling gmtime() multiple time leads to the same shared object being modified. :dizzy_face:

bassosimone: We should use `gmtime_r()` then!

bassosimone:
Thank you! Merging!

PR #447, opened 2016-04-06
Fix cross compilation for iOS

AntonioLangiu:
This PR fix the compilation for iOS.


bassosimone:
Looks good to me! I'm a bit swamped at the moment (see below). In any case it's ok to me if you squash-and-merge this yourself once travis completes. Thank you!

Wrestling in the mud

PR #450, opened 2016-04-07
Remove tests and example from the coverage check

AntonioLangiu:
I think before we were distorting the real coverage of measurement_kit.
I think that examples and tests should not be considered in the coverage calculatioin. Examples was never executed and was decreasing the coverage and tests was always executed and was increasing the coverage. I think the only files we should cover are those included in `src`, excluding `src/ext`.


bassosimone:
I agree!

AntonioLangiu:
This is always logic for me but it seems it is not logic for coveralls. When removing the test folder it doesn't calculate good the coverage. I'm going to re-add it!

bassosimone:
Okay, just `git revert` fd41978b176 then!

PR #451, opened 2016-04-07
http: sketch out new implementation

bassosimone:
This pull request is based on #406. It adds to such pull request a simple but effective implementation of http. <strike>What is not working yet is socks5, which requires some extra work in `net`</strike>.

When this is merged and also socks5 is fixed, we can finally start cleaning up the tree and remove a number of functionalities, including old http code, old connect code, etc.


AntonioLangiu:
I'm working on it!

src/http/request.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+namespace mk {
+namespace http {
+
+using namespace mk::net;
+
+void request_connect(Settings settings, RequestConnectCb cb,
+        Poller *poller, Logger *logger) {
+    if (settings.find("url") == settings.end()) {
+        cb(GenericError(), nullptr);
```
AntonioLangiu: Use a more explicit error!

src/http/request.cpp
```diff
@@ -0,0 +1,117 @@
...
+    connect(url->address, url->port, cb, settings, logger, poller);
+}
+
+void request_send(Var<Transport> transport, Settings settings, Headers headers,
+        std::string body, RequestSendCb callback) {
+    RequestSerializer serializer;
+    try {
+        serializer = RequestSerializer(settings, headers, body);
+    } catch (std::exception &) {
+        callback(GenericError());
```
AntonioLangiu: Here too, use more specific error!

src/http/request.cpp
```diff
@@ -0,0 +1,117 @@
...
+    transport->write(buff);
+}
+
+void request_recv_response(Var<Transport> transport, RequestRecvResponseCb cb,
+        Poller *poller, Logger *logger) {
+    Var<ResponseParserNg> parser(new ResponseParserNg);
+    Var<Response> response(new Response);
+    Var<bool> prevent_emit(new bool(false));
+
+    transport->on_data([=](Buffer data) { parser->feed(data); });
```
AntonioLangiu: I think here you can give to the lambda just the variables it needs!

test/http/request.cpp
```diff
@@ -292,3 +292,134 @@ TEST_CASE("The callback is called if input URL parsing fails") {
...
+        });
+    });
+}
+
+static inline bool status_code_ok(int code) {
+    return code == 302 or code == 200;
+}
+
+TEST_CASE("http::request_recv_response() works as expected") {
+    loop_with_initial_event([]() {
```
AntonioLangiu: should you check the connectivity before the test?

test/http/request.cpp
```diff
@@ -292,3 +292,134 @@ TEST_CASE("The callback is called if input URL parsing fails") {
...
+    });
+}
+
+TEST_CASE("http::request_cycle() works as expected using Tor") {
+    loop_with_initial_event([]() {
+        request_cycle({
+            {"method", "GET"},
+            {"url", "httpo://www.google.com/robots.txt"},
+        }, {}, "", [](Error error, Var<Response> r) {
+            // Test disable because we don't know if Tor is running
```
AntonioLangiu: I think we should find a way to exclude this test if we don't know that tor is runnig. Maybe we can put all the tests that need tor in a test file and exclude if it is not necessary, or maybe we can define a preprocessor variable in the configure that enables this code, something like 
#ifdef TOR_TEST...

AntonioLangiu:
I really like the implementation of http/request! I'm building it locally to see the coverage, because coveralls is hiding some files!

test/http/request.cpp
bassosimone: I wrote a check that is able to detect the two cases where Tor is running and where Tor is not running and adapt what to `REQUIRE` accordingly. This should address your comment, me thinks.

src/http/request.cpp
bassosimone: Done!

src/http/request.cpp
bassosimone: This should also be fixed!

src/http/request.cpp
bassosimone: > I think here you can give to the lambda just the variables it needs!

I prefer not doing that. In my opinion, having verified that it's safe to share all the variables through `[=]`, the safest thing is to always use `[=]` in the closure, so that everything is always shared.

This style, in my opinion, makes it simpler to reason about the code than when you selectively indicate what yo share, because you have a closure of the relevant variables.

I also understand that creating closures like this causes copies that may have impact on performance, but honestly I'd rather measure that we have a problem before starting to optimize.

As for the possibility of sharing pointers that one should not share using `[=]`, this is true but in our case it does not apply. We could use `Var<Poller>` and `Var<Logger> in the future to reduce the theoretical risk of this issue and be more clean. But, again, before going down that road I'd rather do some more experience points trying to use closures.

> should you set the callback to null also here?

The transport's callbacks are cleared as part of the handling of the error event. The error event should be emitted on network error, or when the parser reaches end (with a boolean value protecting double invocation of the error event). We should reason a bit more on whether this would be enough to break possible reference cycles (which is something to be very careful about until we properly hold a reference to transport instances in use in transport code). I will add a comment to remind us to check whether we need to do more.

test/http/request.cpp
bassosimone: Yes, definitely! I've deployed check-connectivity checks to all the file!

PR #452, opened 2016-04-08
Big cleanup: remove http/client

bassosimone:
First pull request in a wave of pull requests for cleaning up MK after recent changes in DNS, net, and HTTP code that improved API safety and ease of usage.

Here we take down the HTTP client, refactor existing tests, and mark with XXX and/or TODO places in the code that need further love.

Before we proceed, it seems there are [memory errors introduced by this cleanup](https://travis-ci.org/measurement-kit/measurement-kit/jobs/122739892#L3075).

- [x] fix memory errors highlighted above
- [x] add todo list regarding how to proceed from now at bottom

### DIFF
--- a/include/measurement_kit/http/http.hpp
+++ b/include/measurement_kit/http/http.hpp
@@ -68,32 +68,6 @@ typedef std::function<void(Error, Response)> RequestCallback;
 // Forward declaration of internally used class.
 class Request;
 
-/// HTTP client.
-class Client {
-
-  protected:
-    std::set<Request *> pending;
-
-  public:
-    /// Issue HTTP request.
-    void request(Settings settings, Headers headers, std::string body,
-                 RequestCallback callback, Logger *lp = Logger::global(),
-                 Poller *poller = Poller::global());
-
-    /// Get the global HTTP client instance
-    static Var<Client> global() {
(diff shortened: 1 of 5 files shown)


bassosimone:
So, memory errors are fixed by this diff:

```patch
diff --git a/src/http/request.cpp b/src/http/request.cpp
index 542cf63..8c40b98 100644
--- a/src/http/request.cpp
+++ b/src/http/request.cpp
@@ -84,12 +84,12 @@ void request_recv_response(Var<Transport> transport, RequestRecvResponseCb cb,
             *prevent_emit = true;
             parser->eof();
         }
-        cb(err, response);
+        // TODO: make sure we remove all cycles
         transport->on_error(nullptr);
         transport->on_data(nullptr);
         poller->call_soon([=]() {
             logger->debug("request_recv_response: end of closure");
-            // TODO: make sure we remove all cycles
+            cb(err, response);
         });
     });
 }
```

Let's now try to justify why this happens.

bassosimone:
# TODO list for @AntonioLangiu 
- [ ] create issue of migrating `Poller *` and `Logger *` to `Var<Poller>` and `Var<Logger>`.
- [ ] review and merge this pull request

Issue #448, opened 2016-04-06
OSX on travis failure


AntonioLangiu:
The build on the OSX machine on travis fails due to the test [`test/net/connect`](https://travis-ci.org/measurement-kit/measurement-kit/jobs/120867977#L4721).

The problem is that it expects a `NetworkError` but it receives a `TimeoutError`. I was able yesterday to reproduce this problem locally, but it happens just in some Network conditions. (For example now the test pass)

@bassosimone and @DavideAllavena do you think that we should check for both? Is it normal that when you try to make a connection to a closed port of a server you have a timeout problem?

bassosimone:
Yes, this seems reasonable. I did some testing as well and managed to reproduce the issue here for me in a couple of cases, under OSX. Please, go ahead!

AntonioLangiu:
So, I was able to reproduce the error setting a rule to the firewall to block all the icmp packets. 
So the problem seems that the travis osx machine blocks icmp (or maybe just the type 3: Destination Unreachable, that is what we need in this case), so the test receive a `TimeoutError` instead of a `NetworkError`. 
If you want to reproduce the test on your machine you can try with ubuntu and iptables, or like I did with mac and pfctl. Try to add this rule to your `/etc/pf.conf`:

    block in log proto icmp all

and then reload the pfctl configuration with:

    pfctl -Fa -f /etc/pf.conf

You should get the same error reported on travis!

AntonioLangiu:
#456 fixed the problem, as you can see in [this travis build](https://travis-ci.org/measurement-kit/measurement-kit/builds/121883739)

bassosimone:
Thank you for investigating this issue! I think the issue itself could now be closed and the pull request deal with once the requested changes have been applied!

PR #457, opened 2016-04-09
Refactor the build matrix

AntonioLangiu:
The follow improvements are added:
    - more clear code
    - "install" inside ubuntu trusty using beta apt (avoiding sudo)
    - export just one variable in the environment and using it for CC and CXX
    - export CC and CXX for every architecture except for iOS build
    - fix valgrind using suppression file
    - correct the order of the builds
    - check witch combination takes less time

Now 4 builds are allowed to fail. One should fail, the other 3 not, but I did so
because with fast-finish we are able to check the build and merge the PR faster.

I think is useful to have valgrind working, so we can check if we are introducing a memory leak without checking the build, and if we are introducing a memory leak and we don't want to fix it for the moment, we can just add it to the suppression file. Every time valgrind detect an error it prints the suppression code, that must be just copied in `build/travis/suppression_file.supp`.


.travis.yml
```diff
@@ -1,24 +1,19 @@
...
 
 install:
-    - if [ "$EXPORT" = "true" ]; then
-        if [ "$CXX" = "g++" ]; then export CXX="g++-$CXX_V" CC="gcc-$CC_V"; fi;
-        if [ "$CXX" = "clang++" ]; then
-          export CXX="clang++-$CXX_V" CC="clang-$CC_V";
-        fi;
-      fi
-    - ./autogen.sh
+    - env
```
AntonioLangiu: This should be removed

Makefile.am
```diff
@@ -29,7 +29,9 @@ TESTS = $(ALL_TESTS)
 check_PROGRAMS = $(ALL_TESTS)
 
 VALGRIND = valgrind --trace-children=yes --error-exitcode=1                    \
-                    --dsymutil=yes --leak-check=yes
+                    --dsymutil=yes --leak-check=yes 						   \
+					--gen-suppressions=all									   \
```
AntonioLangiu: Reindent

test/net/connect.cpp
```diff
@@ -110,14 +110,20 @@ TEST_CASE("connect_base works with ipv4") {
...
+
 TEST_CASE("connect_base works with ipv4 and closed port") {
     if (CheckConnectivity::is_down()) {
         return;
     }
     loop_with_initial_event([]() {
         connect_base("130.192.16.172", 81,
                 [](Error err, bufferevent *bev) {
-                    REQUIRE(err == NetworkError());
+                    REQUIRE(checkError (err));
```
AntonioLangiu: This should be fixed with #448

AntonioLangiu:
The only problem of the fact that the valgrind build is not allowed to fail is that sometimes we have a fail with the DelayedCall unit test. @bassosimone what can be a solution for this?

bassosimone:
I am against the suppression file part. I do not want to suppress memory errors in our code. I think this is very bad practice and that we should not do that. I would prefer, instead, a patch where Valgrind is not allowed to fail and our entire build fails until we fix the issues. This, I think, would be the more integralist way to address the problem that certain parts of our code have memory errors.

.travis.yml
```diff
@@ -1,24 +1,18 @@
 language: cpp
 
 before_install:
-  - if [ "$INSTALL" = "true" ]; then
-      sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test;
-      sudo apt-get update -qq;
-      sudo apt-get install -qq -y --force-yes gcc-4.8 g++-4.8;
-    fi
```
bassosimone: This has been removed because made redundant by specifying the packages, correct?

.travis.yml
```diff
@@ -1,24 +1,18 @@
...
-          export CXX="clang++-$CXX_V" CC="clang-$CC_V";
-        fi;
-      fi
-    - ./autogen.sh
+    - if [ "$CXX" = "g++" -a -n "${VER+1}" ]; then 
+          export CXX="g++-$VER" CC="gcc-$VER"; 
+      fi;
+    - if [ "$CXX" = "clang++" -a -n "${VER+1}" ]; then
+          export CXX="clang++-$VER" CC="clang-$VER";
+      fi;
```
bassosimone: I really don't like the way in which this is implemented! Why don't we directly export `CXX` and `CC`? If it's not possible to specify that variables directly, why don't use `MK_CXX` and `MK_CC` to override `CXX` and `CC` in the proper place of the Valgrind file?

.travis.yml
```diff
@@ -58,10 +55,9 @@ after_failure:
 matrix:
   fast_finish: true
   include:
-
     - sudo: false
       compiler: g++-5
-      env: OS="Ubuntu 12.04" CXX_V=5 CC_V=5 EXPORT=true
+      env: OS="Ubuntu 12.04" VER=5
```
bassosimone: Given what I said above, this should read `CXX=g++-5 CC=gcc-5`

.travis.yml
```diff
@@ -71,32 +67,39 @@ matrix:
...
       addons:
         apt:
           packages:
-            - g++-4.8
-            - gcc-4.8
+            - clang-3.6
+            - llvm-3.6-dev
+            - valgrind
           sources:
+            - llvm-toolchain-precise-3.6
```
bassosimone: Here basically you have removed compiling with g++-4.8, correct?

bassosimone:
From the description, I also understand that you changed the order of the tests. I would have preferred to see such cosmetics as part of another (subsequent) pull request, for readability++.

bassosimone:
Awesome work! Now the build time is greatly reduced, we stay in Ubuntu 12.04 where we observed memory issues in the past (so we're not avoiding problems), and we're still measuring everything that matterns. We will add more builds as we approach releases and then remove (comment) them again after release. Thank you!

PR #461, opened 2016-04-10
test/net/buffer.cpp: fix memleak

bassosimone:
This was of the memory errors spotted in #452.


AntonioLangiu:
I think it is ok :+1:

PR #467, opened 2016-04-14
Implement connect_many() required by NDT

bassosimone:
There is a case where the NDT client could open more than one connection to the remote server. Add code in `net` to allow to implement that use case. Also add regress tests.


src/net/connect-many.hpp
```diff
@@ -0,0 +1,67 @@
...
+                return;
+            }
+            ctx->connections.push_back(txp);
+            --ctx->left;
+            connect_many_<do_connect>(ctx);
+        },
+        ctx->settings, ctx->logger, ctx->poller);
+}
+
+static Var<ConnectManyCtx> connect_many_make(std::string address, int port,
```
AntonioLangiu: I think the right position of this function would be as a static method of the class ConnectManyCtx.

AntonioLangiu:
I think is ready to merge, I'm waiting for the travis build!

bassosimone:
It seems [there is a leak](https://travis-ci.org/measurement-kit/measurement-kit/jobs/123769352#L2745). I will look into it in the next days.

src/net/connect-many.hpp
bassosimone: I understand you comment. I think it's more or less the same because you can refactor one thing into the other. In this case I preferred to not mix the class and methods style and the struct and functions style, since this file in particular is totally implemented using the latter.

bassosimone:
Done!

AntonioLangiu:
Ok, thanks! merging!

PR #469, opened 2016-04-16
big cleanup: convert, remove http test cases

bassosimone:
The objective here is to avoid losing coverage but to arrive to
a point where there are no test cases for old HTTP API.

At that point, we can then safely remove such API without causing
any issue to the tree, because no users are left.

- convert Request() test cases to use request(), but skip for now
  the test cases where such conversion is not straightforward

- remove Stream() test cases that duplicate test cases that use
  the request() API (no point in converting them)

- note that there is more work to do


bassosimone:
No new Valgrind errors introduced by this cleanup. Merging!

PR #478, opened 2016-04-16
utils: remove unused code

bassosimone:


bassosimone:
Close because I was removing code used by traceroute (did not notice that because traceroute code is not compiled under OSX, which is the system I use on my laptop).

PR #482, opened 2016-04-16
valgrind build is not allowed to fail

AntonioLangiu:
valgrind build is not allowed to fail


.travis.yml
```diff
@@ -44,7 +44,6 @@ after_failure:
       fi
 
 matrix:
-  fast_finish: true
```
bassosimone: <bikeshedding>I'd leave `fast_finish` there, we will need it in the future, right?</>

bassosimone: Ah, never mind, we can always re-add it later.

bassosimone:
Merging!

PR #483, opened 2016-04-16
Make dns callback robust to exceptions

### LINKED ISSUE #481: Make dns_callback robust to exceptions
When the [dns call the user callback](https://github.com/measurement-kit/measurement-kit/blob/master/src/dns/query.hpp#L205) it doesn't catch the possible exception that the callback can throw!

AntonioLangiu:
As found by @bassosimone, we didn't consider the case a user callback throws an exception. 
I think this should fix the problem and catch all the exception. 

@bassosimone do you know how to test if the context has been really deleted? 

This closes #481

### DIFF
--- a/src/dns/query.hpp
+++ b/src/dns/query.hpp
@@ -203,13 +203,17 @@ static void dns_callback(int code, char type, int count, int ttl,
 
     context->message.answers =
             build_answers_evdns(code, type, count, ttl, addresses);
-    if (context->message.error_code != DNS_ERR_NONE) {
-        context->callback(mk::dns::dns_error(context->message.error_code),
-                context->message);
-    } else {
-        context->callback(NoError(), context->message);
+    try {
+        if (context->message.error_code != DNS_ERR_NONE) {
+            context->callback(mk::dns::dns_error(context->message.error_code),
+                    context->message);
+        } else {
+            context->callback(NoError(), context->message);
+        }
+    } catch (const std::exception& e) {
+        delete context;
+        throw;
     }
-
     delete context;
 }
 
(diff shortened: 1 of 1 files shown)


src/dns/query.hpp
```diff
@@ -201,13 +201,17 @@ static void dns_callback(int code, char type, int count, int ttl,
...
+    try {
+        if (context->message.error_code != DNS_ERR_NONE) {
+            context->callback(mk::dns::dns_error(context->message.error_code),
+                    context->message);
+        } else {
+            context->callback(NoError(), context->message);
+        }
+    } catch (const std::exception& e) {
+        delete context;
+        throw;
```
bassosimone: I am undecided whether we should propagate exceptions here. I think we should stop here some exceptions, namely the ones derived from `Error` that MK users would probably like to throw. Other exceptions should probably continue because typically they indicate either errors external to MK or very serious conditions.

bassosimone:
Moved to 0.3.0 because this is not priority for the upcoming release.

src/dns/query.hpp
```diff
@@ -203,13 +203,17 @@ static void dns_callback(int code, char type, int count, int ttl,
...
-    } else {
-        context->callback(NoError(), context->message);
+    try {
+        if (context->message.error_code != DNS_ERR_NONE) {
+            context->callback(mk::dns::dns_error(context->message.error_code),
+                    context->message);
+        } else {
+            context->callback(NoError(), context->message);
+        }
+    } catch (const Error& e) {
```
bassosimone: Good for [using const](http://stackoverflow.com/questions/2145147/why-catch-an-exception-as-reference-to-const)! We should use it in more places!

Issue #484, opened 2016-04-16
Increase robustness of s/n/connection lifecycle


bassosimone:
The idea here is to refactor Connection such that libevent's C code has somehow access to a Var<Connection> so that it's impossible that the Connection is destroyed while C is using it.

Another thing to implement is to avoid calling `bufferevent_free()` from a function called by a libevent callback, because we have seen that the stable libevent does not like that for SSL bufferevents.

bassosimone:
This should have been implemented in #490, #491, #492, #493, and #494.

PR #488, opened 2016-04-16
net: preparatory changes for #484

bassosimone:
1) make Connection() private and introduce a factory called make() to
   return a Var<Transport> instead and propagate changes

2) introduce static methods that shall deal with events generated
   by libevent and call existing non static methods

3) introduce safe_upcall() macro to implement said methods, for now
   with a very trivial implementation

4) add Poller * member which will be used soon

### DIFF
--- a/src/net/connection.hpp
+++ b/src/net/connection.hpp
@@ -4,6 +4,9 @@
 #ifndef SRC_NET_CONNECTION_HPP
 #define SRC_NET_CONNECTION_HPP
 
+#include "src/common/utils.hpp"
+#include "src/net/bufferevent.hpp"
+#include "src/net/emitter.hpp"
 #include <event2/bufferevent.h>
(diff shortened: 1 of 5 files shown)


src/net/connection.hpp
```diff
@@ -54,8 +61,27 @@ class Connection : public Emitter, public NonMovable, public NonCopyable {
...
+    static void emit_libevent_event_(Connection *me, short what) {
+        safe_upcall(me, handle_event_(what));
+    }
+    static void emit_libevent_read_(Connection *me) {
+        safe_upcall(me, handle_read_());
+    }
+    static void emit_libevent_write_(Connection *me) {
+        safe_upcall(me, handle_write_());
+    }
+#undef safe_upcall
```
AntonioLangiu: :+1:  cool

PR #489, opened 2016-04-16
net/connection: implement garbage collection

bassosimone:
The general idea is that a Connection contains a smart pointer
to itself and is constructed using a factory that returns a smart
pointer and saves such smart pointer inside the connection such
that reference counting is always at least one.

This models the fact that C code is using the Connection as long as
we don't change the opaque pointer stored inside the bufferevent.

Then, when we `close()` a connection we cancel all the C++ callbacks
to remove self references, and reset the bufferevent.

Finally, when a callback of bufferevent is invoked, it checks
the usage count and, if it is one, it means that no C++ code is
referencing the Connection. This fact triggers deletion.

Deletion is delayed. We do this because in stable libevent the SSL
bufferevent does not cope well with immediate deletion called by
a bufferevent callback, leading to memory error.

Thus, when deletion is needed, a poller call_soon is scheduled and
such call_soon will remove the latest reference, the self reference
created by the factory, thus triggering deletion of Connection.

Finally, the destructor of Connection also disposes of the bufferevent
that was passed as argument to the factory.

This diff makes net/bufferevent irrelevant, hence this diff
removes src/net/bufferevent.hpp, test/net/bufferevent.cpp


bassosimone:
I've found a better way to do this. I will probably close this tomorrow after the better solution is merged.

PR #492, opened 2016-04-17
connection: clear state inside close()

bassosimone:
While there also repair a failing DNS test which prevented the build to terminate.

### DIFF
--- a/src/net/connection.cpp
+++ b/src/net/connection.cpp
@@ -77,6 +77,14 @@ void Connection::close(std::function<void()> cb) {
         throw std::runtime_error("already closed");
     }
     isclosed = true;
+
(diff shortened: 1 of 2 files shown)


src/net/connection.cpp
```diff
@@ -77,6 +77,14 @@ void Connection::close(std::function<void()> cb) {
         throw std::runtime_error("already closed");
     }
     isclosed = true;
+
+    on_connect(nullptr);
+    on_data(nullptr);
+    on_flush(nullptr);
+    on_error(nullptr);
+    bufferevent_setcb(bev, nullptr, nullptr, nullptr, nullptr);
+    disable_read();
```
AntonioLangiu: :+1:

PR #496, opened 2016-04-17
build system: remove compile builtin feature

bassosimone:
I will further refine this. The general idea is that you should
have dependencies installed beforehand, except of course those
dependencies that we compile unconditionally with MK.

This simplifies a lot and it is good to do that, because we are
about to add LibreSSL as a dependency, which is huge and would
have complicated even more the build system.


AntonioLangiu:
is there a way to avoid adding every time the openssl path to configure while building with mac?

Issue #485, opened 2016-04-16
Add code to connect a SSL bufferevent


bassosimone:
This would be the basic building block required to implement HTTPS

bassosimone:
This has now been merged in #498

PR #498, opened 2016-04-17
Add initial SSL implementation

bassosimone:
This is very rough. In particular, we do not validate certificates, so
a man in the middle is very possible for now.

But I feel like it's important to have the basic scaffolding such
that things that should go in parallel over SSL can.

<strike>Crossing fingers as for Valgrind. I recall that used to reported
several uninitialized variables in OpenSSL.</strike>

### DIFF
--- a/src/net/ssl-context.hpp
+++ b/src/net/ssl-context.hpp
@@ -1,0 +1,29 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+#ifndef SRC_NET_SSL_CONTEXT_HPP
+#define SRC_NET_SSL_CONTEXT_HPP
+
+// Forward declarations
+struct ssl_st;
+struct ssl_ctx_st;
+
+namespace mk {
+namespace net {
+
+class SslContext {
+  public:
+    static ssl_ctx_st *get_client_context();
+
+    static ssl_st *get_client_ssl();
+
+  private:
+    SslContext();
+    ~SslContext();
+
+    ssl_ctx_st *ctx = nullptr;
+};
+
+} // namespace net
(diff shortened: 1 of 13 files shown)


bassosimone:
- [ ] review and merge pull request
- [ ] add "bug" issue saying that we should validate certificates

bassosimone:
Closes #485.

test/net/transport.cpp
```diff
@@ -47,6 +47,20 @@ TEST_CASE("net::connect() can connect to open port") {
...
 
+TEST_CASE("net::connect() can connect to SSL port") {
+    if (CheckConnectivity::is_down()) {
+        return;
+    }
+    set_verbose(1);
+    loop_with_initial_event([]() {
+        connect("nexa.polito.it", 443, [](Error error, Var<Transport> txp) {
+            REQUIRE(!error);
+            txp->close([]() { break_loop(); });
```
AntonioLangiu: :+1:

src/net/ssl-context.hpp
```diff
@@ -0,0 +1,29 @@
...
+#define SRC_NET_SSL_CONTEXT_HPP
+
+// Forward declarations
+struct ssl_st;
+struct ssl_ctx_st;
+
+namespace mk {
+namespace net {
+
+class SslContext {
```
AntonioLangiu: this should be not movable and not copyable.
we don't want to have copy of the singleton

bassosimone: I don't think you cannot easily get an instance of the class, given that the constructor is private, but I agree that for robustness we should declare it like so! Hold on, cooking the diff...

AntonioLangiu:
Ok, i tested it locally with OSX and with ubuntu trusty on vagrant (using valgrind). 
It's ok for me!

bassosimone:
Thanks for reviewing!

PR #505, opened 2016-04-19
Add static_pointer_cast to Var

AntonioLangiu:
As in the description!

### DIFF
--- a/test/common/var.cpp
+++ b/test/common/var.cpp
@@ -13,11 +13,18 @@
 
 using namespace mk;
 
-struct Foo {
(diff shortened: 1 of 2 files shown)


test/common/var.cpp
```diff
@@ -13,13 +13,20 @@
 
 using namespace mk;
 
-struct Foo {
+class Foo {
+ public:
```
bassosimone: hahaha, neat!

bassosimone:
I really look this! Thank you!

Issue #403, opened 2016-03-10
HTTP Invalid Request Line should be completed


hellais:
Currently the HTTP Invalid Request Line test doesn't check for tampering and doesn't generate all of the requests that are specified by OONI.

We should re-implement this based on a TCP test template instead of the HTTP test template as this gives us more freedom to read invalid HTTP responses.

hellais:
This has been merged into master. Closing!

PR #507, opened 2016-04-19
Convert HTTP Invalid Request Line test to use the TCPTest template

hellais:
This implements #403


src/ooni/http_invalid_request_line_impl.hpp
```diff
@@ -5,75 +5,107 @@
...
         // randomSTR(4) + " / HTTP/1.1\n\r"
-        request(
-            {
-                {"url", options["backend"]},
-                {"method", mk::random_str_uppercase(4)},
-                {"http_version", "HTTP/1.1"},
-            },
-            headers, "", handle_response);
+        std::string test_random_invalid_method(mk::random_str_uppercase(4));
+        test_random_invalid_method += " / HTTP/1.1\n\r";
```
bassosimone: Here, is it intended to use `\n\r` as opposed to `\r\n`?

hellais: Actually this is a bug in ooniprobe and the relative specification for the test.

For consistencies sake I have implemented this in the same way of how it's specified in ooni-probe (https://github.com/TheTorProject/ooni-probe/blob/master/ooni/nettests/manipulation/http_invalid_request_line.py#L75) and in ooni-spec (https://github.com/TheTorProject/ooni-spec/blob/master/test-specs/ts-007-http-invalid-request-line.md), but I think that eventually they should all be updated.

When they are updated they should probably also include a blank line at the end of them.

I will file a ticket about this, since otherwise I will forget.

bassosimone: ack!

bassosimone:
:+1: 

The diff looks good to me! I am going to spend the afternoon working on a paper that cannot be delayed further and then will regain some happiness later this afternoon by finish testing and merging this!

bassosimone:
Merging!

PR #509, opened 2016-04-19
Add ip lookup function

AntonioLangiu:
Add an ip lookup function for the ooni tests!


test/ooni/utils.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+#define CATCH_CONFIG_MAIN
+#include "src/ext/Catch/single_include/catch.hpp"
+
+#include "src/ooni/utils.hpp"
+#include <measurement_kit/common.hpp>
+#include <measurement_kit/http.hpp>
+
+TEST_CASE("ip lookup works") {
+    mk::loop_with_initial_event([]() {
```
hellais: Should probably add a check for `CheckConnectivity::is_down()`

AntonioLangiu: yes, thank you :+1:

PR #511, opened 2016-04-20
Add geoip

AntonioLangiu:
In this PR I'm adding:

- geoip library
- a function to get a json containing geoip informations given an ip!

Actually the build fails because we should somehow download or add to the repository the geoip database.
Maybe we can search for a debug version of it!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,7 +64,6 @@ src/ext/http-parser/
 src/ext/jansson/
 src/ext/json/
 src/ext/libevent/
-src/ext/libmaxminddb/
 stamp-h1
 test-driver
 test/common/async
--- a/.gitignore
+++ b/.gitignore
@@ -61,9 +61,7 @@ missing
 site
 src/ext/Catch/
 src/ext/http-parser/
-src/ext/jansson/
 src/ext/json/
-src/ext/libevent/
 stamp-h1
 test-driver
 test/common/async
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,8 @@ test/common/var
 test/dns/defines
 test/dns/error
 test/dns/query
+test/fixtures/GeoIP.dat
+test/fixtures/GeoIPASNum.dat
 test/http/client
 test/http/funcs
 test/http/parse_url
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -109,7 +109,7 @@ AC_DEFUN([MK_AM_GEOIP], [
 
   mk_not_found=""
   AC_CHECK_HEADERS(GeoIP.h, [], [mk_not_found=1])
-  AC_CHECK_LIB(geoip, GeoIP_open, [], [mk_not_found=1])
+  AC_CHECK_LIB(GeoIP, GeoIP_open, [], [mk_not_found=1])
 
   if test "$mk_not_found" = "1"; then
     AC_MSG_WARN([Failed to find dependency: geoip])
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -86,7 +86,7 @@ AC_DEFUN([MK_AM_GEOIP], [
 
   if test "$mk_not_found" = "1"; then
     AC_MSG_WARN([Failed to find dependency: geoip])
-    echo "    - to install on Debian: sudo apt-get install libgeoip1 libgeoip-dev geoip-bin"
+    echo "    - to install on Debian: sudo apt-get install libgeoip-dev"
     echo "    - to install on OSX: brew install libgeoip"
     echo "    - to compile from sources: ./build/dependency geoip"
     AC_MSG_ERROR([Please, install geoip and run configure again])
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -67,33 +67,6 @@ AC_DEFUN([MK_AM_JANSSON], [
   echo ""
 ])
 
-AC_DEFUN([MK_AM_LIBMAXMINDDB], [
-  echo "> checking for dependency: libmaxminddb"
-
-  AC_ARG_WITH([libmaxminddb],
-              [AS_HELP_STRING([--with-libmaxminddb],
-                [GeoIP library @<:@default=check@:>@])
-              ],
-              [
-                CPPFLAGS="$CPPFLAGS -I$withval/include"
-                LDFLAGS="$LDFLAGS -L$withval/lib"
-              ],
-              [])
-
-  mk_not_found=""
-  AC_CHECK_HEADERS(maxminddb.h, [], [mk_not_found=1])
-  AC_CHECK_LIB(maxminddb, MMDB_open, [], [mk_not_found=1])
-
-  if test "$mk_not_found" = "1"; then
-    AC_MSG_WARN([Failed to find dependency: libmaxminddb])
-    echo "    - to install on Debian: sudo apt-get install libmaxminddb-dev"
-    echo "    - to install on OSX: brew install libmaxminddb"
-    echo "    - to compile from sources: ./build/dependency libmaxminddb"
-    AC_MSG_ERROR([Please, install libmaxminddb and run configure again])
-  fi
-  echo ""
-])
-
 AC_DEFUN([MK_AM_GEOIP], [
   echo "> checking for dependency: geoip"
 
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -94,6 +94,33 @@ AC_DEFUN([MK_AM_LIBMAXMINDDB], [
   echo ""
 ])
 
+AC_DEFUN([MK_AM_GEOIP], [
+  echo "> checking for dependency: geoip"
+
+  AC_ARG_WITH([geoip],
+              [AS_HELP_STRING([--with-geoip],
+                [GeoIP OLD API library @<:@default=check@:>@])
+              ],
+              [
+                CPPFLAGS="$CPPFLAGS -I$withval/include"
+                LDFLAGS="$LDFLAGS -L$withval/lib"
+              ],
+              [])
+
+  mk_not_found=""
+  AC_CHECK_HEADERS(GeoIP.h, [], [mk_not_found=1])
+  AC_CHECK_LIB(geoip, GeoIP_open, [], [mk_not_found=1])
+
+  if test "$mk_not_found" = "1"; then
+    AC_MSG_WARN([Failed to find dependency: geoip])
+    echo "    - to install on Debian: sudo apt-get install libgeoip1 libgeoip-dev geoip-bin"
+    echo "    - to install on OSX: brew install libgeoip"
+    echo "    - to compile from sources: ./build/dependency geoip"
+    AC_MSG_ERROR([Please, install geoip and run configure again])
+  fi
+  echo ""
+])
+
 AC_DEFUN([MK_AM_OPENSSL], [
   echo "> checking for dependency: openssl"
 
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -72,7 +72,7 @@ AC_DEFUN([MK_AM_GEOIP], [
 
   AC_ARG_WITH([geoip],
               [AS_HELP_STRING([--with-geoip],
-                [GeoIP OLD API library @<:@default=check@:>@])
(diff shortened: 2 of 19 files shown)


src/ooni/utils.hpp
```diff
@@ -5,34 +5,86 @@
...
+                    callback(GenericError(), "");
+                    return;
+                }
+                callback(NoError(), m[1]);
+            },
+            {}, "", {}, Logger::global(), Poller::global());
+}
+
+// this is needed to return a const char instead of a char for asn resolver
+inline GEOIP_API const char *
```
bassosimone: Here GEOIP_API is not needed. (This GEOIP_API is a compiler definition used when building GeoIP that basically tells the compiler that the following function shall be exported. It is typically a no op on Unix systems and has meaning on Windows. As such, it may even break our build if we attempt building on Win).

bassosimone: (Apart from that, neat trick of the inline function. I like it.)

src/ooni/utils.hpp
```diff
@@ -5,34 +5,86 @@
...
+std::string geoip_query(std::string ip, std::string path) {
+    GeoIP *gi;
+    GeoIPLookup gl;
+    gi = GeoIP_open(path.c_str(), GEOIP_MEMORY_CACHE);
+    if (gi == nullptr) {
+        return "";
+    }
+    const char *result;
+    result = resolver(gi, ip.c_str(), &gl);
+    if (result == NULL) {
```
bassosimone: Here you should close the geoip handle, otherwise we're leaking it.

src/ooni/utils.hpp
```diff
@@ -5,34 +5,86 @@
...
+// this is needed to return a const char instead of a char for asn resolver
+inline GEOIP_API const char *
+MK_GeoIP_name_by_name_gl(GeoIP *gi, const char *host, GeoIPLookup *gl) {
+    return (const char *)GeoIP_name_by_name_gl(gi, host, gl);
+}
+
+template <decltype(GeoIP_country_code3_by_name_gl) resolver>
+std::string geoip_query(std::string ip, std::string path) {
+    GeoIP *gi;
+    GeoIPLookup gl;
```
bassosimone: Shouldn't this variable be initialized to something? That is, is this variable in-out or only out?

src/ooni/utils.hpp
```diff
@@ -5,34 +5,86 @@
...
+    if (gi == nullptr) {
+        return "";
+    }
+    const char *result;
+    result = resolver(gi, ip.c_str(), &gl);
+    if (result == NULL) {
+        return "";
+    }
+    GeoIP_delete(gi);
+    return std::string(result);
```
bassosimone: The `std::string()` this is not necessary. If you state that the function returns `std::string` and you return a `const char *`, conversion is done automatically. (The zen of Python says that "explicit is better than implicit", but here we have a compiler controlling this kind of things for us, so I'd say it's ok to err on the side of tersness.)

src/ooni/utils.hpp
```diff
@@ -5,34 +5,86 @@
...
+}
+
+inline ErrorOr<Json> geoip(std::string ip, std::string path_country,
```
bassosimone: If there's no need to templatize this function, I'd rather not inline it and have it in `ooni/utils.cpp`.

src/ooni/utils.hpp
```diff
@@ -5,34 +5,86 @@
...
+        return GenericError();
+    }
+    json["country_name"] = result;
+
+    result = geoip_query<MK_GeoIP_name_by_name_gl>(ip, path_asn);
+    if (result == "") {
+        return GenericError();
+    }
+    json["asn"] = result;
+    return ErrorOr<Json>(json);
```
bassosimone: Here you don't need `ErrorOr<Json>(json)`. Just `return json` as I've beefed up the definition of `ErrorOr` to automatically infer when it's used in such a context.

bassosimone:
@AntonioLangiu:

- [ ] in Debian there is a [geoip-database](https://packages.debian.org/it/sid/geoip-database) package. Perhaps there is also one in Ubuntu. If so, guess we can install the package during travis setup and then point our code to such database for running tests

- [ ] let's also remind to get rid of libmaxminddb, since we've decided to use libgeoip in all cases, given that there are no free ASN databases for libmaxmindbd. Perhaps we could do also this in the context of this pull request?

- [ ] I've also added inline comments above of which you should take care of before proceeding

acinclude.m4
```diff
@@ -94,6 +94,33 @@ AC_DEFUN([MK_AM_LIBMAXMINDDB], [
...
+              ],
+              [])
+
+  mk_not_found=""
+  AC_CHECK_HEADERS(GeoIP.h, [], [mk_not_found=1])
+  AC_CHECK_LIB(geoip, GeoIP_open, [], [mk_not_found=1])
+
+  if test "$mk_not_found" = "1"; then
+    AC_MSG_WARN([Failed to find dependency: geoip])
+    echo "    - to install on Debian: sudo apt-get install libgeoip1 libgeoip-dev geoip-bin"
```
bassosimone: I guess `geoip-bin` is not needed and `libgeoip1` is implied by `libgeoip-dev`, correct?

src/ooni/utils.hpp
AntonioLangiu: GeoIPLookup is [a struct defined in GeoIP.h](https://github.com/maxmind/geoip-api-c/blob/master/libGeoIP/GeoIP.h#L80). We need it because we are using the new api to do queries (because the old [are deprecated](https://github.com/maxmind/geoip-api-c/blob/master/libGeoIP/GeoIP.h#L378)), but I was not able to find examples for the new apis. 

The struct GeoIPLookup contains an integer that should be the netmask. I don't know actually if it is just an out variable or also an in-out variable, and I was not able to find documentation for this. @bassosimone

acinclude.m4
AntonioLangiu: I will try it with vagrant!

src/ooni/utils.hpp
AntonioLangiu: I have an answer now! It is just an [out variable](https://github.com/maxmind/geoip-api-c/blob/v1.6.9/libGeoIP/GeoIP.c#L1413), at least in the functions we use! (I didn't check the whole library)

AntonioLangiu:
Now this PR should be ready. I removed libmaxminddb and I fixed the build on travis and on vagrant (the build/vagrand/ script).

src/ooni/utils.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "src/ooni/utils.hpp"
+
+namespace mk {
+namespace ooni {
+
+ErrorOr<Json> geoip(std::string ip, std::string path_country,
+                    std::string path_asn) {
+    Json json;
+    GeoIP *gi;
+    GeoIPLookup gl;
```
bassosimone: Please, add a comment stating that we know that opening the database every time *may* be inefficient and that, in such case, we will consider keeping it open.

src/ooni/utils.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    gi = GeoIP_open(path_country.c_str(), GEOIP_MEMORY_CACHE);
+    if (gi == nullptr) {
+        GeoIP_delete(gi);
```
bassosimone: Here you should not pass `nullptr` to `GeoIP_delete()`. No need to free resources in this error path because we are in the case where no resources have actually been allocated.

src/ooni/utils.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if (result == nullptr) {
+        GeoIP_delete(gi);
+        return GenericError();
+    }
+    json["country_name"] = result;
+    GeoIP_delete(gi);
+
+    gi = GeoIP_open(path_asn.c_str(), GEOIP_MEMORY_CACHE);
+    if (gi == nullptr) {
+        GeoIP_delete(gi);
```
bassosimone: Same here.

src/ooni/utils.cpp
```diff
@@ -0,0 +1,57 @@
...
+        return GenericError();
+    }
+    char *res;
+    res = GeoIP_name_by_name_gl(gi, ip.c_str(), &gl);
```
bassosimone: Don't we have an initial value with which to initialize `gl`? This is one of the classic places where `valgrind` would infer that a variable has been used without being initialized first.

src/ooni/utils.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if (res == nullptr) {
+        GeoIP_delete(gi);
+        return GenericError();
+    }
+    json["asn"] = res;
+    free(res);
+    GeoIP_delete(gi);
+
+    return json;
+}
```
bassosimone: As a more general comment: I don't like the way in which this function is implemented. I'd rather prefer having two separate functions each one doing a specific thing with a specific database. Then, for convience, it's ok to me if you have a single function implemented in the terms of those two functions, which could even be static given that we don't need to expose them.

src/ooni/utils.hpp
```diff
@@ -5,36 +5,44 @@
...
 #define SRC_OONI_UTILS_HPP
 
+#include "src/ext/json/src/json.hpp"
+#include <GeoIP.h>
 #include <measurement_kit/common.hpp>
 #include <measurement_kit/http.hpp>
 #include <regex>
 #include <string>
 
+using Json = nlohmann::json;
```
bassosimone: I understand why here you are using uppercase. I guess it's for uniformity with other places in MK where classes begin with uppercase, right? Yet, when I encountered `Json` in uppercase this stroke me, because I knew the class in `nlohmann::json` is lowercase. Given that, I'd keep it lowercase. Makes sense?

.gitignore
```diff
@@ -64,7 +64,6 @@ src/ext/http-parser/
 src/ext/jansson/
 src/ext/json/
 src/ext/libevent/
-src/ext/libmaxminddb/
```
bassosimone: Since some more changes are to be performed in any case, would you mind making sure that we don't list in `.gitignore` stuff that - after recent changes - we don't put in `src/ext` anymore?

build/spec/all
```diff
@@ -2,6 +2,6 @@
 set -e
 pkg_rootdir=$(cd $(dirname $(dirname $(dirname $0))) && pwd -P)
 cd $pkg_rootdir
-for dependency in jansson libevent libmaxminddb libressl; do
+for dependency in jansson libevent libressl; do
```
bassosimone: Shouldn't we add `geoip` here?

build/ios/scripts/build_arch.sh
```diff
@@ -42,7 +42,6 @@ export CXXFLAGS="-isysroot ${DEVELOPER}/Platforms/${PLATFORM}.platform/Developer
                 --disable-examples \
                 --with-libevent=builtin \
                 --with-jansson=builtin \
-                --with-libmaxminddb=builtin \
```
bassosimone: Shouldn't we add `--with-geoip=builtin` here?

build/ios/scripts/build_frameworks.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 ROOTDIR=$(cd $(dirname $0) && pwd -P)
-EXTLIBRARIES="libevent libevent_pthreads libjansson libmaxminddb"
+EXTLIBRARIES="libevent libevent_pthreads libjansson"
```
bassosimone: Shouldn't geoip be a framework?

bassosimone:
Apart from two residual doubts indicated by comments above, everything else looks good to me now!

build/ios/scripts/build_arch.sh
AntonioLangiu: Yes

build/ios/scripts/build_frameworks.sh
AntonioLangiu: yes

AntonioLangiu:
Ok I solved the last issues.

bassosimone:
Thanks, and very good work! Now merging!

Issue #500, opened 2016-04-17
Validate certificates in SSL implementation


AntonioLangiu:
As in [this PR](https://github.com/measurement-kit/measurement-kit/pull/498).

PR #512, opened 2016-04-21
Add support for verifying SSL certificates

hellais:
In here we add support for validating SSL certificates. The root CAs to be used for verifying the certificate chain are passed using either the transport Settings option, or via a new global() options.

What is supported:

* Validation of hostnames via CN and SAN certificate attributes (including wildcard certificates)

* SNI to have the server provide different certs depending on the hostname being requested

What is not supported:

* Validation of IPv4 or IPv6 IP addresses in the SAN field.

Most of the certificate validation code is patchwork between [this](https://github.com/iSECPartners/ssl-conservatory/blob/master/openssl/openssl_hostname_validation.c) and [this](https://github.com/libressl-portable/openbsd/blob/eea365c0133c49cf771f299bdcfe1dcfc7950556/src/lib/libtls/tls_verify.c)


hellais:
This implements what is outlined in issues https://github.com/measurement-kit/measurement-kit/issues/500.

I am still not fully satisfied of the overall quality of this branch, but would greatly appreciate feedback on how to improve it.

src/net/ssl-context.cpp
```diff
@@ -7,43 +7,65 @@
...
     SSL_load_error_strings();
     OpenSSL_add_all_algorithms();
 
-    // FIXME: the code in here is really simple because we only aim to
-    // smoke test the OpenSSL functionality. What is missing here is at
-    // the minimum code to verify the peer certificate. Also we should
-    // not allow SSLv2 (and probably v3) connections, and we should also
-    // probably set SNI for the other end to respond correctly.
-    ctx = SSL_CTX_new(SSLv23_client_method());
+    Logger::global()->debug("ssl: creating ssl context with bundle %s", ca_bundle_path.c_str());
```
bassosimone: In place of `Logger::global()->debug()` you can just use `debug()` which is syntactic sugar from the former expression (and defined if you pull `measurement_kit/common.hpp`.

src/net/ssl-context.cpp
```diff
@@ -7,43 +7,65 @@
...
     }
+    init(ca_bundle_path);
+}
+
+SslContext::SslContext(std::string ca_bundle_path) {
+    init(ca_bundle_path);
 }
 
 SslContext::~SslContext() {
+    EVP_cleanup();
```
bassosimone: Is this function sort of a global destructor for OpenSSL's EVP library? If so, is it safe to call such a function before, rather than after, `SSL_CTX_free()`? (Asking because to me it would be more natural to start with `SSL_CTX_free()` and then move on with `EVP_cleanup()`)

src/net/ssl-validate.cpp
```diff
@@ -0,0 +1,170 @@
...
+namespace mk {
+namespace net {
+
+typedef enum {
+    MatchFound,
+    MatchNotFound,
+    NoSANPresent,
+    MalformedCertificate,
+    ValidationError
+} HostnameValidationResult;
```
bassosimone: In C++ it would be enough to write `typedef HostnameValidationResult`

src/net/ssl-validate.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+        if (current_name->type == GEN_DNS) {
+            // Current name is a DNS name, let's check it
+            char *dns_name = (char *)ASN1_STRING_data(current_name->d.dNSName);
+
+            // Make sure there isn't an embedded NUL character in the DNS name
+            if (ASN1_STRING_length(current_name->d.dNSName) != strlen(dns_name)) {
+                result = MalformedCertificate;
+                break;
+            } else { // Compare expected name with the DNS name
```
bassosimone: I think here you don't need an `else` branch given that the above branch is exited using `break`

src/net/ssl-validate.cpp
```diff
@@ -0,0 +1,170 @@
...
+                if (match_name(dns_name, name) == 0) {
+                    result = MatchFound;
+                    break;
+                }
```
bassosimone: If I'm understanding correctly, this is the place where you patched the original `iSECPartners` code to use the piece of code extracted from libressl, correct?

Vagrantfile
```diff
@@ -1,14 +1,16 @@
 Vagrant.configure(2) do |config|
+  config.vm.synced_folder ".", "/mk"
```
bassosimone: Afaik, Vagrant already synces this directory as `/vagrant`. Would this cause `.` to appear twice or would this just cause `.` to appear once as `/mk`. In the latter case, I don't have any objections to this change (I actually find it an improvement).

src/net/connect.cpp
```diff
@@ -28,6 +30,30 @@ void mk_bufferevent_on_event(bufferevent *bev, short what, void *ptr) {
     delete cb;
 }
 
+void mk_bufferevent_on_event_ssl(bufferevent *bev, short what, void *ptr) {
+    // XXX: Maybe this can be refactored a bit to have less duplication with
+    // the above function.
```
bassosimone: To do that, which I think would be neat, I suggest to refactor `mk_bufferevent_on_event` such that, in the error paths, the bufferevent is freed after the callbacks are called and to pass such bufferevents to the callee for further inspection. With this change in, you can refactor the piece of code that checks for invalid certificate to stay in the callback (i.e., rather than just checking `if (err)` you can also check if more specifically there was an openssl error and, in such case, be more informative with the caller.

src/net/ssl-validate.cpp
```diff
@@ -0,0 +1,170 @@
...
+    // Compare expected name with the CN
+    if (match_name(common_name_str, name) == 0) {
+        return MatchFound;
+    }
+    else {
+        return MatchNotFound;
+    }
+}
+
+Error ssl_validate_hostname(std::string hostname, const X509 *server_cert) {
```
bassosimone: Question: have you checked whether the rest of the code in `tls_verify.c` of libressl is functionally equivalent to the `iSECPartners` code we imported? I was wondering whether it would be possible to directly drop the whole `tls_verify.c` in `src/ext` (perhaps `src/ext/libtls`?) and then here just call its public `tls_ceck_name()` api. I think this change could be an improvement because it would allow us to more easily track future updates in the `libtls` code base. What do you think?

bassosimone:
> What is not supported:
>
>  -  Validation of IPv4 or IPv6 IP addresses in the SAN field.

If I read correctly libtls code, they implement this feature. So, if it's safe to entirely use libtls as I suggested above, that would perhaps be another argument in favor of doing that.

> I am still not fully satisfied of the overall quality of this branch, but would greatly appreciate feedback on how to improve it.

I have reviewed the patch with great attention, trying to provide suggestions on how it could be improved in places you tagged for possible improvements. Overall, I am very happy of what I have seen and think that in one more iteration of review we would be able to have good quality certificate verification in MK :-).

src/net/ssl-context.cpp
hellais: In theory it should be fine as long as nobody is using the context to perform crypto operations. EVP_cleanup only deals with data structures related to the OpenSSL ciphers and hashes so it doesn't actually touch the context.

That said it is also a deprecated function since OpenSSL 1.1.0. Which version of OpenSSL are we targeting?

Vagrantfile
hellais: Yes I can confirm that this leads to not synching the CWD in the `/vagrant` path anymore, but only in the `/mk` path.

src/net/ssl-validate.cpp
hellais: I took a look at tls_verify.c and it seems like it's all functionally equivalent to the code by iSecPartners. It actually is even IP address aware, so I think it's a good idea to drop our code in favour of tls_verify.c.

I will proceed in doing so.

src/net/ssl-context.cpp
bassosimone: For the apps, we are using LibreSSL because it is simpler to compile and because they appreciate the team's focus on proactive security.

LibreSSL is a fork of OpenSSL 1.0.1g, but afaik LibreSSL aims to stay as much compatible as possible in API (and probably also in ABI) with OpenSSL.

So, probably, when EVP_cleanup() would be removed in OpenSSL it would also be removed in LibreSSL.

In any case, for now, I'd say let's leave EVP_cleanup() there. It won't hurt.

Vagrantfile
bassosimone: Okay, perfect!

Vagrantfile
```diff
@@ -1,14 +1,16 @@
...
   config.vm.define "trusty64" do |trusty64|
     trusty64.vm.box = "ubuntu/trusty64"
     trusty64.vm.provider "virtualbox" do |v|
       v.memory = 2048
     end
     trusty64.vm.provision "shell", inline: <<-SHELL
+      sudo apt-add-repository -y ppa:maxmind/ppa
```
bassosimone: This is not necessary anymore. In master we have now removed libmaxminddb. I will revert this specific hunk when I am going to merge the diff in master.

Vagrantfile
```diff
@@ -1,14 +1,16 @@
...
       sudo apt-get update
       sudo apt-get dist-upgrade -y
       sudo apt-get install -y g++ autoconf automake libtool valgrind git
-      sudo apt-get install -y libevent-dev libssl-dev libgeoip-dev libjansson-dev
+      sudo apt-get install -y libevent-dev libssl-dev libjansson-dev libmaxminddb-dev
```
bassosimone: Likewise, here we should use `libgeoip-dev` rather than `libmaxminddb` and likewise I am going to take care of this minor edit as part of merging the pull request.

src/ext/tls_internal.h
```diff
@@ -0,0 +1,21 @@
...
+#ifndef HEADER_TLS_INTERNAL_MOCK_H
+#define HEADER_TLS_INTERNAL_MOCK_H
+
+#include "tls_verify.h"
+
+/* Empty macro definitions to mock out the tls_set_error* functions
+ */
+
+#define tls_set_errorx(ctx, errmsg, name) {}
+#define tls_set_error(ctx, errmsg, name) {}
```
bassosimone: Strictly speaking, these are not empty macros. Rather they expand to `{}`. I would convert them as empty macro as part of merging the pull request.

src/net/connect.hpp
```diff
@@ -68,10 +70,15 @@ void connect_base(std::string address, int port, Callback<bufferevent *> cb,
         return;
     }
 
-    // WARNING: set callbacks after connect() otherwise we free `bev` twice
-    // NOTE: In case of `new` failure we let the stack unwind
```
bassosimone: I don't think we should remove these two comments. The former is useful to explain why the code structure is laid out like this and why other code structures would lead to memory errors.

The latter justifies the way in which we deal with out of memory.

src/net/ssl-context.cpp
```diff
@@ -7,43 +7,65 @@
...
+    SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, NULL);
+}
+
+SslContext::SslContext() {
+    std::string ca_bundle_path;
+    Settings *global_settings = Settings::global();
+    if (global_settings->find("ca_bundle_path") != global_settings->end()) {
+        ca_bundle_path = (*global_settings)["ca_bundle_path"];
+    } else {
+        // XXX should we add other system default locations in here?
```
bassosimone: Perhaps. Since for now our focus is on mobile apps, I'd say we can leave the comment in place and we would deal with this problem later in the future.

src/net/transport.cpp
```diff
@@ -33,8 +33,16 @@ void connect(std::string address, int port,
...
         if (settings.find("ssl") != settings.end()) {
-            connect_ssl(r->connected_bev, SslContext::get_client_ssl(),
-                        [=](Error err, bufferevent *bev) {
+            Var<SslContext> ssl_context;
+            if (settings.find("ca_bundle_path") != settings.end()) {
+                logger->debug("ssl: using custom ca_bundle_path");
+                ssl_context = Var<SslContext>(new SslContext(settings.at("ca_bundle_path")));
+            } else {
+                logger->debug("ssl: using default context");
+                ssl_context = Var<SslContext>(new SslContext());
```
bassosimone: We could probably add a `SslContext::global()` factory that returns `Var<SslContext>` where the return value here is a singleton `Var<SslContext>` instantiated on the first call.

Saying this because, IIRC, construction of a new `SslContext` is not a lightweight operation.

What do you think?

src/net/connect.cpp
```diff
@@ -18,12 +19,9 @@ void mk_bufferevent_on_event(bufferevent *bev, short what, void *ptr) {
     if ((what & BEV_EVENT_CONNECTED) != 0) {
         (*cb)(mk::NoError(), bev);
     } else if ((what & BEV_EVENT_TIMEOUT) != 0) {
-        bufferevent_free(bev);
-        (*cb)(mk::net::TimeoutError(), nullptr);
+        (*cb)(mk::net::TimeoutError(), bev);
     } else {
-        bufferevent_free(bev);
-        // TODO: here we should map to the actual error that occurred
```
bassosimone: I don't think this comment should be removed. It's here to remind us that we should portably access the error that occurred and accordingly map it onto our corresponding error.

Or, if you prefer the approach, we could modify `NetworkError` such that, when instantiated, it performs this operation. In that case the comment should probably be moved near `NetworkError`.

src/net/ssl-context.cpp
```diff
@@ -7,43 +7,65 @@
...
-    ctx = SSL_CTX_new(SSLv23_client_method());
+    debug("ssl: creating ssl context with bundle %s", ca_bundle_path.c_str());
+
+    ctx = SSL_CTX_new(TLSv1_client_method());
     if (ctx == nullptr) {
+        debug("ssl: failed to create SSL_CTX");
         throw std::exception();
     }
+
+    if (SSL_CTX_load_verify_locations(ctx, ca_bundle_path.c_str(), NULL) == 0) {
```
bassosimone: I'd prefer if this would be `!= 1`. With C libraries I am always afraid they would agument semantic adding a novel error code in the future (like adding `-1` to distinguish from `0`), thus I prefer the code to be very defensive and basically check "if it's different from what the expected positive outcome would be".

src/ext/tls_internal.h
bassosimone: Another, perhaps more modern approach would be making them empty inline functions.

src/net/connect.cpp
```diff
@@ -157,8 +156,44 @@ void connect_ssl(bufferevent *orig_bev, ssl_st *ssl,
     }
 
     bufferevent_setcb(bev, nullptr, nullptr, mk_bufferevent_on_event,
-            new Callback<bufferevent *>([cb, logger](Error err, bufferevent *bev) {
+            new Callback<bufferevent *>([cb, logger, ssl, hostname](Error err, bufferevent *bev) {
```
bassosimone: I don't think `ssl` should be passed in the lambda capture list, as we get it from the bufferevent below.

src/net/connect.cpp
```diff
@@ -157,8 +156,44 @@ void connect_ssl(bufferevent *orig_bev, ssl_st *ssl,
...
+                    cb(SSLInvalidCertificateError(X509_verify_cert_error_string(verify_err)), nullptr);
+                    return;
+                }
+
+                if (err) {
+                    logger->debug("error in connection.");
+                    bufferevent_free(bev);
+                    cb(err, nullptr);
+                    return;
+                }
```
bassosimone: I'd refactor this `if (err) {` block before the above block, such that first we check for error and then we see what LibreSSL has to tell us. I've checked the code, and in the current code having the blocks are they are is not an error, but in any case I'd be more happy having the check for error at the beginning as I think structuring the code like this is more future proof.

bassosimone:
Awesome improvements! I've pinpointed a couple of improvements that I think should be committed before finally merging this piece of great work!

src/net/connect.cpp
hellais: Yes you are right and no I don't think it's wise to change the constructor of `NetworkError` as it would make it not aligned to the constructors of the other `Error` subclasses.

I will put this comment back then.

src/net/connect.cpp
hellais: Good catch.

src/net/connect.cpp
hellais: Sounds reasonable.

src/net/connect.hpp
hellais: Yes you are right, somehow I thought that in refactoring to do the free inside of the callback we would have ended up avoiding the double free, but you are right that this issue still would arise.

src/net/ssl-context.cpp
hellais: Yes I do see your point. Will do.

src/net/transport.cpp
hellais: Yeah you are right. I will make a static method that implements the singleton pattern.

src/net/ssl-context.cpp
```diff
@@ -39,14 +41,21 @@ void SslContext::init(std::string ca_bundle_path) {
...
     };
 
     SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, NULL);
 }
 
+Var<SslContext> SslContext::global() {
+    if (global_ == nullptr){
+        global_ = new SslContext();
+    }
+    return Var<SslContext>(global_);
```
bassosimone: This is not correct: calling `Var<SslContext>(global_)` you create a smart point with reference counting equal to one and associated pointer `global_`. The returned `Var<>` can be copied but, when it and all its copies go out of scope, this leads to `delete global_` being called. I suggest instead this pattern for the singleton:

```
Var<SslContext> SslContext::global() {
    static Var<SslContext> singleton(new SslContext);
    return singleton;
}```

This works because, since `singleton` is static, it is initialized only once the first time that `global()` is called.

File: src/net/connect.cpp [hunk 20]
```diff
@@ -156,22 +157,19 @@ void connect_ssl(bufferevent *orig_bev, ssl_st *ssl,
...
-                }
 
                 if (err) {
                     logger->debug("error in connection.");
+                    long verify_err = SSL_get_verify_result(ssl);
+                    if (verify_err != X509_V_OK) {
+                        debug("ssl: got an invalid certificate");
+                        err = SSLInvalidCertificateError(X509_verify_cert_error_string(verify_err));
+                    }
                     bufferevent_free(bev);
```
bassosimone: Question: could it happen that `err` evaluates to false but still certificate verification failed? Asking out of ignorance because I do not fully understand how this specific bit of OpenSSL works!

bassosimone: Uhm, perhaps this diff was updated while I was commenting it :-D

src/net/ssl-context.cpp
hellais: Hum, yes I do see your point here. We would not be gaining so much in terms of times a new instance is created. I will refactor this to the other singleton pattern you suggest.

bassosimone:
Excellent! I am going to merge this into master!

PR #513, opened 2016-04-22
Add libressl patch 

AntonioLangiu:
Add libressl patch to build just the target we need.
This patch needs to be applied to Makefile.am, and so I moved the patching part of the script after the autogen.sh


bassosimone:
Thank you! I know that writing such a simple diff was actually much harder than it seems because you needed to research how things are working, so well done!

PR #514, opened 2016-04-22
Ios build improvement

AntonioLangiu:
In this PR I fixed the ios build script. Now it's able to build all the dependencies, but it is not complete. We must fix the `build/dependency` script before we can complete the ios build script!

### DIFF
--- a/build/ios/scripts/build.sh
+++ b/build/ios/scripts/build.sh
@@ -12,9 +12,9 @@ echo "Building for this architectures: $ARCHS"
 
 for ARCH in ${ARCHS}; do
     if [ "${ARCH}" == "i386" ] || [ "${ARCH}" == "x86_64" ]; then
-        PLATFORM="iPhoneSimulator"
+        PLATFORM="iphonesimulator"
     else
-        PLATFORM="iPhoneOS"
+        PLATFORM="iphoneos"
     fi
     $ROOTDIR/build_arch.sh ${PLATFORM} ${ARCH}
 done
--- a/build/ios/scripts/build_arch.sh
+++ b/build/ios/scripts/build_arch.sh
@@ -11,17 +11,14 @@ PLATFORM=$1
 ARCH=$2
 
 OPT_FLAGS="-Os -g3"
-MAKE_JOBS=16
+MAKE_JOBS=2
 
 ROOTDIR=$(cd $(dirname "$0") && pwd -P)
-echo $ROOTDIR
 SOURCEDIR=$(cd "${ROOTDIR}/../../../" && pwd -P)
-echo $SOURCEDIR
 BUILDDIR=$(cd "${ROOTDIR}/../" && pwd -P)
-echo $BUILDDIR
+
 mkdir -p "$BUILDDIR/build/${PLATFORM}/${ARCH}/"
 DESTDIR=$(cd "$BUILDDIR/build/${PLATFORM}/${ARCH}/" && pwd -P)
-echo $DESTDIR
 
 
 MIOS_VERSION="7.1"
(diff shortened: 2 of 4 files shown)


build/ios/scripts/build_arch.sh
```diff
@@ -10,44 +10,69 @@ fi
...
-if [ "$PLATFORM" = "iPhoneOS" ]; then
-    EXTRA_CONFIG="--host=arm-apple-darwin14 --target=arm-apple-darwin14"
-fi
+OPT_FLAGS="-Os -g3"
+MAKE_JOBS=16
+
+ROOTDIR=$(cd $(dirname "$0") && pwd -P)
+echo $ROOTDIR
+SOURCEDIR=$(cd "${ROOTDIR}/../../../" && pwd -P)
+echo $SOURCEDIR
```
AntonioLangiu: This are just for debug! They should be removed!

build/spec/all
```diff
@@ -2,6 +2,6 @@
 set -e
 pkg_rootdir=$(cd $(dirname $(dirname $(dirname $0))) && pwd -P)
 cd $pkg_rootdir
-for dependency in jansson libevent libmaxminddb libressl; do
+for dependency in libressl jansson libevent libmaxminddb; do
```
bassosimone: Can we keep libraries sorted?

bassosimone:
I'd be very happy if this diff could be trimmed down so to avoid unnecessary changes into the diff, such that it's apparent what was actually changed!

build/ios/scripts/build.sh
```diff
@@ -12,9 +12,9 @@ echo "Building for this architectures: $ARCHS"
 
 for ARCH in ${ARCHS}; do
     if [ "${ARCH}" == "i386" ] || [ "${ARCH}" == "x86_64" ]; then
-        PLATFORM="iPhoneSimulator"
+        PLATFORM="iphonesimulator"
```
AntonioLangiu: For @bassosimone: this is not just a variable renaming, we need lower case values to pass it to xcrun.
The name with CamelCase was used to retrieve the SDK directly with the folder, but I think using xcrun is more robust!

build/ios/scripts/build_arch.sh
```diff
@@ -10,44 +10,66 @@ fi
...
+MIOS_VERSION="7.1"
+if [ $PLATFORM == "iphoneos" ]; then
+    MIOS_VERSION="-miphoneos-version-min=$MIOS_VERSION"
+else 
+    MIOS_VERSION="-mios-simulator-version-min=$MIOS_VERSION"
+fi
+
+HOST_FLAGS="-arch ${ARCH} ${MIOS_VERSION} -isysroot $(xcrun -sdk ${PLATFORM} --show-sdk-path)"
 
-DEVELOPER=$(xcode-select -print-path)
```
AntonioLangiu: We don't need this because xcrun finds the correct directory

build/ios/scripts/build_arch.sh
```diff
@@ -10,44 +10,66 @@ fi
...
 
-if [ "$PLATFORM" = "iPhoneOS" ]; then
-    EXTRA_CONFIG="--host=arm-apple-darwin14 --target=arm-apple-darwin14"
-fi
+OPT_FLAGS="-Os -g3"
+MAKE_JOBS=2
+
+ROOTDIR=$(cd $(dirname "$0") && pwd -P)
+SOURCEDIR=$(cd "${ROOTDIR}/../../../" && pwd -P)
+BUILDDIR=$(cd "${ROOTDIR}/../" && pwd -P)
```
bassosimone: This change is purely cosmetic compared to the original. Can you revert it?

bassosimone:
Hmm, you should start with rebasing this diff to master. There are things inside this diff that have already been committed to master.

AntonioLangiu:
Closing this because I refactor all this changes in another branch! #523

PR #516, opened 2016-04-30
Build dependency script improvements

AntonioLangiu:
I moved the commit related to build/dependency in this branch.

The explanation of the changes are in the commit messages.

Please check this branch locally before merging, because it can't be tested with travis, we are not building the dependencies


bassosimone:
- [ ] tested locally with `./build/dependency all` on osx
- [ ] tested locally with `./build/dependency all` on linux

PR #518, opened 2016-05-01
Repair build of iOS frameworks

bassosimone:
This is the smallest set of changes from #514 that allow to build all the dependencies, including libressl, for iOS and the iOS emulator. Detail of changes:

1. need to define EXTRA_CONFIG also when using the emulator to ensure dependencies' `./configure` knows we're cross compiling

2. forward `pkg_foo` variables for `./build/dependency`

3. call `./build/dependency` to ensure we have dependencies at the beginning of each build loop

4. tell `./configure` where dependencies are

5. make sure we create all frameworks

6. make sure we can cross compile GeoIP by setting variables that affect its `./configure`


AntonioLangiu:
I'm building this locally!

AntonioLangiu:
I think it is ok. Merging!

PR #519, opened 2016-05-04
Pass settings through levels

AntonioLangiu:
The idea of this PR is to pass the settings variable from the connection functions to the dns resolution function, so we are able to set options for the different levels with just one dictionary.

### DIFF
--- a/test/net/connect.cpp
+++ b/test/net/connect.cpp
@@ -175,8 +175,7 @@ TEST_CASE("connect_first_of works with empty vector") {
                     REQUIRE(errors.size() == 0);
                     REQUIRE(bev == nullptr);
                     break_loop();
-                },
-                3.14);
+                });
     });
 }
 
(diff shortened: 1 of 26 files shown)


test/net/connect.cpp
```diff
@@ -175,8 +175,7 @@ TEST_CASE("connect_first_of works with empty vector") {
                     REQUIRE(errors.size() == 0);
                     REQUIRE(bev == nullptr);
                     break_loop();
-                },
-                3.14);
+                });
```
bassosimone: Here you remove timeout, right? If so, why?

bassosimone:
Uhm, this is probably not ready for review yet. In any case, it looks a promising piece of work that simplifies matters and rationalizes the library. I like what I see.

test/net/connect.cpp
AntonioLangiu: No reason, sorry 👍  Good catch :)

AntonioLangiu:
I force pushed on this branch to rebase it on top of refactor/settings_tags

PR #523, opened 2016-05-05
Improvents for iOS build script

AntonioLangiu:
As in the description. This is the refactor of #514

### DIFF
--- a/build/ios/scripts/build_arch.sh
+++ b/build/ios/scripts/build_arch.sh
@@ -29,12 +29,6 @@ ROOTDIR=$(cd $(dirname $0) && pwd -P)
 SOURCEDIR="$ROOTDIR/../../../"
 BUILDDIR="$ROOTDIR/../"
 
-# We assume that the user has the iphoneos sdk installed
-AVAIL_SDKS=`xcodebuild -showsdks | grep "iphoneos"`
-FIRST_SDK=`echo "$AVAIL_SDKS" | head -n1`
-SDKVERSION=`echo "$FIRST_SDK" | cut -d\  -f2`
-echo "Using this SDK version: $SDKVERSION"
-
 DEVELOPER=$(xcode-select -print-path)
 
 export PATH="${DEVELOPER}/Toolchains/XcodeDefault.xctoolchain/usr/bin/:${DEVELOPER}/Platforms/${PLATFORM}.platform/Developer/usr/bin/:${DEVELOPER}/Toolchains/XcodeDefault.xctoolchain/usr/bin:${DEVELOPER}/usr/bin:${PATH}"
--- a/build/ios/scripts/build_arch.sh
+++ b/build/ios/scripts/build_arch.sh
@@ -33,12 +33,12 @@ DEVELOPER=$(xcode-select -print-path)
 
 export PATH="${DEVELOPER}/Toolchains/XcodeDefault.xctoolchain/usr/bin/:${DEVELOPER}/Platforms/${PLATFORM}.platform/Developer/usr/bin/:${DEVELOPER}/Toolchains/XcodeDefault.xctoolchain/usr/bin:${DEVELOPER}/usr/bin:${PATH}"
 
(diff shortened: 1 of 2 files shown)


build/ios/scripts/build_arch.sh
```diff
@@ -3,48 +3,41 @@ set -e
...
+DESTDIR="$ROOTDIR/../build/${PLATFORM}/${ARCH}/"
 
-export CC="/usr/bin/gcc -arch ${ARCH} -miphoneos-version-min=${MINIOSVERSION}"
-export CXX="/usr/bin/g++ -arch ${ARCH} -miphoneos-version-min=${MINIOSVERSION}"
-export CPPFLAGS="-isysroot ${DEVELOPER}/Platforms/${PLATFORM}.platform/Developer/SDKs/${PLATFORM}${SDKVERSION}.sdk"
-export CFLAGS="-isysroot ${DEVELOPER}/Platforms/${PLATFORM}.platform/Developer/SDKs/${PLATFORM}${SDKVERSION}.sdk"
-export CXXFLAGS="-isysroot ${DEVELOPER}/Platforms/${PLATFORM}.platform/Developer/SDKs/${PLATFORM}${SDKVERSION}.sdk"
-export LDFLAGS="-isysroot ${DEVELOPER}/Platforms/${PLATFORM}.platform/Developer/SDKs/${PLATFORM}${SDKVERSION}.sdk"
+export CC="$(xcrun -find -sdk ${PLATFORM} cc)"
+export CXX="$(xcrun -find -sdk ${PLATFORM} g++)"
```
bassosimone: Why was `arch ${ARCH}` moved from the two variables above into flags? Similarly, why not replacing the version min command line setting with the newly defined `${MINIOSVERSION}`?

build/ios/scripts/build_arch.sh
```diff
@@ -3,48 +3,41 @@ set -e
...
+if [ "$PLATFORM" = "iphoneos" ]; then
+    EXTRA_CONFIG="--host=arm-apple-darwin14 --target=arm-apple-darwin14 --enable-static --disable-shared"
+    MINVERSION="-miphoneos-version-min=$MINIOSVERSION"
 elif [ "$ARCH" = "i386" ]; then
-    EXTRA_CONFIG="--host=i386-apple-darwin14 --target=i386-apple-darwin14"
+    EXTRA_CONFIG="--host=i386-apple-darwin14 --target=i386-apple-darwin14 --enable-static --disable-shared"
+    MINVERSION="-mios-simulator-version-min=$MINIOSVERSION"
 elif [ "$ARCH" = "x86_64" ]; then
-    EXTRA_CONFIG="--host=x86_64-apple-darwin14 --target=x86_64-apple-darwin14"
+    EXTRA_CONFIG="--host=x86_64-apple-darwin14 --target=x86_64-apple-darwin14 --enable-static --disable-shared"
```
bassosimone: AFAIK `./configure` generated by autoconf enables both shared and static. So, what is the purpose of explicit `--enable-static`? Is this needed for specific libraries to build? Otherwise, I'll just `--disable-shared`.

build/ios/scripts/build_arch.sh
```diff
@@ -56,15 +49,15 @@ export pkg_make_flags=-j4
     test -x ./configure || ./autogen.sh
     ./configure -q --disable-shared \
```
bassosimone: You already pass `--disable-shared` as part of `$EXTRA_CONFIG` below. Thus, I'd remove it from here.

build/ios/scripts/build_arch.sh
AntonioLangiu: Because I'm using the CC and CXX variables to define the tools, CCFLAGS and CXXFLAGS to define the flags for the tools

bassosimone: Okay, I see your point towards consistency with GNU's standard. My concern was that `gcc -arch ${ARCH}` could be considered "the compiler". But, given that your version builds as well, I am going to accept this hunk, since it's less surprising than before. Thanks.

bassosimone:
It works! Thank you! 👍

Issue #522, opened 2016-05-04
Use well defined strings to define options


AntonioLangiu:
I'm trying to implement the feature to pass the settings dictionary from one level to the other of mk (expecially from the hightest levels to the lowers). For example, it will empower us to set the dns resolver and all the options of the dns when executing an http request.

To do this wee need to have specific options for the different levels, and this options should not collide.

For example, now I have the problem that, when I'm passing the timeout value to the connection, it will be taken also from the dns, because both of them search in the settings dictionary a string called "timeout".

A possibile solution discussed with @bassosimone today, was to use strings like `dns/timeout` and `net/timeout`.

I need this modifications to accomplish [this goal](https://github.com/measurement-kit/measurement-kit/issues/331)

PR #524, opened 2016-05-05
Use unique tags as key in the settings dictionary

AntonioLangiu:
See issue #522. 

To finish this PR I need to change the tag in the documentation.

### DIFF
--- a/src/net/connect.cpp
+++ b/src/net/connect.cpp
@@ -32,7 +32,7 @@ namespace mk {
 namespace net {
 
 void connect_first_of(std::vector<std::string> addresses, int port,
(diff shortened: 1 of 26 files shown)


src/net/connect.cpp
```diff
@@ -110,11 +111,11 @@ void resolve_hostname(
             }
             cb(*result);
         }, {}, poller);
```
bassosimone: Shouldn't this `{}` become `settings`? Or is this something to be done in another pull request?

AntonioLangiu: yes, It should be done in another PR. 

In this PR I just changed the tags.

bassosimone: ack

PR #538, opened 2016-05-08
Teach tcp_connect_test to pass the settings options to the dns level

AntonioLangiu:
Teach tcp_connect_test to pass the settings options to the dns level
and teach the connect function in src/ooni/tcp_test_impl to deal
with an invalid transport!


bassosimone:
Looks good to me! Adding to the list of things I should merge!

PR #539, opened 2016-05-08
Teach dns_injection_test to pass the options to the dns level

AntonioLangiu:
The changes are: 

- Teach dns_injection_test to pass the options to the dns level 
- Speed up the unit test phase setting a lower timout value (because dns_injection_test was one of the longest unit test) 

This PR is based on top of #538


bassosimone:
Excellent, thank you!

PR #540, opened 2016-05-08
Templatize poller and remove libs.hpp

bassosimone:
This diff should probably increase our coverage a little bit.

Unfortunately it was not possible to avoid rewriting the unit test. I suggest to read the new version of the file and to use the coverage to understand whether we're correctly testing the code.


AntonioLangiu:
It looks good to me. 👍 Merging!

PR #542, opened 2016-05-08
Teach http_invalid_request line to pass the options

AntonioLangiu:
As in the title!


bassosimone:
Looks good to me! Correct me if I'm wrong: the `-w` feature of `git diff

AntonioLangiu:
Yes, you are right. The changes are:
- add a check for an invalid tranport 
- add 2 tests to check if it works
- reformat the code using clang-format

bassosimone:
Okay, thanks for confirming that! Going to merge!

PR #543, opened 2016-05-08
Reactor: add background reactor

bassosimone:
This new functionality should make the code in Async unnecessary
because it provides the same features with less code and complexity.

Here I have just added a test showing how simple it is to run
things in the background. More work needed to replace Async.

- - -

This pull request should be tested and reviewed carefully, because it
contains possible breaking changes.

This pull request supersedes #390, which however I would like to
keep open until we've made progresses with removing Async.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,7 @@ test/common/error_or
 test/common/funcs
 test/common/log
 test/common/poller
+test/common/reactor
 test/common/settings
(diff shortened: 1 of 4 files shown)


bassosimone:
We have ported and improved and merged into master (with #604) the goodies contained into this pull request. Thus, we can now close this PR.

PR #550, opened 2016-05-09
async: radically simplify

bassosimone:
This is similar to work done in #543 and #390, except that here the
changes are small and incremental with respect to master.

As in #390, this diff changes behavior, because we always have a running
thread, which is compatible with a Tor based loop.

Unlike #543, here the patch only operates in src/common/async.cpp.

In the long run, I'd like to end up in a situation where we have #543,
but I recognize that this diff causes less changes in the tree.

Also, #543 is more fancy with respect to threads. Here I have stayed
dumb on purpose (see the 1.0 second delay when running new tests), not
to complicate to much the diff and hopefully expedite review.


src/common/async.cpp
```diff
@@ -46,6 +49,13 @@ void Async::join() {
     }
 }
 
+Async::~Async() {
+    // WARNING: This MUST be here to make sure we break the loop before we
+    // stop the thread. Not doing that leads to undefined behavior.
+    break_loop();
+    join();
```
AntonioLangiu: 1. Should you check if it's running before calling break_loop()? 
2. Why do you need to call join here, and witch thread are you blocking?

bassosimone: > 1. Should you check if it's running before calling break_loop()? 

No. Break_loop() has no side effect if the loop is not running. (There was on purpose a regress test that checks that nothing bad happens if we're not running the loop and the thread.)

> 2. Why do you need to call join here, and which thread are you blocking?

I need to call `join()` to join the thread possibly started by `run_test()`. (Note that if the thread has not been started `join()` is a no-op.) More specifically, the reason by which I must perform both operations here is because such thread uses the `this` of the current object (you can probably imagine me mumbling something about using `Var<T>` head instead, which incidentally would be memory safe by construction). As such, `this` is shared and we must make sure we stop also the background thread when the `Async` object goes out of scope. (Note that the default destructor of the `std::thread` would have joined in any case, but given that in here there is potential of doing mistakes, I've decided to join the thread explicitly.)

test/common/async.cpp
```diff
@@ -74,3 +74,20 @@ TEST_CASE("The async engine works as expected") {
...
+    while (!async.empty()) {
+        sleep(1);
+    }
+    async.break_loop();
+    async.join();
+}
+
+TEST_CASE("Nothing strange happens if no thread is bound to Async") {
+    Async async;
+    REQUIRE(async.empty()); // Mainly to avoid unused variable warning
```
bassosimone: @AntonioLangiu This is the test I was mentioning in my earlier comment

src/common/async.cpp
```diff
@@ -7,138 +7,58 @@
...
-                    });
+    active += 1;
+    debug("async: scheduling %llu", test->identifier());
+    reactor->call_later(1.0, [=]() {
+        debug("async: starting %llu", test->identifier());
+        test->begin([=]() {
+            debug("async: ending %llu", test->identifier());
+            test->end([=]() {
+                debug("async: cleaning-up %llu", test->identifier());
+                reactor->call_soon([=]() {
```
bassosimone: Deferring calling the final callback one more cycle may not be needed in all cases, but it's a safety measure to ensure that, even if other parts of the code do not defer cleanup correctly, still the `test` object would not be below the current point in the stack when the final callback is called. That is, I coded it like this for extra robustness.

bassosimone: I should probably add a comment to explain that, because otherwise it's not super clear. @AntonioLangiu what do you think?

AntonioLangiu: Yes, I think you should! So it will be clear for every one reading the code!

bassosimone: ack, will do this later!

src/common/async.cpp
AntonioLangiu: Ok thank you! I was wondering if the correct way of calling join was on the thread object, something like:

    if (running)
        thread.join();

src/common/async.cpp
bassosimone: @AntonioLangiu: added comment!

AntonioLangiu:
Ok, thank you! I'm updating the branch and then if everything is ok I'm going to merge it!

PR #558, opened 2016-05-10
Teach http_invalid_request_line_test to deal with invalid backend

AntonioLangiu:
As in the title!


src/ooni/http_invalid_request_line_impl.hpp
```diff
@@ -76,7 +76,13 @@ class HTTPInvalidRequestLineImpl : public TCPTestImpl {
     void main(Settings options, std::function<void(json)> &&cb) {
         entry["tampering"] = nullptr;
 
-        http::Url backend_url = mk::http::parse_url(options["backend"]);
+        ErrorOr<http::Url> backend_url = mk::http::parse_url_noexcept(options["backend"]);
+
+        if (!backend_url) {
+            std::cout << "hey" << std::endl;
```
bassosimone: This debugging information should not be merged into master. Could you remove it?

test/ooni/http_invalid_request_line_test.cpp
```diff
@@ -76,3 +76,16 @@ TEST_CASE("Make sure that it can pass options to the other levels") {
...
     for (auto &s : *logs) std::cout << s << "\n";
 }
+
+TEST_CASE("Make sure that the test can deal with an invalid backend") {
+    Var<std::list<std::string>> logs(new std::list<std::string>);
+    bool done = false;
+    ooni::HttpInvalidRequestLineTest()
+        .set_backend("nexacenter.org")
+        .on_log([=](const char *s) { logs->push_back(s); })
+        .run([&done]() { done = true; });
```
bassosimone: /me puzzled: Shouldn't `run()` be blocking and only terminate at the end of the test?

AntonioLangiu: Yes, I used the implementation of the test used [here](https://github.com/measurement-kit/measurement-kit/blob/fix/http_invalid_request_line_backend/test/ooni/http_invalid_request_line_test.cpp#L36). The blocking and non blocking version are actually the same. I'm going to use the blocking version to more concise!

PR #572, opened 2016-05-11
Implement neubot's dash test

helloimcarmine:
I've added an early version of a DASH test developed with mk.


bassosimone:
@helloimcarmine could you update this branch to master? Thanks!

example/neubot/client_dash.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <string>
+#include <unistd.h>
+
+using namespace mk;
+using namespace mk::neubot::negotiate;
+using namespace mk::net;
+using namespace mk::http;
+
+static const char *kv_usage =
+    "usage: ./example/http/client_dash [-vn] [-a address]\n";
```
bassosimone: Should be `./example/neubot/dash` (which in 0.3.0 would perhaps become the `neubot` command and the `dash` subcommand). Could you rename the file accordingly?

example/neubot/client_dash.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+int main(int argc, char **argv) {
+    Settings settings;
+    settings["http/url"] = "http://127.0.0.1";
```
bassosimone: It would be optimal if this key was called `url` and the test machinery provides to set up `http/url` according to the value of `url`. It would also be optimal if, when the `url` is empty, it is looked up using mlabns.

If the latter is too complex to implement before the end of the internship, never mind. But, to start off, I'd like to see the address of a real mlab server to be used as default (e.g. `neubot.mlab.mlab1.mil01.measurement-lab.org`).

example/neubot/client_dash.cpp
```diff
@@ -0,0 +1,54 @@
...
+            break;
+        default:
+            std::cout << kv_usage;
+            exit(1);
+        }
+    }
+
+    loop_with_initial_event([=]() {
+        run_negotiation(settings, [=](Error) {
+
```
bassosimone: Printing the value of the error at the end of the test, perhaps with level `debug` could be useful in some cases.

include/measurement_kit/neubot/dash.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <functional>
+#include <measurement_kit/common.hpp>
+#include <measurement_kit/ext.hpp>
+
+namespace mk {
+namespace neubot {
+namespace dash {
+
+// Run a dash test
+void run(Settings settings, Callback<Error,  Var<nlohmann::json>> cb,
```
bassosimone: Am I right in assuming that this step cannot run without knowing the server address? In such case, I'd refactor this function to receive a string first argument and move `settings` with optional arguments before `reactor` and initialized by default using `{}`. Also, given that `auth` is optional, I'd probably pass it using the settings.

include/measurement_kit/neubot/dash.hpp
```diff
@@ -0,0 +1,26 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+#ifndef MEASUREMENT_KIT_NEUBOT_DASH_HPP
+#define MEASUREMENT_KIT_NEUBOT_DASH_HPP
+
+#include <functional>
```
bassosimone: Strictly speaking, header not needed, because already pulled in by `common.hpp`.

include/measurement_kit/neubot/negotiate.hpp
```diff
@@ -0,0 +1,22 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+#ifndef MEASUREMENT_KIT_NEUBOT_NEGOTIATE_HPP
+#define MEASUREMENT_KIT_NEUBOT_NEGOTIATE_HPP
+
+#include <functional>
```
bassosimone: Again, header not needed.

include/measurement_kit/neubot/dash.hpp
bassosimone: Also, rather than exposing `nlohmann::json`, could you see whether the code would work exposing `report::Entry`? If it does not work, let me know why, and I'll see if I could quickly patch Entry up.

include/measurement_kit/neubot/negotiate.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <measurement_kit/common.hpp>
+
+namespace mk {
+namespace neubot {
+namespace negotiate {
+
+void run_negotiation(Settings settings, Callback<Error> cb,
```
bassosimone: I'd rename this function run, because I see lots of sense in it being called as `negotiate::run()`. I'd also make `settings` optional, move it before `reactor` and initialize it with `{}`.

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+                        s_k *= relerr;
+                        if (s_k < 0) {
+                            s_k = 100;
+                        }
+                    }
+
+                    loop_request(transport, s_k, cb, auth, measurements,
+                                 reactor, logger, iteration + 1);
+
+                    return;
```
bassosimone: `return` not needed here: you're at the end of a lambda.

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+                            {"iteration", iteration}, // OK
+                            //{"platform", sys.platform}
+                            {"rate", rate_kbit}, // OK
+                            //{"real_address", self.parent.real_address}
+                            //{"received", received}
+                            //{"remote_address", stream.peername[0]}
+                            //{"request_ticks", self.saved_ticks}
+                            //{"timestamp", mk::time_now}
+                            //{"uuid", self.conf.get("uuid")}
+                            //{"version", utils_version.NUMERIC_VERSION}
```
bassosimone: This would be `MEASUREMENT_KIT_VERSION` defined in `measurement_kit/common.hpp`.

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+                        cb(error, nullptr);
+                        return;
+                    }
+
+                    double new_times = mk::time_now();
+                    float length = res->body.length();
+                    double time_elapsed = new_times - saved_times;
+
+                    if (time_elapsed < 0) {
+                        std::cout << "Error: " << (int)error;
```
bassosimone: Please, use `logger->warn()` to log about errors

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+                        cb(error, nullptr);
```
bassosimone: Hmm, I think here you should create a new error (see `error.hpp` for how to create errors). Reading the code above, it's otherwise clear that here `err` is `NoError()` which is not what you want.

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+                std::cout << "Error: " << (int)error;
+                cb(error, nullptr);
+                return;
+            }
+
+            request_recv_response(
+                transport,
+                [=](Error error, Var<Response> res) {
+                    if (error) {
+                        std::cout << "Error: " << (int)error;
```
bassosimone: Same as above: do not use `std::cout` use `logger->warn()` and similar functions (see `common/logger.hpp`)

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+    http::request_send(
+        transport, settings,
+        {
+            {"Authorization", auth},
+        },
+        "",
+        [=](Error error) {
+            if (error) {
+                std::cout << "Error: " << (int)error;
```
bassosimone: As said below

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+namespace dash {
+
+static inline void loop_request(Var<Transport> transport, int speed_kbit,
+                                Callback<Error, Var<json>> cb, std::string auth,
+                                Var<json> measurements, Var<Reactor> reactor,
+                                Var<Logger> logger, int iteration = 1) {
+
+    int rate_index = 0;
+    Settings settings;
+    int DASH_RATES[] = {100,  150,  200,  250,  300,   400,  500,
```
bassosimone: This should probably be `static const int`

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+                        700,  900,  1200, 1500, 2000,  2500, 3000,
+                        4000, 5000, 6000, 7000, 10000, 20000};
+    std::string url = "http://127.0.0.1/dash/download/";
```
bassosimone: Shouldn't the url be set by the caller?!

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+    if (iteration > DASH_MAX_ITERATION) {
+        transport->close([=]() { cb(NoError(), measurements); });
+        return;
+    };
```
bassosimone: Semicolon not needed here

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+    // Bisect
+    while (DASH_RATES[rate_index] < speed_kbit) {
+        rate_index++;
```
bassosimone: I think here you should check not to go beyond the vector's boundaries, shouldn't you?

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+    rate_index -= 1;
+    if (rate_index < 0)
+        rate_index = 0;
```
bassosimone: Please, always use `{` and `}` also for one line statements.

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+                    if (time_elapsed > DASH_SECONDS) {
+                        float relerr = 1 - (time_elapsed / DASH_SECONDS);
+                        s_k *= relerr;
+                        if (s_k < 0) {
+                            s_k = 100;
+                        }
+                    }
+
+                    loop_request(transport, s_k, cb, auth, measurements,
```
bassosimone: I know I said differently in the past, but after #644 I prefer to avoid explicit recursion. could you wrap this invocation of `loop_request()` inside of `poller->call_soon([=]() { ... });`?

src/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+                    [=](Error error, Var<Transport> transport) {
+
+                        if (error) {
+                            std::cout << "Error: " << (int)error;
+                            cb(error, nullptr);
+                            return;
+                        }
+
+                        Var<json> measurements(new json);
+                        *measurements = {};
```
bassosimone: I am not sure this initialization step is needed, is it?

bassosimone:
Thanks! I've added some comments. Could you please apply the related changes (and similar changes in spirit to other locations in the code that have similar issues but that I've not touched with my comments) and then ping me back? Thanks again!

src/neubot/negotiate_impl.hpp
```diff
@@ -127,16 +134,31 @@ static inline void loop_req_negotiate(Var<Transport> transport, Callback<Error>
...
+        mlabns::query(
+           tool, [&url](Error error, mlabns::Reply reply) {
+               if (error) {
+                   logger -> warn("Error: %d", (int) error);
+                   cb(error);
+                   return;
+               }
+               url = reply.url;
+           });
+        settings["http/url"] = url;
```
helloimcarmine: @bassosimone I have taken this part of code from the example of use of mlabns, but I don't understand why it doesn't work. It seems like it doesn't return an URL.

src/neubot/dash_impl.hpp
```diff
@@ -87,12 +87,12 @@ static inline void loop_request(Var<Transport> transport, int speed_kbit,
 
                     if (time_elapsed < 0) {
                         logger -> warn("Time elapsed can't be negative");
-                        cb(error, nullptr);
+                        cb(ValueError(), nullptr);
```
helloimcarmine: @bassosimone I thought that this is the most appropriate error to use. If you think that a new error would be more accurate tell me the name to use for it.

src/neubot/negotiate_impl.hpp
```diff
@@ -59,19 +61,21 @@ static inline void loop_req_negotiate(Var<Transport> transport, Callback<Error>
...
+    std::string body = value.dump();
 
     settings["http/path"] = "/negotiate/dash";
     settings["http/method"] = "POST";
 
+
     if (iteration > DASH_MAX_NEGOTIATION) {
         std::cout << "Too many negotiations\n";
-        transport->close([=]() { cb(NoError()); });
+        transport->close([=]() { cb(ValueError()); });
```
helloimcarmine: @bassosimone same as before!

src/neubot/negotiate_impl.hpp
bassosimone: Yes, because `mlabns::query()` is asynchronous. You don't have the URL until the delayed lambda passed to `mlabns::query()` is called.

src/neubot/dash_impl.hpp
bassosimone: I think it would be more appropriate to define a new error in `mlabns.hpp` header that indicates what is happening here and then use that error.

src/neubot/negotiate_impl.hpp
bassosimone: Again, here I'd define and use an error like `TooManyNegotiationsError`

bassosimone:
@helloimcarmine can you merge this with master?

test/neubot/negotiate.cpp
```diff
@@ -53,7 +53,7 @@ static void receive_invalid_status_code(Var<net::Transport>, Settings, Headers,
 TEST_CASE("Make sure that an error is passed to callback if the response "
           "status is not 200") {
     loop_negotiate<receive_invalid_status_code>( nullptr,
-        [](Error error) { REQUIRE(error == http::HttpRequestFailedError()); }, {},
+        [](Error error) { REQUIRE(http::HttpRequestFailedError()); }, {},
```
bassosimone: What is the meaning of this assertion?

bassosimone:
Closed in favor of #769

PR #573, opened 2016-05-11
Autogenerate version.hpp.in

bassosimone:


bassosimone:
Valgrind failed on travis. The reason is that [connect failed in a test, an exception was raised, thus leading to not all the memory being freed](https://travis-ci.org/measurement-kit/measurement-kit/jobs/129471638#L2550). Restarting the build.

PR #580, opened 2016-05-12
Faster iOS build by using precompiled dependencies

bassosimone:
1) add set of scripts to precompile dependencies in advance and publish them

2) modify build process to pull precompiled dependencies

3) build/ios/build => build/ios/tmp because imho build/ios/build was confusing


bassosimone:
cc: @AntonioLangiu @lorenzoPrimi

bassosimone:
This pull request cannot be merged because of a bug. We must at least remove the `.la` files inside `tmp` otherwise the binary bundle downloaded to another computer would not work because the `.la` files hardcode the path of libraries to a nonexistent location. I've manually fixed the binaries that are currently published online but it would be better to automate the process.

bassosimone:
This should be ready for review again, as the above mentioned problem should be fixed now.

AntonioLangiu:
I'm testing this locally

AntonioLangiu:
Ok, I tested it locally and it works. Good work @bassosimone, we need now just 5 minutes to build it!

PR #581, opened 2016-05-12
Teach mlabns to honour net/ca_bundle_path

bassosimone:
While there simplify implementation to receive input from a Settings
object rather than from a special purpose object.

This diff is best viewed appending `?w=1` to github diff URL because I
run `clang-format` over the file and formatting changed.

(That is, using [this URL](https://github.com/measurement-kit/measurement-kit/pull/581/files?w=1).)

This diff assumes that @AntonioLangiu has already implemented the
default-ca-path diff, therefore it does not bother with setting default
values for the CA path in the example.


AntonioLangiu:
It looks good to me, I'm updating it to master and then merging!

AntonioLangiu:
Now, merging this!

bassosimone:
Awesome, thanks for expediting its review!

PR #582, opened 2016-05-12
All tests now use the same DSL

bassosimone:
Seems better than the previous shell script based approach, isn't it, @AntonioLangiu?


src/ooni/dns_injection_impl.hpp
```diff
@@ -37,7 +37,7 @@ class DNSInjectionImpl : public DNSTestImpl {
     void main(std::string input, Settings options,
               std::function<void(json)> &&cb) {
         entry["injected"] = nullptr;
-        query("A", "IN", input, options["dns/nameserver"],
+        query("A", "IN", input, options["backend"],
```
bassosimone: Here I changed the name used for the option because what we are actually setting here is the backend for the test (i.e., "dns/nameserver" is an lower-level option).

test/ooni/dns_injection.cpp
```diff
@@ -14,7 +14,7 @@ using namespace mk::ooni;
 TEST_CASE(
     "The DNS Injection test should run with an input file of DNS hostnames") {
     Settings options;
-    options["dns/nameserver"] = "8.8.8.1:53";
+    options["backend"] = "8.8.8.1:53";
```
bassosimone: This is because I changed the option passed to DnsInjection test

test/ooni/dns_injection.cpp
```diff
@@ -25,7 +25,7 @@ TEST_CASE(
 TEST_CASE("The DNS Injection test should throw an exception if an invalid file "
           "path is given") {
     Settings options;
-    options["dns/nameserver"] = "8.8.8.1:53";
+    options["backend"] = "8.8.8.1:53";
```
bassosimone: Ditto

test/ooni/dns_injection.cpp
```diff
@@ -34,7 +34,7 @@ TEST_CASE("The DNS Injection test should throw an exception if an invalid file "
 TEST_CASE("The DNS Injection test should throw an exception if no file path is "
           "given") {
     Settings options;
-    options["dns/nameserver"] = "8.8.8.1:53";
+    options["backend"] = "8.8.8.1:53";
```
bassosimone: Ditto

AntonioLangiu:
Yes, I think it is really better to use this approach. I'm going to open two issues to remeber we need a template to set the options, and we need to fix the documentation about the DSL, specifying exactly whitch parameters are accepted from the tests!

bassosimone:
> specifying exactly which parameters are accepted from the tests!

Regarding this, since I anticipated that passing around `Settings` would create some debugging issues, I recently wrote and committed [code to dump settings](https://github.com/measurement-kit/measurement-kit/blob/master/src/common/utils.cpp#L279), `#KnowIt`.

AntonioLangiu:
Travis is actually failing because of [this](https://travis-ci.org/measurement-kit/measurement-kit/jobs/130195474#L1103) that should be refactored as: `set_options("backend", "antani")`
I'll do it!

AntonioLangiu:
I tested it locally and it works!

AntonioLangiu:
Updating and merging!

PR #590, opened 2016-05-14
Teach configure to search for the default ca-bundle

AntonioLangiu:
A little bit inspired from cURL this PR add to configure the --with-ca-bundle option. If it is not given it searches for a default installation of ca-bundle in this places:

- /etc/ssl/certs/ca-certificates.crt
- /etc/pki/tls/certs/ca-bundle.crt
- /usr/share/ssl/certs/ca-bundle.crt
- /usr/local/share/certs/ca-root.crt
- /etc/ssl/cert.pem


acinclude.m4
```diff
@@ -212,6 +212,58 @@ CXXFLAGS="$measurement_kit_saved_cxxflags $measurement_kit_cxx_stdlib_flags"
...
+    fi
+  fi
+
+  if test "x$ca" != "xno"; then
+    MK_CA_BUNDLE="$ca"
+    AC_DEFINE_UNQUOTED(MK_CA_BUNDLE, "$ca", [Location of default ca bundle])
+    AC_SUBST(MK_CA_BUNDLE)
+    AC_MSG_RESULT([$ca])
+  fi
+  if test "x$ca" = "xno"; then
```
bassosimone: wouldn't it make sense to replace the two lines above with just `else`?

bassosimone: never mind, I would take care of this when merging.

PR #593, opened 2016-05-14
Revert: autogeneration of version.hpp

AntonioLangiu:
We noticed that autogenerating the version file was more a problem that a feature!


include/measurement_kit/common/version.hpp
```diff
@@ -6,6 +6,6 @@
 #define MEASUREMENT_KIT_COMMON_VERSION_HPP
 
 // Note: we use semantic versioning (see: http://semver.org/)
-#define MEASUREMENT_KIT_VERSION "0.2.0-alpha.1"
+#define MEASUREMENT_KIT_VERSION "0.2.0-alpha"
```
bassosimone: This should not be part of the revert

AntonioLangiu: You are right, it was in the commit I reverted. I will add a new commit fixing it!

bassosimone:
Okay, now it looks good to me!

Issue #465, opened 2016-04-12
write some tests for mlabns


bassosimone:
After #463 is done, it should be nice to have some tests for mlabns. To start off, it would probably be good to have integration tests where we call directly the service.

I'd also like to have a test where we pass the body received from mlab-ns to a function and we obtain back the structure representing the response. This should allows us to assert that the library behaves even when invalid input (or unexpected!) is returned by the mlab-ns api.

bassosimone:
We already implemented some tests.

It would be nice to also cover the following cases:

- [ ] make sure that an error is passed to callback if input leads to an invalid query because of policy, metro, and/or address family
- [ ] make sure that an error is passed to callback if `http::request()` fails (you should probably mock the implementation using templates to do that)
- [ ] make sure that an error is passed to callback if the response status is not `200`
- [ ] make sure that an error is passed to callback if:
    - [ ] json parsing fails
    - [ ] a nonexistent json field is accessed

Regarding the two latter cases, it would also be nice to distinguish between them by defining two different errors, one for proper json parsing error and the other for missing field error.

AntonioLangiu:
Start working on this...

PR #596, opened 2016-05-14
Add mlabns unit test

AntonioLangiu:
In this unit-tests I assume that the ca-bundle has been found.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,7 @@ test/http/parse_url
 test/http/request
 test/http/request_serializer
 test/http/response_parser
+test/mlabns/mlabns
 test/net/buffer
 test/net/connect
 test/net/connect-many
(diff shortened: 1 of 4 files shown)


bassosimone:
Awesome!

It would be nice to also cover the following cases:

- [ ] make sure that an error is passed to callback if input leads to an invalid query because of policy, metro, and/or address family
- [ ] make sure that an error is passed to callback if `http::request()` fails (you should probably mock the implementation using templates to do that)
- [ ] make sure that an error is passed to callback if the response status is not `200`
- [ ] make sure that an error is passed to callback if:
    - [ ] json parsing fails
    - [ ] a nonexistent json field is accessed

bassosimone:
Closes #465.

bassosimone:
Uhm, I am going to merge this and going to move the TODO list above in #465.

PR #604, opened 2016-05-16
Runner: make sure thread is running before continuing

bassosimone:
This ports changes in #543 to the current implementation of Runner.

### DIFF
--- a/src/common/runner.cpp
+++ b/src/common/runner.cpp
@@ -30,7 +30,7 @@ void Runner::run_test(Var<NetTest> test, std::function<void(Var<NetTest>)> fn) {
     }
     active += 1;
     debug("runner: scheduling %llu", test->identifier());
-    reactor->call_later(1.0, [=]() {
+    reactor->call_soon([=]() {
         debug("runner: starting %llu", test->identifier());
         test->begin([=]() {
             debug("runner: ending %llu", test->identifier());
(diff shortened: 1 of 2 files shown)


AntonioLangiu:
Did you try to remove [this](https://github.com/measurement-kit/measurement-kit/pull/604/files#diff-763f0dd5e353e22c05dcd623e4551dedR34) timeout? I think that now we don't need it anymore.

src/common/runner.cpp
```diff
@@ -2,19 +2,32 @@
...
     if (!running) {
+        std::mutex mutex;
+        std::unique_lock<std::mutex> lock(mutex);
+        std::condition_variable condition;
         // WARNING: below we're passing `this` to the thread, which means that
         // the destructor MUST wait the thread. Otherwise, when the thread dies
         // many strange things could happen (I have seen SIGABRT).
-        thread = std::thread([this]() { reactor->loop(); });
+        debug("runner: starting reactor in background...");
+        thread = std::thread([&condition, &lock, &mutex, this]() {
```
AntonioLangiu: Why are you passing the lock and the mutex to this callback? The documentation says that the lock does not need to be held for notification.

bassosimone:
This branch has reset and pushed again to contain a better implementation written by @AntonioLangiu.

Issue #399, opened 2016-02-24
Roadmap for improving mlab-ns implementation


bassosimone:
- [x] use a real JSON library (perhaps [nlohmann/json](https://github.com/nlohmann/json)) rather than yaml-cpp to parse results (see #463)
- [x] implement https and use https to access mlabns
- [x] add some regress tests (#465)

AntonioLangiu:
This should be closed! @bassosimone

PR #606, opened 2016-05-17
Add some mlabns tests

### LINKED ISSUE #465: write some tests for mlabns

AntonioLangiu:
I added the tests listed in #465 
I applyed clang-format as last commit.

This closes #465 and #399.

### DIFF
--- a/src/mlabns/mlabns.cpp
+++ b/src/mlabns/mlabns.cpp
@@ -2,6 +2,7 @@
 // Measurement-kit is free software. See AUTHORS and LICENSE for more
 // information on the copying conditions.
 
+#include "src/mlabns/mlabns.hpp"
 #include "src/ext/json/src/json.hpp"
 #include <functional>
 #include <initializer_list>
@@ -13,7 +14,6 @@
 #include <string>
 #include <utility>
 #include <vector>
-#include "src/mlabns/mlabns.hpp"
 
 using json = nlohmann::json;
 
--- a/src/mlabns/mlabns.cpp
+++ b/src/mlabns/mlabns.cpp
@@ -3,19 +3,6 @@
 // information on the copying conditions.
 
 #include "src/mlabns/mlabns.hpp"
-#include "src/ext/json/src/json.hpp"
-#include <functional>
-#include <initializer_list>
-#include <map>
-#include <measurement_kit/common.hpp>
-#include <measurement_kit/http.hpp>
-#include <measurement_kit/mlabns.hpp>
-#include <regex>
-#include <string>
-#include <utility>
-#include <vector>
-
-using json = nlohmann::json;
 
 namespace mk {
 namespace mlabns {
--- a/src/mlabns/mlabns.cpp
+++ b/src/mlabns/mlabns.cpp
@@ -2,52 +2,12 @@
 // Measurement-kit is free software. See AUTHORS and LICENSE for more
 // information on the copying conditions.
 
-#include "src/mlabns/mlabns.hpp"
+#include "src/mlabns/mlabns_impl.hpp"
 
 namespace mk {
 namespace mlabns {
 
-ErrorOr<std::string> as_query(Settings &settings) {
-    std::string query;
-    std::string policy = settings.get<std::string>("mlabns/policy", "");
-    std::string metro = settings.get<std::string>("mlabns/metro", "");
-    std::string address_family =
-        settings.get<std::string>("mlabns/address_family", "");
-    if (policy == "" && metro == "" && address_family == "") {
-        return query;
-    }
-    if (policy != "") {
-        if (policy != "geo" && policy != "random" && policy != "metro" &&
-            policy != "country") {
-            return InvalidPolicyError();
-        }
-        if (query != "") {
-            query += "&";
-        }
-        query += "policy=" + policy;
-    }
-    if (metro != "") {
-        std::regex valid_metro("^[a-z]{3}$");
-        if (!std::regex_match(metro, valid_metro)) {
-            return InvalidMetroError();
-        }
-        if (query != "") {
-            query += "&";
-        }
-        query += "metro=" + metro;
-    }
-    if (address_family != "") {
-        if (address_family != "ipv4" && address_family != "ipv6") {
-            return InvalidAddressFamilyError();
-        }
-        if (query != "") {
-            query += "&";
-        }
-        query += "address_family=" + address_family;
-    }
-    query = "?" + query;
-    return query;
-}
+ErrorOr<std::string> as_query(Settings &settings);
 
 void query(std::string tool, Callback<Error, Reply> callback, Settings settings,
            Var<Reactor> reactor, Var<Logger> logger) {
--- a/src/mlabns/mlabns.cpp
+++ b/src/mlabns/mlabns.cpp
@@ -11,7 +11,7 @@ ErrorOr<std::string> as_query(Settings &settings);
 
 void query(std::string tool, Callback<Error, Reply> callback, Settings settings,
            Var<Reactor> reactor, Var<Logger> logger) {
-    query_debug(tool, callback, settings, reactor, logger);
+    query_impl(tool, callback, settings, reactor, logger);
 }
 
 } // namespace mlabns
--- a/src/mlabns/mlabns.cpp
+++ b/src/mlabns/mlabns.cpp
@@ -13,13 +13,14 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include "src/mlabns/mlabns.hpp"
 
 using json = nlohmann::json;
 
 namespace mk {
 namespace mlabns {
 
-static ErrorOr<std::string> as_query(Settings &settings) {
+ErrorOr<std::string> as_query(Settings &settings) {
     std::string query;
     std::string policy = settings.get<std::string>("mlabns/policy", "");
     std::string metro = settings.get<std::string>("mlabns/metro", "");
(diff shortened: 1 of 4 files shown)


src/mlabns/mlabns.cpp
```diff
@@ -2,24 +2,12 @@
...
-#include <vector>
-
-using json = nlohmann::json;
+#include "src/mlabns/mlabns.hpp"
 
 namespace mk {
 namespace mlabns {
 
-static ErrorOr<std::string> as_query(Settings &settings) {
+ErrorOr<std::string> as_query(Settings &settings) {
```
bassosimone: I think that the whole content of this file should become the private header and that the `.cpp` file should not do much more than including the private header. You should see how these things are sketched out in NDT, where typically the private header with templates is named `$name_impl.hpp`.

Hence, I'd move the above function into the header, as a `static inline` function, under the argument that to ease code review it's better to have all relevant code into the same place (as opposed to having part in `.cpp` and part into the internal header file).

src/mlabns/mlabns.cpp
```diff
@@ -63,54 +51,7 @@ static ErrorOr<std::string> as_query(Settings &settings) {
...
-                          return;
-                      } catch (std::out_of_range &) {
-                          callback(JsonKeyError(), Reply());
-                          return;
-                      }
-                      logger->info("mlabns says to use %s", reply.fqdn.c_str());
-                      callback(NoError(), reply);
-                  },
-                  {}, "", settings, logger, reactor);
+    query_debug(tool, callback, settings, reactor, logger);
```
bassosimone: I'd rename `query_debug` as `query_impl`, because that template is the *actual* implementation of the function, and instead I think that the name `query_debug` is a bit misleading. (This is how things are done in the NDT branch.)

src/mlabns/mlabns.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <utility>
+
+using json = nlohmann::json;
+
+namespace mk {
+namespace mlabns {
+
+ErrorOr<std::string> as_query(Settings &settings);
+
+template <decltype(http::get) get = http::get>
```
bassosimone: Here you should the recently added use `MK_MOCK_NAMESPACE` macro.

test/mlabns/mlabns.cpp
```diff
@@ -35,12 +38,198 @@ TEST_CASE("Query can pass the settings to the dns level") {
...
+                      [](Error error, mlabns::Reply) {
+                          REQUIRE(error);
+                          break_loop();
+                      },
+                      settings);
+    });
+}
+
+TEST_CASE(
+    "Make sure that an error is passed to callback with invalid query 1") {
```
bassosimone: I'd use more descriptive test cases descriptions here

test/mlabns/mlabns.cpp
```diff
@@ -35,12 +38,198 @@ TEST_CASE("Query can pass the settings to the dns level") {
...
+                      settings);
+    });
+}
+
+TEST_CASE(
+    "Make sure that an error is passed to callback with invalid query 2") {
+    Settings settings;
+    settings["mlabns/address_family"] = "ipv4";
+    settings["mlabns/metro"] = "trn";
+    settings["mlabns/policy"] = "antani"; // Invalid
```
bassosimone: :-D

test/mlabns/mlabns.cpp
```diff
@@ -35,12 +38,198 @@ TEST_CASE("Query can pass the settings to the dns level") {
...
+                          break_loop();
+                      },
+                      settings);
+    });
+}
+
+static void get_debug_error(std::string, http::RequestCallback cb,
+                            http::Headers, std::string, Settings, Var<Logger>,
+                            Var<Reactor>) {
+    cb(GenericError(), http::Response());
```
bassosimone: Rather than `GenericError` you should use for regress tests of which you mock failure the newly added `MockedError`.

test/mlabns/mlabns.cpp
```diff
@@ -35,12 +38,198 @@ TEST_CASE("Query can pass the settings to the dns level") {
...
+    Settings settings;
+    settings["mlabns/address_family"] = "ipv4";
+    settings["mlabns/metro"] = "trn";
+    settings["mlabns/policy"] = "random";
+    std::string tool = "neubot";
+
+    loop_with_initial_event([=]() {
+        mlabns::query_debug<get_debug_error>(tool,
+                                             [](Error error, mlabns::Reply) {
+                                                 REQUIRE(error);
```
bassosimone: Here you should probably check whether the error is what you generated.

bassosimone:
Thanks! That's a very good and complete coverage testing of the mlabns functionality. I'd be very happy to merge this once the requested refactorings have been implemented.

src/mlabns/mlabns.hpp
AntonioLangiu: Thanks, good feature!

test/mlabns/mlabns.cpp
AntonioLangiu: I think that we should just check that the error passed here is the same as the error receved [here](https://github.com/measurement-kit/measurement-kit/pull/606/files#diff-ec81c929655972c47f1ecc7393d23ee5R138).

test/mlabns/mlabns.cpp
AntonioLangiu: Yes, thanks 👍

test/mlabns/mlabns.cpp
bassosimone: My comment is that, when you need to mock the failure of an API for testing purposes, you should not emit `GenericError` because that kind of error is also emitted by MK in many other cases, rather you should emit `MockedError` which shall only be used for indicating that the failure of an API is currently being mocked.

src/mlabns/mlabns_impl.hpp
```diff
@@ -0,0 +1,114 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
```
bassosimone: I don't understand why this file is not reported as being a copy with small modifications of the original `.cpp` file. Since the changes between the two should be very small.

AntonioLangiu: maybe because I used git mv?

bassosimone: I don't know. Never mind, I could always do comparison locally.

src/mlabns/mlabns_impl.hpp
```diff
@@ -0,0 +1,114 @@
...
+            query += "&";
+        }
+        query += "address_family=" + address_family;
+    }
+    query = "?" + query;
+    return query;
+}
+
+template <MK_MOCK_NAMESPACE(http, get)>
+inline void query_impl(std::string tool, Callback<Error, Reply> callback,
```
bassosimone: AFAIK, there's no need to declare `inline` template functions.

src/mlabns/mlabns_impl.hpp
```diff
@@ -0,0 +1,114 @@
...
+                return;
+            }
+            if (response.status_code != 200) {
+                callback(UnexpectedHttpStatusCodeError(), Reply());
+                return;
+            }
+            Reply reply;
+            try {
+                auto node = json::parse(response.body);
+                reply.city = node.at("city");
```
bassosimone: I am *not* happy with this diff. You *should not* have run clang-format in this case. This complicated the diff and forced me to do comparison by hand, locally, between two branches. This led me to discover that, in addition to reformatting everything, you also applied the following changes:

```diff
-                          reply.city = node["city"];
-                          reply.url = node["url"];
-                          for (auto ip2 : node["ip"]) {
+                reply.city = node.at("city");
+                reply.url = node.at("url");
+                for (auto ip2 : node.at("ip")) {
```

Why did you did that? Why did you hide that into a reformatted diff? Why did you not document that? I understand that you probably did this for what is probably a good reason, but I don't think that's the correct way of proceeding.

By producing diff complex to easily review without a good reason, you add burden on who is reviewing the diff, and I have the feeling that proceeding like this is the way to add unnoticed bugs to an implementation.

Could you please revert back everything to the old behavior such that the diff it's just the *obvious* moving of code from a source file to an header without any other change?

src/mlabns/mlabns_impl.hpp
bassosimone: Now I see why the file here did not appear as a copy with small modifications: because clang-format was run and hence the two files differed too much. As a general rule, I think we SHOULD NOT (capitalized as in RFCs to convey the same meaning) combine refactoring and reformatting because that has the potential of hiding other changes, as I also argued below.

src/mlabns/mlabns_impl.hpp
```diff
@@ -0,0 +1,114 @@
...
+                reply.url = node.at("url");
```
bassosimone: I see that you have changed `node["city"]` in `node.at("city")`. As we discussed in private, I think this change is an improvement because the documentation of nlohmann/json clearly points out that `at()` throws if the element is missing, while it is not as clear on that point for `[]` (even though currently `[]` throws as well). But, at the same time, I think that in this case it would probably have been better to separate the refactoring and further changes (namely clang-format and applying this other change from `[]` to `at()`), to simplify the job of who is reviewing the diff.

AntonioLangiu:
I think this PR should also close #399. I'm writing in the first comment to be sure it works.

test/mlabns/mlabns.cpp
```diff
@@ -35,12 +38,201 @@ TEST_CASE("Query can pass the settings to the dns level") {
...
+        mlabns::query(tool,
+                      [](Error error, mlabns::Reply) {
+                          REQUIRE(error);
+                          break_loop();
+                      },
+                      settings);
+    });
+}
+
+MK_DEFINE_ERR(0001, MockError, "mock_error 0001")
```
bassosimone: Actually there is a `MockError` class defined in `<common/error.hpp>` that we should instead use here. I will apply this change as part of merging the pull request.

bassosimone:
Merging now! Thanks for working on this!

PR #607, opened 2016-05-17
configure: add -pthread to flags only when needed

bassosimone:
Prodded by @AntonioLangiu.

I have omitted the action to be performed when cross compiling, under the assumption that omitting the action leads to executing the main action specified by the macro also when cross compiling.

If my assumption is wrong, then perhaps this diff is broken. I shall now check whether this works by cross-compiling MK for iOS and then update on that in a short while.

- - - 

Update: the assumption was OK. Cross compiling for iOS went straight on.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -212,7 +212,6 @@ CXXFLAGS="$measurement_kit_saved_cxxflags $measurement_kit_cxx_stdlib_flags"
 AC_LANG_POP([C++])
 ])
 
-
 AC_DEFUN([MK_CHECK_CA_BUNDLE], [
   AC_MSG_CHECKING([CA bundle path])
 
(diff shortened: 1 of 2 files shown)


AntonioLangiu:
So are we ready to merge or you want to make some additional check? (maybe android?!)

bassosimone:
>  So are we ready to merge or you want to make some additional check? (maybe android?!)

Yes, probably also testing with Android could be wise. Will do and report back.

bassosimone:
> Yes, probably also testing with Android could be wise. Will do and report back.

Looking more closely the Android build is currently broken for many reasons, meaning that testing this diff with Android would require too much effort for this pull request.

I'd say that we proceed with merging this pull request, which reasonably does its job. I've just noticed, in fact, that `AC_COMPILE_IFELSE` does not take a fourth action in case of cross compiling. Hence we could rest assured that if it works properly on ordinary Unix system, this macro would also work when cross compiling for Android.

Regarding Android build, I am opening a meta ticket right now.

AntonioLangiu:
Ok, thank you 👍

PR #608, opened 2016-05-17
Fix warnings while building the tests

AntonioLangiu:
As in the description. The warnings are explained in the commit text.

### DIFF
--- a/test/dns/query.cpp
+++ b/test/dns/query.cpp
@@ -45,6 +45,11 @@ static int null_evdns_base_set_option_randomize(evdns_base *, const char *,
     return -1;
 }
 
+static int null_evdns_base_set_option_timeout(evdns_base *, const char *,
+                                                const char *) {
+    return -1;
+}
+
 #define BASE_FREE(name)                                                        \
     static bool base_free_##name##_flag = false;                               \
     static void base_free_##name(struct evdns_base *base, int fail_requests) { \
@@ -94,6 +99,18 @@ TEST_CASE("throw error while fails evdns_set_options for attempts") {
             {{"dns/attempts", "nexa"}}, Reactor::global());
     } catch (std::runtime_error &) {
         REQUIRE(base_free_evdns_set_options_attempts_flag);
+        return;
+    }
+    REQUIRE(false);
+}
+
+TEST_CASE("throw error while fails evdns_set_options for timeout") {
+    try {
+        create_evdns_base<::evdns_base_new, ::evdns_base_nameserver_ip_add,
+                          base_free_evdns_set_options_timeout>(
+            {{"dns/attempts", "nexa"}}, Reactor::global());
+    } catch (std::runtime_error &) {
+        REQUIRE(base_free_evdns_set_options_timeout_flag);
         return;
     }
     REQUIRE(false);
--- a/test/dns/query.cpp
+++ b/test/dns/query.cpp
@@ -46,7 +46,7 @@ static int null_evdns_base_set_option_randomize(evdns_base *, const char *,
 }
 
 static int null_evdns_base_set_option_timeout(evdns_base *, const char *,
-                                                const char *) {
+                                              const char *) {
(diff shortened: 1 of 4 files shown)


test/dns/query.cpp
```diff
@@ -99,6 +104,18 @@ TEST_CASE("throw error while fails evdns_set_options for attempts") {
...
 
+TEST_CASE("throw error while fails evdns_set_options for timeout") {
+    try {
+        create_evdns_base<::evdns_base_new, ::evdns_base_nameserver_ip_add,
+                          base_free_evdns_set_options_timeout>(
+            {{"dns/attempts", "nexa"}}, Reactor::global());
+    } catch (std::runtime_error &) {
+        REQUIRE(base_free_evdns_set_options_timeout_flag);
+        return;
+    }
```
bassosimone: I think this could be rewritten as:

```REQUIRE_THROWS_AS(( /* statements here surrounded by braces */), <ErrorClass>)```

Could you try to write it like this? I guess the thing that fixes the invocation of the macro is enclosing the statement or statements by another layer of braces. This is a trick I learned recently.

Could you try this solution for here and see whether it works?

Comment (bassosimone) [inline test/dns/query.cpp]: (If possible I'd like to use the native `Catch` facility rather than handcrafting our exception catching code)

File: test/ooni/tcp_test.cpp [hunk 2]
```diff
@@ -48,10 +48,10 @@ TEST_CASE("TCPTestImpl works as expected in a common case") {
                           {"port", "80"},
                           {"dns/nameserver", "8.8.8.1"},
                           {"dns/timeout", 0.0001}},
-                         [&count](Error err, Var<net::Transport> txp) {
+                         [&count](Error err, Var<net::Transport>) {
                              REQUIRE(err);
                              if (++count >= 3) {
-                                 []() { break_loop(); };
+                                 break_loop();
```
bassosimone: This makes me wonder how this test could have worked :-D

test/dns/query.cpp
AntonioLangiu: The problem was that I defined [this functions using the macro](https://github.com/measurement-kit/measurement-kit/pull/608/files/2a54277abae3fa3b3ba8ae678f92e0406e80f779#diff-8bb0df03508f949bf0f0057d91bdbf40R63), but it was never used, so I added a test to use it!

bassosimone:
Thanks! Much appreciated!

bassosimone:
(Removed `bikeshedding` because this is useful stuff. E.g., one warning was indication of a bug.)

test/ooni/tcp_test.cpp
AntonioLangiu: Because the [other test was ablet to call break loop](https://github.com/measurement-kit/measurement-kit/pull/608/files#diff-01d046af7ec7bc1862230cb8263d88e5R41)

test/dns/query.cpp
bassosimone: Yes, that was clear to me. I am just asking whether you could try rewriting that piece of code using Catch's native support for checking whether code excepts, as opposed to rolling out an handcrafted piece of code doing that functionality.

AntonioLangiu: Ah ok, I'll try!

test/ooni/tcp_test.cpp
bassosimone: Indeed! We just were lucky!

test/dns/query.cpp
AntonioLangiu: We are not able to use this Catch functionality while using templates. I did it so the first time for this reason.

bassosimone: Can you try by wrapping the template invocation with `( ... )` as suggested above? I think I encountered this problem when developing NDT and fixed it using one extra layer of braces.

bassosimone:
Thanks for working on this! Merging!

PR #613, opened 2016-05-18
Avoid duplicating code in dns_injection_impl and tcp_connect_impl

### LINKED ISSUE #338: ooni: factor code to validate input file path
The code to validate input file path is common to the dns-injection and tcp-connect tests. So, we could create a common function in the base class to reduce duplication.

AntonioLangiu:
Please see the issue #338 

This closes #338!


bassosimone:
Thanks, merging!

PR #627, opened 2016-05-22
Implement entry as private child of nlohmann::json

bassosimone:
We already discussed in the past that we would not like to expose directly nlohmann::json as part of our API, because it would be unpractical to bind us (and other users) so tightly do a dependency.

Yet, for exporting src/report headers, as well as for passing tests results in a manipulable format to users, it would be nice to have a type describing an entry.

At the same time, writing such type from scratch is not easy and mocking nlohmann/json without exposing it using a public header is overcomplicated, since in the latter case you need to do magic and use Var<T> to hide the real type used for the implementation.

Thus, I've come to the conclusion that it would be acceptable to include nlohmann/json to our public headers and to use it *only* to implement the entry type by means of private inheritance.

This way, consumers of our API use our interface for accessing an entry, which currently happens to be based on nlohmann/json but in the future could possibly be based on any other JSON library.

- - -

If this approach is accepted, before merging this pull request, it would probably make sense to amend code to always pull `<measurement_kit/ext/json.hpp>` rather than `"src/ext/json/src/json.hpp"`


hellais:
LGTM!

bassosimone:
Thank you!

PR #642, opened 2016-05-25
Fix the android build system

AntonioLangiu:
In this PR I'm starting sketching the new build script for Android.

The idea is to make it as similar as possible to the iOS build system, using the same principle of a `cross` script to set up the environment and a `dependency` script to build the dependencies. 

The changes applyed in this PR are:

1. Build in the root folder of MK for all the architectures (like in iOS), so remove the folder `build/android/build/` because is not more needed;
2. Put the result of the builds (dependencies and MK) in the jni forlder;

The changes to the specific files are:

`build/android/scripts/make_toolchain.sh`

1. Make the script robust to the directory form where it is called
2. Use the correct parameters for the new script introduced with the ndk `r11c` (before we were using r10e)

` build/android/scripts/build_target.sh`

1. Use a BUILDDIR variable to indicate the output dir
2. export the variable that will be used by the build/dependency script 
3. Check and remove unnecessary folders and files inside the build directory

This PR contains also some part that will be rewritten by next PR, for example the part to build the dependecies in ` build/android/scripts/build_target.sh`. I put it also here to make sure we are able to test that everything works before merging this, otherwise with just the diff on ` build/android/scripts/make_toolchain.sh ` you would be not able to test it!


AntonioLangiu:
@bassosimone I have a question about the problem with encounter yesterday when building libressl.
I tried to fix it setting the sysroot but It didn't work, and then I tried with --disable-asm and it worked. 

The question is: since the flag --disable-asm is just for libressl where should I put it? Now it is in the build script, but maybe it should be in the spec of the build dependency system, and if so we should introduce a variable like pkg_cross_building to the build dependency script because we are not able to detect it right now!

bassosimone:
@AntonioLangiu setting `--disable-asm` was meant as a temporary fix. Ideally, it should not be permanent and ASM should be enabled also on devices. I think it would be optimal to compile without `V=0` to see the exact assembly command invoked and see whether it is what we expected. Another way to understand what is going on would be to see whether there were recent fixes in libressl in this area and/or whether building with `master` branch of libressl fixes the build.

build/android/scripts/build_target.sh
```diff
@@ -73,27 +73,33 @@ export STRIP=${TOOL_PATH}-strip
...
+
 (
-    cd $ROOTDIR
-    install -d ${ROOTDIR}/build/${ARCH}-${API}
-    cd ${ROOTDIR}/build/${ARCH}-${API}
+    cd "${ROOTDIR}/../../"
     test -f Makefile && make clean
     echo "Configure with --host=${ARCH} and toolchain ${ANDROID_TOOLCHAIN}"
     test -x ${ROOTDIR}/../../configure || (cd ${ROOTDIR}/../.. && ./autogen.sh)
+    ${ROOTDIR}/../../build/dependency all
```
bassosimone: @AntonioLangiu Perhaps you could rewrite this line as:

```sh
${ROOTDIR}/../../build/dependency geoip
pkg_configure_flags="$pkg_configure_flags --disable-asm" ${ROOTDIR}/../../build/dependency libressl  # XXX
${ROOTDIR}/../../build/dependency libevent
```

This should pass the extra configure flag only to libressl and clearly highlight the temporary workaround.

What do you think? Would you like to try whether this approach works?

bassosimone:
Tested: it works! Thank you so much!

PR #643, opened 2016-05-25
android: add scripts to cross-build the dependencies

AntonioLangiu:
This PR is a Continuation of #642, and it should be reviewed and merged after #642, because it needs some modification to the make-toolchain script. 

I created two script (as in the iOS build) to build the dependencies. They are very similar to the `build/android/scripts/build_target.sh` and the `build/ios/dependencies` scripts.


bassosimone:
Thank you!

PR #648, opened 2016-06-03
Use docker to measure coverage, run valgrind

bassosimone:
I've figured out that the best way to measure coverage and run valgrind is through docker. This allows us to run the same scripts in many different environments (e.g. locally or on gitlab) and works around all the oddities of the standard travis-ci build environment.

Before merging, I'd like to cleanup this a bit more, and to test real integration with travis-ci. It could also make sense to enable other build environments based on docker.

Should fix the bug that clang++-3.7 is currently not available for travis because [clang developers have temporarily shut down their DEB repository](http://stackoverflow.com/questions/37558151/travis-ci-build-failing).


bassosimone:
Update: I was not able to run docker on travis-ci without flaws. For unspecified reasons, [most runs of Valgrind failed in that case](https://travis-ci.org/measurement-kit/measurement-kit/jobs/134996295#L1690). So, I decided to be very practical and removed all Valgrind builds from travis and benchmarked some other CI providers.

It turns out that GitLab provides something that is very close to our needs. So, I set up a clone of MK on GitLab and instructed GitLab to automatically pull from here periodically. For every pull, Valgrind tests would be run on docker using the infrastructure of GitLab.

All other tests, instead, have been migrated to use `gcc-5` and can stay on Travis.

.travis.yml
```diff
@@ -21,58 +18,40 @@ script:
...
+    - if [ "$BUILD_TYPE" != "ios" ]; then
         make -j2 check-am V=0;
       fi
 
 # Silence gcov standard output since it produces way too much output
 after_success:
     - if [ "$COVERAGE" = "true" ]; then
-        coveralls --gcov /usr/bin/gcov-4.6 --exclude src/ext
-                  --exclude example --exclude third_party > /dev/null;
+        coveralls --gcov /usr/bin/gcov-5 --exclude src/ext -b .
```
bassosimone: Adding `-b .` fixed running `gcov` for `gcc`, which was previously not possible.

.travis.yml
```diff
@@ -21,58 +18,40 @@ script:
...
+                  --exclude example --exclude third_party
+                  --exclude include/measurement_kit/ext > /dev/null;
```
bassosimone: I also added to `--exclude` third-party headers shipped along with measurement-kit.

.travis.yml
```diff
@@ -21,58 +18,40 @@ script:
...
-      env: OS="12.04" VALGRIND=true VER=3.6
-      addons:
-        apt:
-          packages:
-            - clang-3.6
-            - llvm-3.6-dev
-            - valgrind
-          sources:
-            - llvm-toolchain-precise-3.6
-            - ubuntu-toolchain-r-test
```
bassosimone: This entry was removed; the same functionality is now implemented using gitlab-ci.

PR #655, opened 2016-06-04
Implementation of OONI collector protocol

bassosimone:
- [x] initial implementation
- [x] use test collector in example
- [x] write unit tests
- [x] write integration tests
- [x] validate entry using regexps
- [x] cross check with spec
- [x] call this implementation from ooni tests
- [x] ensure 100% coverage of newly added files
- [x] resolve [memory leak](https://circleci.com/gh/measurement-kit/measurement-kit/27)


src/ooni/collector_client_impl.hpp
```diff
@@ -0,0 +1,289 @@
...
+    logger->info("closing report...");
+    collector_post(transport, "/report/" + report_id + "/close", "",
+                   [=](Error err, nlohmann::json reply) {
+                       logger->info("closing report... %d", err.code);
+                       if (err) {
+                           callback(err);
+                           return;
+                       }
+                       logger->debug("xxx response: %s", reply.dump().c_str());
+                       callback(NoError());
```
bassosimone: This could be simplified by calling `callback(err)` in any case

src/ooni/collector_client_impl.hpp
```diff
@@ -0,0 +1,289 @@
...
+                              if (response->status_code != 200) {
+                                  callback(HttpRequestFailedError(), "");
+                                  return;
+                              }
+                              // If response is empty, don't parse it
+                              if (response->body == "") {
+                                  callback(NoError(), nlohmann::json::object());
+                                  return;
+                              }
+                              nlohmann::json reply;
```
bassosimone: This could be moved above and also used as argument when body is empty

src/ooni/collector_client_impl.hpp
```diff
@@ -0,0 +1,289 @@
...
+            logger->debug("scheduling read of next entry...");
+            reactor->call_soon([=]() {
+                logger->debug("reading next entry");
+                ErrorOr<Entry> entry = get_next_entry(file, logger);
+                if (!entry) {
+                    if (entry.as_error() != FileEofError()) {
+                        callback(entry.as_error());
+                        return;
+                    }
+                    logger->info("closing report...");
```
bassosimone: This log line should be removed.

src/ooni/collector_client_impl.hpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+template <MK_MOCK_NAMESPACE(collector, post)>
+void update_report_impl(Var<Transport> transport, std::string report_id,
+                        Entry entry, Callback<Error> callback,
+                        Settings settings, Var<Reactor> reactor,
+                        Var<Logger> logger) {
+    // TODO: validate entry?
+    nlohmann::json request{
+        {"content", entry.dump()}, {"format", "json"},
```
bassosimone: As pointed out by @hellais, this should be `entry` and not `entry.dump()`

src/ooni/collector_client.cpp
```diff
@@ -0,0 +1,111 @@
...
+static std::map<std::string, std::regex> mandatory_re{
+    {"software_name", re_name},
+    {"software_version", re_version},
+    {"probe_asn", std::regex{"^AS[0-9]+$"}},
+    {"probe_cc", std::regex{"^[A-Z]{2}$"}},
+    {"test_name", re_name},
+    {"test_version", re_version},
+    {"data_format_version", re_version},
+    {"test_start_time",
+     std::regex{"^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}$"}},
```
bassosimone: Here I wrote what I consider reasonable regexps given the specification.

src/ooni/ooni_test_impl.hpp
```diff
@@ -206,13 +206,15 @@ class OoniTestImpl : public mk::NetTest {
         }
     }
 
-    /*!
-     * \brief Make sure that the report is written.
-     * \param cb Callback called when the report is written.
-     */
-    virtual void end(std::function<void()> cb) override {
+    void end(std::function<void()> cb) override {
```
bassosimone: In addition to removing the doxygen comment here I also removed `virtual` which is redundant because in the same line we are already saying that we `override`.

bassosimone:
I'm done with development, this branch is now *ready for review*. cc: @hellais

src/ooni/collector_client_impl.hpp
```diff
@@ -0,0 +1,277 @@
...
+    }
+    request["software_name"] = entry["software_name"];
+    request["software_version"] = entry["software_version"];
+    request["probe_asn"] = entry["probe_asn"];
+    request["probe_cc"] = entry["probe_cc"];
+    request["test_name"] = entry["test_name"];
+    request["test_version"] = entry["test_version"];
+    request["test_start_time"] = entry["test_start_time"];
+    if (entry["input_hashes"] != nullptr) {
+        request["input_hashes"] = entry["input_hashes"];
```
hellais: `input_hashes` key should default to an empty list.

include/measurement_kit/mlabns/mlabns.hpp
```diff
@@ -19,7 +19,7 @@ MK_DEFINE_ERR(MK_ERR_MLABNS(0), InvalidPolicyError, "")
 MK_DEFINE_ERR(MK_ERR_MLABNS(1), InvalidAddressFamilyError, "")
 MK_DEFINE_ERR(MK_ERR_MLABNS(2), InvalidMetroError, "")
 MK_DEFINE_ERR(MK_ERR_MLABNS(3), InvalidToolNameError, "")
-MK_DEFINE_ERR(MK_ERR_MLABNS(4), UnexpectedHttpStatusCodeError, "")
+//MK_DEFINE_ERR(MK_ERR_MLABNS(4), UnexpectedHttpStatusCodeError, "")
```
hellais: Is it ok to have this line commented out?

If we are not using it anymore (because `HttpRequestFailedError` is used instead), maybe we can just delete it?

src/ooni/collector_client.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+void submit_report(std::string filepath, std::string collector_base_url,
+                   Callback<Error> callback, Settings conf,
+                   Var<Reactor> reactor, Var<Logger> logger) {
+    submit_report_impl(filepath, collector_base_url, callback, conf, reactor,
+                       logger);
+}
+
+std::string default_collector_url() {
+    return "http://a.collector.test.ooni.io";
```
hellais: Maybe it's cleaner to have this as a `#DEFINE` inside of the header?

hellais:
Looking into this!

What I see looks very good!

* [x] First pass at reading diff
* [ ] Test with the test collector
* [ ] Merge 🎉

src/ooni/collector_client.cpp
bassosimone: Yes, will do!

include/measurement_kit/mlabns/mlabns.hpp
bassosimone: Yes, I agree it's better to delete it!

src/ooni/collector_client_impl.hpp
bassosimone: Okay, will add an `else` branch!

hellais:
I have tested it on OSX and everything seems to be working properly.

Once the outstanding comments are addressed I will proceed with merging!

bassosimone:
Thanks @hellais! All proposed changes have now been implemented in c852d85127!

hellais:
LGTM!

PR #659, opened 2016-06-05
ooni: Lookup IP address and GeoIP information

bassosimone:


src/ooni/ooni_test_impl.hpp
```diff
@@ -174,36 +205,33 @@ class OoniTestImpl : public mk::NetTest {
...
-     */
-    virtual void begin(std::function<void()> cb) override {
-        geoip_lookup();
-        open_report();
-        if (input_filepath != "") {
-            logger->debug("net_test: found input file");
-            if (input != nullptr) {
-                delete input;
-                input = nullptr;
+    void begin(Callback<> cb) override {
```
bassosimone: Here I removed the redundant `virtual` keyword, doxygen comment, and modernized the signature of the function using the more compact `Callback<>` alias rather than `std::function<>`.

bassosimone: The rest of the code [is exactly the same](https://github.com/measurement-kit/measurement-kit/pull/659/files?w=1) except that now it is executed after `geoip_lookup()` returns.

src/ooni/ooni_test_impl.hpp
```diff
@@ -69,10 +70,40 @@ class OoniTestImpl : public mk::NetTest {
             });
     }
 
-    void geoip_lookup() {
+    void geoip_lookup(Callback<> cb) {
         probe_ip = "127.0.0.1";
         probe_asn = "AS0";
         probe_cc = "ZZ";
+        ip_lookup([=](Error err, std::string ip) {
+            if (!err) {
```
hellais: Maybe it would be useful to have at least some logging be printed when the geoip lookup fails.

src/ooni/ooni_test_impl.hpp
```diff
@@ -69,10 +70,40 @@ class OoniTestImpl : public mk::NetTest {
...
+                    ErrorOr<nlohmann::json> res = geoip(ip, country_p, asn_p);
+                    if (!!res) {
+                        logger->debug("GeoIP result: %s", res->dump().c_str());
+                        try {
+                            probe_asn = (*res)["asn"];
+                            logger->info("probe_asn: %s", probe_asn.c_str());
+                            probe_cc = (*res)["country_code"];
+                            logger->info("probe_cc: %s", probe_cc.c_str());
+                        } catch (std::domain_error &) {
+                            logger->warn("failed to access geoip result");
```
hellais: Looking at the code for `geoip()` it seems like it will never throw a domain error. I guess it's good defensive coding to take that into account, but maybe take note of that in a comment or something.

hellais:
Overall this looks good.

I would say that the comments should be addressed, rebased onto master and then I will merge!

src/ooni/ooni_test_impl.hpp
bassosimone: Will add!

src/ooni/ooni_test_impl.hpp
bassosimone: Ah, you're right! I think it would be even better to let the exception unwind when there is an error here; since that is the usual path and errors here could only happen after some future, bad refactoring, I'd say we would notice for sure an exception while we may miss a warning. So, will remove the `try...catch` and instead add explanatory comment!

bassosimone:
Thank you, @hellais! Please, see 102699aba87, which should address your feedback!

hellais:
LGTM!

PR #660, opened 2016-06-05
Separate OONI logic and OONI test implementation

bassosimone:
- [x] initial implementation
- [x] explain (below) the basic purpose of this branch
- [ ] fix or at least report the bug of missing dns port that leads to incorrect `Entry` port value
- [ ] add tests for improved `Entry` (or alternative rewrite regress to use `nlohmann::json` directly)
- [ ] understand why when port was converted to string nothing happened (would have expected instead to see an exception because conversion is not possible)
- [ ] add test showing that the DNS template allows to run more query at once


bassosimone:
# Purpose of this branch

The main purpose of this branch is to separate the core OONI logic that drives testing over test lists and all the nice details such as getting the IP address and reporing from the code that instead deals with implementing a single test. In particular, I'd like the latter to become well definite pieces of code (i.e. functions) that you can invoke with specific input and test their output for conformance in regress tests.

I plan to implement this with minimal refactoring of existing code (so to minimize the diff and help review) and by adding as much regress tests as possible so to demonstrate that what is done here helps.

include/measurement_kit/report/entry.hpp
```diff
@@ -26,6 +26,13 @@ class Entry : private nlohmann::json {
             return JsonDomainError();
```
bassosimone: *Note*: we can avoid the changes in this file by using `nlohmann::json` in the unit test. Perhaps this would be better.

src/ooni/dns_injection_impl.hpp
```diff
@@ -27,19 +27,29 @@ class DNSInjectionImpl : public DNSTestImpl {
...
+                      (*entry)["injected"] = true;
                   } else {
-                      entry["injected"] = false;
+                      (*entry)["injected"] = false;
                   }
-                  cb(entry);
-              }, options);
+                  callback(error, entry);
+              }, options, reactor, logger);
     }
```
bassosimone: *Note*: I've not reindented this code to avoid complicating the diff.

src/ooni/dns_test_impl.hpp
```diff
@@ -68,13 +59,12 @@ class DNSTestImpl : public ooni::OoniTestImpl {
...
                 // TODO add support for bytes received
                 // query_entry["bytes"] = response.get_bytes();
-                entry["queries"].push_back(query_entry);
+                (*entry)["queries"].push_back(query_entry);
                 logger->debug("dns_test: callbacking");
-                cb(message);
+                cb(error, message);
                 logger->debug("dns_test: callback called");
             }, options, reactor);
     }
```
bassosimone: *Note*: I've not re-indented the code to avoid complicating the diff.

src/report/entry.cpp
```diff
@@ -37,6 +37,14 @@ void Entry::push_back(Entry value) {
...
 }
 
+Entry &Entry::operator[](int key) {
+    try {
+        return static_cast<Entry &>(nlohmann::json::operator[](key));
+    } catch (std::domain_error &) {
+        throw JsonDomainError();
+    }
+}
+
```
bassosimone: *Note*: if we use `nlohmann::json` in the unit test, this is unnecessary.

test/ooni/dns_injection.cpp
```diff
@@ -5,11 +5,89 @@
...
-using namespace mk;
 using namespace mk::ooni;
+using namespace mk::report;
+using namespace mk;
+
+TEST_CASE("Check dns-injection output for injected entry") {
+    loop_with_initial_event([=]() {
+        dns_injection("8.8.8.8:53", "nexa.polito.it", [=](Error err,
+                                                          Var<Entry> entry) {
+            // We have called the DNS injection test with an actually
```
bassosimone: The main advantage of this new code is that, as shown here, we can call and test directly the code for running this specific test (i.e. we can test a smaller bit of code and go towards precise unit testing)

test/ooni/dns_injection.cpp
```diff
@@ -5,11 +5,89 @@
...
+            Entry answers = first_query["answers"];
+            REQUIRE(answers.is_array());
+            REQUIRE(answers.size() == 1);
+            Entry first_answer = answers[0];
+            std::string answer_type = first_answer["answer_type"];
+            REQUIRE(answer_type == "A");
+            std::string ipv4 = first_answer["ipv4"];
+            REQUIRE(ipv4 == "130.192.16.172");
+            int ttl = first_answer["ttl"];
+            REQUIRE(ttl >= 0);
```
bassosimone: The above is quite ugly, but I doubt that it would be much better if using `nlohmann::json`. It could perhaps be possible, using `nlohmann::json` to define an expected object to use as template to verify the received one.

include/measurement_kit/report/entry.hpp
bassosimone: On second thought, users would probably use the `Entry` to parse the result. So it would probably make sense to make `Entry` easier to use, as I've tried to do here.

src/ooni/dns_test_impl.hpp
```diff
@@ -5,27 +5,18 @@
...
-        : ooni::OoniTestImpl(input_filepath_, options_) {
-        test_name = "dns_test";
-        test_version = "0.0.1";
-    };
+using namespace mk::report;
 
-    void query(dns::QueryType query_type, dns::QueryClass query_class,
-               std::string query_name, std::string nameserver,
-               std::function<void(dns::Message)> cb, Settings options = {}) {
+    void dns_template(Var<Entry> entry, std::string nameserver,
```
hellais: I would call this not `dns_template`, but rather `dns_query`

bassosimone:
Superseded by #677

PR #666, opened 2016-06-06
Make sure JSON reports are compliant with spec

bassosimone:
1) change lifecycle of `entry` object: rather than keeping it inside
   the measurement class, create it before running a test, pass it to
   test helper and finally pass it back to be inserted into the keys
   of the test into the final entry.

   Judging from the documentation on github, this seems quite the
   right way of proceedings as reports look much more like what they
   oughta be, except for dns-injection for which a real JSON report
   is not provided in github but the result feels right.

2) make sure that tests running with no input receive exactly same
   treatment of tests with input as regards variables that should
   go inside the report (e.g. test runtime)

3) make sure that we use the two digit country names as opposed
   to three digit country names, to follow OONI spec

4) remove some unneeded semi colons

- - -

Given the pull request number, I think some music can help:

[The number of the beast](https://www.youtube.com/watch?v=7-iRf9AWoyE)

### DIFF
--- a/src/ooni/dns_injection_impl.hpp
+++ b/src/ooni/dns_injection_impl.hpp
@@ -23,20 +23,21 @@ class DNSInjectionImpl : public DNSTestImpl {
         test_version = "0.0.1";
 
         validate_input_filepath();
-    };
+    }
 
     void main(std::string input, Settings options,
               std::function<void(report::Entry)> &&cb) {
-        entry["injected"] = nullptr;
-        query("A", "IN", input, options["backend"],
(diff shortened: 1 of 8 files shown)


hellais:
Yes I very much like the design of this code, better than how it was before. Creating the `Entry` as a `Var` and passing it around to alter it is I think safer.

At the first pass of reading the diff I have no comments to add.

I will test this and merge it.

hellais:
I can confirm this is working as expected.

Merging!

PR #670, opened 2016-06-07
Fixed request serializer

helloimcarmine:
I've fixed the possibility to use an explicit path during an HTTP request. Moreover I've added a test for this case.


bassosimone:
Beware that I cannot even attempt merging because of conflicts with master.

example/ndt/ndt
```diff
@@ -0,0 +1,210 @@
...
+      func_exec_program ${1+"$@"}
+    fi
+  else
+    # The program doesn't exist.
+    $ECHO "$0: error: '$progdir/$program' does not exist" 1>&2
+    $ECHO "This script is just a wrapper for $program." 1>&2
+    $ECHO "See the libtool documentation for more information." 1>&2
+    exit 1
+  fi
+fi
```
bassosimone: You should definitely not commit this file!

example/net/discard
```diff
@@ -0,0 +1,210 @@
...
+      func_exec_program ${1+"$@"}
+    fi
+  else
+    # The program doesn't exist.
+    $ECHO "$0: error: '$progdir/$program' does not exist" 1>&2
+    $ECHO "This script is just a wrapper for $program." 1>&2
+    $ECHO "See the libtool documentation for more information." 1>&2
+    exit 1
+  fi
+fi
```
bassosimone: Ditto

example/neubot/dash
```diff
@@ -0,0 +1,210 @@
...
+      func_exec_program ${1+"$@"}
+    fi
+  else
+    # The program doesn't exist.
+    $ECHO "$0: error: '$progdir/$program' does not exist" 1>&2
+    $ECHO "This script is just a wrapper for $program." 1>&2
+    $ECHO "See the libtool documentation for more information." 1>&2
+    exit 1
+  fi
+fi
```
bassosimone: Ditto!

bassosimone:
Uhm, it seems you committed files autogenerated by libtool. I'd remove then first and see whether this improves the state in terms of mergeability.

bassosimone:
Uhm, still not possible to merge. Could you make sure that your master branch is up to date and then issue the `git merge master` command in your work branch and fix the possible conflicts at hand? Thank you.

src/http/request_serializer.hpp
```diff
@@ -39,15 +40,32 @@ class RequestSerializer {
...
+        path = settings.get("path", std::string(""));
+        if (path != "" && path[0] != '/') {
+            path = "/" + path;
+        }
+
+=======
         url = parse_url(settings.at("http/url"));
         protocol = settings.get("http/http_version", std::string("HTTP/1.1"));
         method = settings.get("http/method", std::string("GET"));
+>>>>>>> master
```
bassosimone: @helloimcarmine you should manually solve the conflict here by choosing which version of the two you prefer or cherry picking the right lines from the two.

src/http/request_serializer.hpp
```diff
@@ -39,15 +40,26 @@ class RequestSerializer {
         if (settings.find("http/url") == settings.end()) {
             throw MissingUrlError();
         }
-        url = parse_url(settings.at("http/url"));
-        protocol = settings.get("http/http_version", std::string("HTTP/1.1"));
-        method = settings.get("http/method", std::string("GET"));
+
+        url = parse_url(settings.at("url"));
+        protocol = settings.get("http_version", std::string("HTTP/1.1"));
+        method = settings.get("method", std::string("GET"));
```
bassosimone: This is wrong: you have changed the keys!

test/http/request_serializer.cpp
```diff
@@ -41,3 +41,29 @@ TEST_CASE("HTTP Request serializer works as expected") {
     expect += "0123456789";
     REQUIRE(serialized == expect);
 }
+
+TEST_CASE("HTTP Request serializer works as expected with explicit path") {
+    auto serializer = RequestSerializer(
+        {
+         {"follow_redirects", "yes"},
+         {"url", "http://www.example.com/antani?clacsonato=yes#melandri"},
+         {"path", "/antani?clacsonato=no#melandri"}
```
bassosimone: Could you pass something more different so it's easier to spot it?

bassosimone:
@helloimcarmine It seems [g++ does not like it when you pass structured default arguments](https://travis-ci.org/measurement-kit/measurement-kit/jobs/135913190#L841). Perhaps this could be fixed by improving `Settings`, but I advise instead to take another route to implement the feature.

test/http/request_serializer.cpp
```diff
@@ -41,3 +41,29 @@ TEST_CASE("HTTP Request serializer works as expected") {
     expect += "0123456789";
     REQUIRE(serialized == expect);
 }
+
+TEST_CASE("HTTP Request serializer works as expected with explicit path") {
+    auto serializer = RequestSerializer(
+        {
+         {"http/follow_redirects", "yes"},
+         {"http/url", "http://www.example.com/antani?clacsonato=yes#melandri"},
+         {"http/path", "/antani?amicimiei=yes#melandri"}
```
bassosimone: Here there is a missing colon at end of line

bassosimone:
Oh, please disregard my previous comment which was wrong! I've added an inline comment to point out the specific place where the build fails and why.

helloimcarmine:
Try now, I've added the missing colon and fixed the expected string, which was wrong. In the previous version a structured default argument was passed too, why is it not accepted now?

bassosimone:
> In the previous version a structured default argument was passed too, why is it not accepted now?

See:

>> Oh, please disregard my previous comment which was wrong!

helloimcarmine:
Ok sorry, I've understood that it was referred to the missing colon!

bassosimone:
No problem!

bassosimone:
The diff looks good to me! Merging!

PR #674, opened 2016-06-09
Changes and refactoring after code review

bassosimone:
This pull request is here to group incremental, small changes since v0.2.0-beta.1, to be merged into master in a single batch as soon as we collect enough of them.


include/measurement_kit/http/http.hpp
```diff
@@ -58,10 +63,26 @@ ErrorOr<Url> parse_url_noexcept(std::string url);
...
+    Url url;
+    std::string protocol;
+    Headers headers;
+    std::string path;
+    std::string body;
+
+    Request() {}
+    Error init(Settings, Headers, std::string);
+    void serialize(net::Buffer &);
+};
```
bassosimone: This class used to be called `RequestSerialized`. Now it's `Request`. I've made it public because I plan soon to chain Requests and Responses, which is useful *ex post* to inspect what really happened when you are redirected in the `web_connectivity` test. (cc: @hellais)

include/measurement_kit/http/http.hpp
```diff
@@ -77,17 +98,34 @@ void request_connect(Settings, Callback<Error, Var<net::Transport>>,
...
+ *       {"http/follow_redirects", boolean},
+ *       {"http/url", std::string},
+ *       {"http/ignore_body", boolean},
+ *       {"http/method", "GET|DELETE|PUT|POST|HEAD|..."},
+ *       {"http/http_version", "HTTP/1.1"},
+ *       {"http/path", by default is taken from the url}
+ *     }
+ *
+ * Currently `http/follow_redirects` is not supported.
+ */
```
bassosimone: This commented was along with `RequestSerialized`. Moved it here where it's most likely to inform users.

src/http/parse_url.cpp
```diff
@@ -2,17 +2,17 @@
...
 namespace mk {
 namespace http {
 
 Url parse_url(std::string url) {
     Url retval;
     http_parser_url url_parser;
-    http_parser_url_init(&url_parser); 
-    if (http_parser_parse_url(url.c_str(), url.length(), 0, &url_parser) != 0) {
+    http_parser_url_init(&url_parser);
+    if (http_parser_parse_url(url.data(), url.size(), 0, &url_parser) != 0) {
```
bassosimone: Use my preferred C-string access functions (the `C++11` ones; more consistent, imho)

src/http/parse_url.cpp
```diff
@@ -38,8 +40,8 @@ Url parse_url(std::string url) {
     }
     retval.pathquery = retval.path;
     if ((url_parser.field_set & (1 << UF_QUERY)) != 0) {
-        retval.query += url.substr(url_parser.field_data[UF_QUERY].off,
-                                   url_parser.field_data[UF_QUERY].len);
+        retval.query = url.substr(url_parser.field_data[UF_QUERY].off,
+                                  url_parser.field_data[UF_QUERY].len);
```
bassosimone: The `+=` rather than `=` was actually a bug (innocuous but still a bug)

src/http/request.cpp
```diff
@@ -2,106 +2,180 @@
...
+
+    // TODO: here we should honour the `ignore_body` setting
     parser->on_body([=](std::string s) { response->body += s; });
 
-    // XXX/TODO: we should improve the parser such that the transport
-    // forwards the "error" event to it and then the parser does the right
-    // thing, so that the code becomes less "twisted" here.
+    // TODO: we should improve the parser such that the transport forwards the
+    // "error" event to it and then the parser does the right thing, so that the
+    // code becomes less "twisted" here.
```
bassosimone: Declassed as `TODO`. I do not think that is unclean anymore.

test/http/request.cpp
```diff
@@ -3,83 +3,204 @@
...
+
+    unsigned char digest[MD5_DIGEST_LENGTH];
+    MD5((const unsigned char *)s.data(), s.size(), digest);
+    std::string retval;
+    for (int i = 0; i < MD5_DIGEST_LENGTH; ++i) {
+        retval += table[((digest[i] & 0xf0) >> 4)];
+        retval += table[(digest[i] & 0x0f)];
+    }
+    return retval;
+}
```
bassosimone: Added because I want to make more assertive tests. Previously, as noted by @AntonioLangiu, too many tests in the `http` suite were not performing assertions, which is bad.

test/http/request.cpp
```diff
@@ -3,83 +3,204 @@
...
 using namespace mk::net;
 using namespace mk::http;
 
-// TODO: after refactoring there are probably duplicated tests, we
-// should merge them to avoid testing things twice
+// Either tor was running and hence everything should be OK, or tor was
+// not running and hence connect() to socks port must have failed.
+static inline bool check_error_after_tor(Error e) {
+    return e == NoError() or e == ConnectFailedError();
+}
```
bassosimone: Just moved from below. This, for now, allows to cover the cases where Tor is running and where Tor instead is not running, but (note to self!) it would probably better to have a specific SOCKS5 err.

test/http/request.cpp
```diff
@@ -3,83 +3,204 @@
...
-                }
-                std::cout << "\r\n";
-                std::cout << response->body.substr(0, 128) << "\r\n";
-                std::cout << "[snip]\r\n";
+            "",
+            [](Error error, Var<Response> response) {
+                REQUIRE(!error);
+                REQUIRE(response->status_code == 200);
+                REQUIRE(md5(response->body) ==
+                        "efa2a8f1ba8a6335a8d696f91de69737");
```
bassosimone: Specific body check

test/http/request.cpp
```diff
@@ -3,83 +3,204 @@
...
-                    }
-                    std::cout << "\r\n";
-                    std::cout << response->body.substr(0, 128) << "\r\n";
-                    std::cout << "[snip]\r\n";
+                "",
+                [](Error error, Var<Response> response) {
+                    REQUIRE(!error);
+                    REQUIRE(response->status_code == 200);
+                    REQUIRE(md5(response->body) ==
+                            "1be9d96d157a3df328faa30e51faf63a");
```
bassosimone: Specific body check

test/http/request.cpp
```diff
@@ -3,83 +3,204 @@
...
+            {
+                {"Accept", "*/*"},
+            },
+            "",
+            [&](Error error, Var<Response> response) {
+                REQUIRE(check_error_after_tor(error));
+                if (!error) {
+                    REQUIRE(response->status_code == 200);
+                    REQUIRE(md5(response->body) ==
+                            "efa2a8f1ba8a6335a8d696f91de69737");
```
bassosimone: Specific body check

test/http/request.cpp
```diff
@@ -3,83 +3,204 @@
...
+                {"http/method", "GET"},
+                {"http/http_version", "HTTP/1.1"},
+                {"net/timeout", "3.0"},
+            },
+            {
+                {"Accept", "*/*"},
+            },
+            "",
+            [](Error error, Var<Response> response) {
+                REQUIRE(error);
```
bassosimone: Here we should probably check for connection refused *or* timeout (guess that inside and outside polito things change significantly).

test/http/request.cpp
```diff
@@ -113,70 +234,6 @@ TEST_CASE("http::request_recv_response() behaves correctly when EOF "
...
-                for (auto &kv : response->headers) {
-                    std::cout << kv.first << ": " << kv.second << "\r\n";
-                }
-                std::cout << "\r\n";
-                std::cout << response->body.substr(0, 128) << "\r\n";
-                std::cout << "[snip]\r\n";
-                break_loop();
-            });
-    });
-}
```
bassosimone: Removed because there is already a test like this

test/http/request.cpp
```diff
@@ -113,70 +234,6 @@ TEST_CASE("http::request_recv_response() behaves correctly when EOF "
...
-
```
bassosimone: Duplicated an existing test

test/http/request.cpp
```diff
@@ -376,37 +438,28 @@ TEST_CASE("http::request_sendrecv() works for multiple requests") {
...
-        request({{"http/method", "GET"},
-                       {"http/url", "http://www.google.com/robots.txt"}},
-                      {}, "", [](Error error, Var<Response> r) {
-                          REQUIRE(!error);
-                          REQUIRE(r->status_code == 200);
-                          REQUIRE(r->body.size() > 0);
-                          break_loop();
-                      });
-    });
-}
```
bassosimone: Duplicated an existing test

test/http/request.cpp
```diff
@@ -376,37 +438,28 @@ TEST_CASE("http::request_sendrecv() works for multiple requests") {
...
-
-// Either tor was running and hence everything should be OK, or tor was
-// not running and hence connect() to socks port must have failed.
-static inline bool check_error_after_tor(Error e) {
-    return e == NoError() or e == ConnectFailedError();
-}
-
```
bassosimone: Moved at top

test/http/request.cpp
```diff
@@ -376,37 +438,28 @@ TEST_CASE("http::request_sendrecv() works for multiple requests") {
...
                 if (!error) {
-                    REQUIRE(r->status_code == 200);
-                    REQUIRE(r->body.size() > 0);
+                    REQUIRE(response->status_code == 200);
+                    nlohmann::json body = nlohmann::json::parse(response->body);
+                    REQUIRE(body["default"]["collector"] ==
+                            "httpo://ihiderha53f36lsd.onion");
+                    REQUIRE(body["dns"]["collector"] ==
+                            "httpo://ihiderha53f36lsd.onion");
+                    REQUIRE(body["dns"]["address"] == "213.138.109.232:57004");
```
bassosimone: Here we make sure the value returned by collector is what ought to be

test/http/request_serializer.cpp
```diff
@@ -1,70 +0,0 @@
...
-    serializer.serialize(buffer);
-    auto serialized = buffer.read();
-    std::string expect = "GET /antani?amicimiei HTTP/1.0\r\n";
-    expect += "User-Agent: Antani/1.0.0.0\r\n";
-    expect += "Host: www.example.com\r\n";
-    expect += "Content-Length: 10\r\n";
-    expect += "\r\n";
-    expect += "0123456789";
-    REQUIRE(serialized == expect);
-}
```
bassosimone: Tests moved in test/http/request.cpp

test/http/funcs.cpp
```diff
@@ -1,45 +0,0 @@
...
-                      "",
-                      [](Error error, Var<Response> response) {
-                          std::cout << "Error: " << (int)error << "\r\n";
-                          if (response) {
-                              std::cout << response->body.substr(0, 128) << "\r\n";
-                              std::cout << "[snip]\r\n";
-                          }
-                          break_loop();
-                      });
-    });
```
bassosimone: Redundant test because `get()` is just an inline wrapper and we already use it in many other tests

test/http/funcs.cpp
```diff
@@ -1,45 +0,0 @@
...
-                  [](Error error, Var<Response> response) {
-                      std::cout << "Error: " << (int)error << "\r\n";
-                      if (response) {
-                          std::cout << response->body.substr(0, 128) << "\r\n";
-                          std::cout << "[snip]\r\n";
-                      }
-                      break_loop();
-                  });
-    });
-}
```
bassosimone: A test like this already exists in test/http/request.cpp

test/http/client.cpp
```diff
@@ -1,223 +0,0 @@
...
-
-TEST_CASE("Make sure that settings are not modified") {
-
-    // XXX Since settings are copied when passed to request() I am not
-    // sure now what was the purpose of this test, or whether this test
-    // has always been conceptually wrong since it cannot fail.
-    //
-    // Leaving it here for now, added a comment to trigger future
-    // investigation when we cleanup http tests in the aftermath of
-    // the current wave of MK cleanups.
```
bassosimone: Removed for the reason explained in the comment

test/http/client.cpp
```diff
@@ -1,223 +0,0 @@
...
-                if (++count >= 3) {
-                    break_loop();
-                }
-            });
-
-    });
-}
-
-TEST_CASE("Make sure that we can access OONI's bouncer using httpo://...") {
-    loop_with_initial_event_and_connectivity([]() {
```
bassosimone: Test moved to test/http/request.cpp and merged with a similar test

test/http/client.cpp
```diff
@@ -1,223 +0,0 @@
...
-                    break_loop();
-                }
-            });
-
-    });
-}
-
-TEST_CASE("http::request() works as expected over Tor") {
-    auto count = 0;
-    loop_with_initial_event_and_connectivity([&]() {
```
bassosimone: Similar test already in test/http/request.cpp

test/http/client.cpp
```diff
@@ -1,223 +0,0 @@
...
-using namespace mk::http;
-
-// TODO: for now I have just refactored former http::Client tests to use
-// instead the http::request() API, we should now remove duplicates!
-
-// TODO: these tests should go in test/http/request.cpp
-
-TEST_CASE("http::request() works as expected") {
-    auto count = 0;
-    loop_with_initial_event_and_connectivity([&]() {
```
bassosimone: Similar test already in test/http/request.cpp

bassosimone:
- [ ] create tickets for the small issues evidenced during code review
- [ ] merge pull request

include/measurement_kit/net/connect.hpp
```diff
@@ -0,0 +1,46 @@
...
+using ConnectManyCb = Callback<Error, std::vector<Var<Transport>>>;
+
+void connect_many(std::string address, int port, int num,
+        ConnectManyCb callback, Settings settings = {},
+        Var<Logger> logger = Logger::global(),
+        Var<Reactor> reactor = Reactor::global());
+
+} // namespace net
+} // namespace mk
+#endif
```
bassosimone: Code moved here from `include/measurement_kit/net/transport.hpp`

src/net/connect-many.cpp
```diff
@@ -1,18 +0,0 @@
...
-
-void connect_many(std::string address, int port, int num,
-        ConnectManyCb callback, Settings settings, Var<Logger> logger,
-        Var<Reactor> reactor) {
-    connect_many_<net::connect>(connect_many_make(address, port, num,
-            callback, settings, logger, reactor));
-}
-
-} // namespace net
-} // namespace mk
```
bassosimone: Merged with `src/net/connect.{c,h}pp`

src/net/connect-many.hpp
```diff
@@ -1,67 +0,0 @@
...
-    ctx->port = port;
-    ctx->settings = settings;
-    ctx->logger = logger;
-    ctx->reactor = reactor;
-    return ctx;
-}
-
-} // namespace net
-} // namespace mk
-#endif
```
bassosimone: Merged with `src/net/connect.{c,h}pp`

src/net/connect.cpp
```diff
@@ -205,5 +217,71 @@ void connect_ssl(bufferevent *orig_bev, ssl_st *ssl,
...
+            Var<Transport> txp =
+                Connection::make(r->connected_bev, reactor, logger);
+            txp->set_timeout(timeout);
+            assert(err == NoError());
+            err.context = r;
+            callback(err, txp);
+        },
+        settings, reactor, logger);
+}
+
```
bassosimone: Code moved verbatim from other places

src/net/connect.hpp
```diff
@@ -33,78 +31,44 @@ void mk_bufferevent_on_event(bufferevent *, short, void *);
...
-            new Callback<Error, bufferevent *>([cb](Error err, bufferevent *bev) {
-                if (err) {
-                    bufferevent_free(bev);
-                    cb(err, nullptr);
-                    return;
-                }
-                cb(err, bev);
-            }));
-}
-
```
bassosimone: Moved in `src/net/connect_impl.hpp` (newly created file)

test/net/connect-many.cpp
```diff
@@ -1,81 +0,0 @@
-// Part of measurement-kit <https://measurement-kit.github.io/>.
```
bassosimone: Merged with `test/net/connect.cpp`

src/http/parse_url.cpp
```diff
@@ -48,7 +50,7 @@ Url parse_url(std::string url) {
 ErrorOr<Url> parse_url_noexcept(std::string url) {
     try {
         return parse_url(url);
-    } catch (Error &error) {
+    } catch (Error error) {
```
bassosimone: This is actually a bug I introduced, damn it

src/net/ssl-context.cpp
```diff
@@ -18,19 +18,24 @@ namespace net {
...
-    OpenSSL_add_all_algorithms();
+    static bool ssl_initialized = false;
+
+    if (!ssl_initialized) {
+        SSL_library_init();
+        ERR_load_crypto_strings();
+        SSL_load_error_strings();
+        OpenSSL_add_all_algorithms();
+        ssl_initialized = true;
+    }
```
bassosimone: @hellais, can you confirm that here we only need to call these OpenSSL functions once?

hellais: Yes this only needs to be called once. It's actually a good idea to do so since most of these functions are not reentrant.

src/net/ssl-context.cpp
```diff
@@ -75,7 +80,7 @@ SslContext::SslContext(std::string ca_bundle_path) {
 }
 
 SslContext::~SslContext() {
-    EVP_cleanup();
+    //EVP_cleanup();  // XXX Should this be called on every destruction?
```
bassosimone: @hellais, here's another ping: what do you think of the above change?

hellais: So if we make this into a singleton the call to `OpenSSL_add_all_algorithms` will happen only once and if this does get called on every destruction it will mean that they will be unloaded from the table, so I guess it's good to remove it.

Moreover since EVP_cleanup() has no more effect since OpenSSL 1.1.0 we can I think not worry too much about the extra memory usage for older versions.

bassosimone:
Before merging, I'd like to comment on why cc4fd14 broke the build. Basically, that comment ensured that OpenSSL initialization was *actually* called once. Unfortunately, before 242acdf, `~SslContext` called `EVP_cleanup()`, meaning that -- after `~SslContext` was called - OpenSSL globals were not initialized anymore, and thus [subsequent attempts to initialize a context failed](https://travis-ci.org/measurement-kit/measurement-kit/jobs/136821551#L1581). After 242acdf, instead, `EVP_cleanup()` was not called anymore, and thus OpenSSL globals remained initialized.

src/net/ssl-context.cpp
```diff
@@ -75,7 +80,6 @@ SslContext::SslContext(std::string ca_bundle_path) {
 }
 
 SslContext::~SslContext() {
-    EVP_cleanup();
```
bassosimone: This was actually a bug that caused us to de-initialize OpenSSL globals every time a `SslContext` died

PR #686, opened 2016-06-17
ndt: save results and propagate errors

bassosimone:


bassosimone:
I believe we can get rid of this old pull request. The most interesting parts of it have now been merged anyway.

PR #689, opened 2016-06-23
Fix net/connect timeout

bassosimone:
Make sure the timeout is correctly passed to connect_logic().

While there, also fix other inconsistencies.

Bugs spotted by @hellais.


hellais:
LGTM!

Issue #676, opened 2016-06-12
http::request() does not save response line


bassosimone:

bassosimone:
Fixed in #690

PR #690, opened 2016-06-24
Feature/web connectivity

hellais:
I am going to use this PR to keep track of the web_connectivity test development.

It's not yet ready to be merged.


hellais:
I fixed the unittests and I would say this branch is now ready to be reviewed.

What is missing is the fix for handling the situation where the TCP connection is terminated without a proper SSL closing handshake. We did do some debugging of this with @bassosimone and I think it's clear how to fix that. I think we can either commit those changes to another branch or on top of this one directly.

bassosimone:
@hellais Awesome!!! Sorry for being so slow but I found lots of backlog accumulated after the week with Nexa board and I am slowly coping with that! Will review this soon!

bassosimone:
To move forward this branch I think we should focus on #711 first and then merge master into this branch and see whether that is enough to fix the issues we experienced

hellais:
I am testing now #711 with this branch to see if it resolves the issue.

src/http/parse_url.cpp
```diff
@@ -55,5 +55,23 @@ ErrorOr<Url> parse_url_noexcept(std::string url) {
     }
 }
 
+std::string Url::str() {
+    std::stringstream sst;
+    sst << schema;
+    sst << "://";
+    sst << address;
+    if ((schema == "http" and port != 80) or
+        (schema == "https" and port != 443)) {
```
bassosimone: @hellais one issue with fsrn.org was that we constructed the URL contained `443` and this lead to a redirect loop; the fix was to make sure no port was added when the port is the default one for a schema.

src/http/request.cpp
```diff
@@ -53,7 +66,8 @@ void Request::serialize(net::Buffer &buff) {
         buff << kv.first << ": " << kv.second << "\r\n";
     }
     buff << "Host: " << url.address;
-    if (url.port != 80) {
+    if ((url.schema == "http" and url.port != 80) or
+         (url.schema == "https" and url.port != 443)) {
```
bassosimone: @hellais one issue with fsrn.org was that we constructed the URL contained `443` and this lead to a redirect loop; the fix was to make sure no port was added when the port is the default one for a schema.

src/http/request.cpp
```diff
@@ -62,6 +76,11 @@ void Request::serialize(net::Buffer &buff) {
         buff << "Content-Length: " << std::to_string(body.length()) << "\r\n";
     }
     buff << "\r\n";
+    for (auto s: mk::split(buff.peek(), "\r\n")) {
+        // TODO: break API, bump minor, pass Var<Logger> to this function
+        // such that this log message is printed with the correct logger
```
bassosimone: If we add this functionality, we need to remember to bump version to `0.3.0-alpha`

src/http/request.cpp
```diff
@@ -82,59 +101,59 @@ void request_connect(Settings settings, Callback<Error, Var<Transport>> txp,
...
+
     parser->on_response([=](Response r) { *response = r; });
 
     // TODO: here we should honour the `ignore_body` setting
     parser->on_body([=](std::string s) { response->body += s; });
 
-    // TODO: we should improve the parser such that the transport forwards the
-    // "error" event to it and then the parser does the right thing, so that the
-    // code becomes less "twisted" here.
-
```
bassosimone: `TODO` addressed by duplicating few lines of code and removing the `prevent_emit` hack

src/http/request.cpp
```diff
@@ -82,59 +101,59 @@ void request_connect(Settings settings, Callback<Error, Var<Transport>> txp,
...
-            *prevent_emit = true;
-            parser->eof();
+            try {
+                logger->debug("Now passing EOF to parser");
+                parser->eof();
+            } catch (Error &second_error) {
+                logger->warn("Parsing error at EOF: %d", second_error.code);
+                err = second_error;
+                // FALLTHRU
+            }
```
bassosimone: Added `try...catch` to properly deal with parser throwing exception on error (this is one of the things that happened with fsrn.org until I figured out how to deal properly with EOF followed by data)

src/http/response_parser.hpp
```diff
@@ -93,6 +93,14 @@ class ResponseParserNg : public NonCopyable, public NonMovable {
         response_.http_major = parser_.http_major;
         response_.status_code = parser_.status_code;
         response_.http_minor = parser_.http_minor;
+        std::stringstream sst;
+        sst << "HTTP/" << response_.http_major << "." << response_.http_minor
+            << " " << response_.status_code << " " << response_.reason;
+        response_.response_line = sst.str();
```
bassosimone: These lines of code address #676

src/http/response_parser.hpp
```diff
@@ -157,9 +165,6 @@ class ResponseParserNg : public NonCopyable, public NonMovable {
         prev_ = cur;
     }
 
-    // XXX parse() and parser_execute() should return Error
-    // and should be tagged with warn_unused_result
-
```
bassosimone: Actually, I noticed that exceptions in this case are not that bad, because one gets automatic routing to `handle_error` if an exception occurs during `handle_data`. I should probably write regress for this, tho.

src/net/connection.cpp
```diff
@@ -47,8 +55,15 @@ void Connection::handle_write_() {
...
 void Connection::handle_event_(short what) {
+    logger->debug("connection: got bufferevent event: %d", what);
 
     if (what & BEV_EVENT_EOF) {
+        auto input = bufferevent_get_input(bev);
+        if (evbuffer_get_length(input) > 0) {
+            logger->debug("Suppress EOF with data lingering in input buffer");
+            suppressed_eof = true;
+            return;
+        }
```
bassosimone: Another error that occurred with fsrn.org: EOF was delivered before "data" with SSL. I fixed this by suppressing EOF when there is buffered data and delivering it after read.

@hellais should I further narrow the check to make sure that we're dealing with a SSL bufferevent by checking whether there is an underlying (i.e. a socket) bufferevent?

src/net/emitter.hpp
```diff
@@ -15,24 +15,42 @@ class Emitter : public Transport {
   public:
     void emit_connect() override {
         logger->log(MK_LOG_DEBUG2, "emitter: emit 'connect' event");
+        if (!do_connect) {
+            logger->log(MK_LOG_DEBUG2, "emitter: no handler set; ignoring");
+            return;
+        }
```
bassosimone: This set of changes is again to increase robustness. This is to make sure that, if we have set callbacks to `nullptr` thus saying *we are not interested*, we do not receive unwanted notifications causing exceptions.

src/ooni/web_connectivity.cpp
```diff
@@ -0,0 +1,677 @@
...
+  "x-xss-protection",
+  "content-language",
+  "cf-ray",
+  "strict-transport-security",
+  "link",
+  "x-varnish"
+};
+
+
+static const std::map<std::string, std::string> COMMON_CLIENT_HEADERS = {
```
bassosimone: Here you'd better use `http::Headers`, which soon will implement case insensitive search.

**TODO for me**: implement case insensitive search.

src/ooni/web_connectivity.cpp
```diff
@@ -0,0 +1,677 @@
...
+
+static void experiment_tcp_connect(Var<Entry> entry,
+        SocketList sockets, Callback<Error> cb,
+        Var<Reactor> reactor, Var<Logger> logger) {
+
+  int socket_count = sockets.size();
+  Var<int> sockets_tested(new int(0));
+  // XXX this is very ghetto
+  if (socket_count == 0) {
+    cb(NoError());
```
bassosimone: Here, shouldn't you return after calling the callback?

src/ooni/web_connectivity.cpp
```diff
@@ -0,0 +1,677 @@
...
+        if (close_txp == true) {
+          txp->close([=]{
+            cb(NoError());
+          });
+        } else {
+          cb(NoError());
+        }
+      } else {
+        if (close_txp == true) {
+          // XXX optimistic closure
```
bassosimone: What do you mean here?

src/ooni/web_connectivity.cpp
```diff
@@ -0,0 +1,677 @@
...
+    ErrorOr<http::Url> url =
+      mk::http::parse_url_noexcept(input);
+
+    if (!url) {
+      logger->debug("Invalid test url.");
+      callback(entry);
+      return;
+    }
+
+    // XXX check if this hostname is actually a hostname and not an IP
```
bassosimone: To do this, you may want to call `inet_pton` for `AF_INET` and `AF_INET6`; DNS code and IIRC also net/connect code do that... perhaps we should export a function to reuse for that.

src/ooni/web_connectivity.cpp
```diff
@@ -0,0 +1,677 @@
...
+    std::string hostname = url->address;
+    // XXX we probably want to do like ooni and use the system resolver
```
bassosimone: From C++ we don't know the system resolver on mobile devices. Currently Apps must pass us a proper `"dns/nameserver"` pointing to the system resolver found using platform specific methods.

One alternative would be to write a DNS engine that uses `getaddrinfo()` rather than calling libevent's DNS; in such case, the response would come from libc and hence from the system's DNS.

include/measurement_kit/http/http.hpp
```diff
@@ -98,7 +108,11 @@ void request_connect(Settings, Callback<Error, Var<net::Transport>>,
                      Var<Logger> = Logger::global());
 
 void request_send(Var<net::Transport>, Settings, Headers, std::string,
-                  Callback<Error>);
+                  Callback<Error, Var<Request>>);
```
bassosimone: Ah, I didn't notice that the API changed. So, I'll crank major as part of this PR.

bassosimone:
Okay, I fixed fsrn.org but [circle-ci failed with this regression](https://circleci.com/gh/measurement-kit/measurement-kit/473):

```
==15607== Memcheck, a memory error detector
==15607== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==15607== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==15607== Command: test/http/request
==15607== 
==15607== Jump to the invalid address stated on the next line
==15607==    at 0x10041C140: ???
==15607==  Address 0x10041c140 is not stack'd, malloc'd or (recently) free'd
==15607== 
{
   <insert_a_suppression_name_here>
   Memcheck:Jump
   obj:*
}

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
request is a Catch v1.5.0 host application.
Run with -? for options

http::request_connect_impl() works for normal connections
test/http/request.cpp:321

test/http/request.cpp:321: FAILED:
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal

test cases: 13 | 12 passed | 1 failed
assertions: 24 | 23 passed | 1 failed

==15607== 
==15607== HEAP SUMMARY:
==15607==     in use at exit: 203,333 bytes in 3,603 blocks
==15607==   total heap usage: 22,931 allocs, 19,328 frees, 36,617,090 bytes allocated
==15607== 
==15607== 184 (152 direct, 32 indirect) bytes in 1 blocks are definitely lost in loss record 694 of 761
==15607==    at 0x4C2FB55: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==15607==    by 0x5484358: event_base_once (in /usr/lib/x86_64-linux-gnu/libevent-2.0.so.5.1.9)
==15607==    by 0x4BE4F6: void mk::Poller::call_later_impl<&event_base_once>(double, std::function<void ()>) (poller.hpp:82)
==15607==    by 0x4BE2FD: mk::Poller::call_later(double, std::function<void ()>) (poller.cpp:31)
==15607==    by 0x4BE230: mk::Poller::call_soon(std::function<void ()>) (poller.cpp:27)
==15607==    by 0x4BEBE8: mk::Poller::loop_with_initial_event(std::function<void ()>) (poller.hpp:92)
==15607==    by 0x46AE6E: mk::loop_with_initial_event(std::function<void ()>) (reactor.hpp:44)
==15607==    by 0x46AC63: mk::loop_with_initial_event_and_connectivity(std::function<void ()>) (reactor.cpp:15)
==15607==    by 0x41C13F: ____C_A_T_C_H____T_E_S_T____321() (request.cpp:322)
==15607==    by 0x43E300: invoke (catch.hpp:7475)
==15607==    by 0x43E300: Catch::RunContext::invokeActiveTestCase() (catch.hpp:6132)
==15607==    by 0x43DD76: Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) (catch.hpp:6103)
==15607==    by 0x4314F4: Catch::RunContext::runTest(Catch::TestCase const&) (catch.hpp:5923)
==15607== 
{
   <insert_a_suppression_name_here>
   Memcheck:Leak
   match-leak-kinds: definite
   fun:calloc
   fun:event_base_once
   fun:_ZN2mk6Poller15call_later_implIXadL_Z15event_base_onceEEEEvdSt8functionIFvvEE
   fun:_ZN2mk6Poller10call_laterEdSt8functionIFvvEE
   fun:_ZN2mk6Poller9call_soonESt8functionIFvvEE
   fun:_ZN2mk6Poller23loop_with_initial_eventESt8functionIFvvEE
   fun:_ZN2mk23loop_with_initial_eventESt8functionIFvvEE
   fun:_ZN2mk40loop_with_initial_event_and_connectivityESt8functionIFvvEE
   fun:_ZL31____C_A_T_C_H____T_E_S_T____321v
   fun:invoke
   fun:_ZN5Catch10RunContext20invokeActiveTestCaseEv
   fun:_ZN5Catch10RunContext14runCurrentTestERNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES7_
   fun:_ZN5Catch10RunContext7runTestERKNS_8TestCaseE
}
==15607== LEAK SUMMARY:
==15607==    definitely lost: 152 bytes in 1 blocks
==15607==    indirectly lost: 32 bytes in 1 blocks
==15607==      possibly lost: 0 bytes in 0 blocks
==15607==    still reachable: 203,149 bytes in 3,601 blocks
==15607==         suppressed: 0 bytes in 0 blocks
==15607== Reachable blocks (those to which a pointer was found) are not shown.
==15607== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==15607== 
==15607== For counts of detected and suppressed errors, rerun with: -v
==15607== ERROR SUMMARY: 2 errors from 2 contexts (suppressed: 0 from 0)
```

Will investigate this tomorrow

include/measurement_kit/report/entry.hpp
```diff
@@ -13,7 +13,7 @@ namespace report {
 // A report entry. This is implemented using private inheritance from
 // nlohmann/json such that we are not exposing externally we are using
 // such json library to implement this class.
-class Entry : private nlohmann::json {
+class Entry : public nlohmann::json {
```
hellais: By this you mean to stop using `Entry` at all and just use `nlohmann::json` or to open it up as public as done here?

src/net/connection.cpp
hellais: I think it's ok to do this unconditionally as there may be some other edge cases we haven't yet found when the connection is closed in an unclean way. In general this logic seems like something we may always want to have, that is if we receive and EOF and there is still some data in the buffer we should process it before the EOF is triggered.

src/ooni/web_connectivity.cpp
hellais: Yes, that is a good idea. Actually we may even want to factor these constants out into some shared header. Where do you think it makes sense to put them: `ooni/constants.hpp`?

src/ooni/web_connectivity.cpp
hellais: It actually doesn't matter since the other things are noop when socket_count is 0, but I agree this should return to avoid breakage in the future.

Will change.

src/ooni/web_connectivity.cpp
hellais: What I mean is that I don't actually wait for the connection to be closed, but call close on it and expect everything to go well.

I guess since closing a socket doesn't require bidirectional communication there shouldn't be much that can go wrong, though I wanted to take note of the fact that I am not handling any errors or waiting for the callback to actually fire.

src/ooni/web_connectivity.cpp
hellais: Right. Will do.

src/ooni/web_connectivity.cpp
hellais: I see. I think the option of using `getaddrinfo()` is suboptimal given the fact that we will want in the future to support more than just `A` lookups. For the time being I guess we can consider this acceptable, but we should make it clear to the user of the library that in order for this test to be accurate the system resolver needs to be set and perhaps provide examples on obtaining it on the various platforms.

include/measurement_kit/report/entry.hpp
bassosimone: To open it as public and perhaps to declare `Entry` as alias for `nlohmann::json`.

src/ooni/web_connectivity.cpp
bassosimone: >  I think the option of using getaddrinfo() is suboptimal given the fact that we will want in the future to support more than just A lookups.

Yes; to clarify, my suggestion was not to drop what we currently have but to enable to use our current dns interface to query using getaddrinfo, if the user wants so; e.g.:

```
dns::query("IN", "A", "www.google.com", [=](Error err, dns::Response response) {
    // ...
}, {
    {"dns/driver", "getaddrinfo"},
});
```

bassosimone:
@hellais I have investigated the above regression, and now I fully understand why reverting f46211efd069705673fa677f53109e039fee86a6 fixed the regression.

The culprit is the following, partial hunk:

```diff
-    // TODO: we should improve the parser such that the transport forwards the
-    // "error" event to it and then the parser does the right thing, so that the
-    // code becomes less "twisted" here.
-
     parser->on_end([=]() {
-        if (*prevent_emit == true) {
-            return;
-        }
-        txp->emit_error(NoError());
+        txp->on_error(nullptr);
+        txp->on_data(nullptr);
+        reactor->call_soon([=]() {
+            logger->log(MK_LOG_DEBUG2, "request_recv_response: end of closure");
+            cb(NoError(), response);
+        });
     });
     txp->on_error([=](Error err) {
         logger->debug("Received error %d on connection", err.code);
         if (err == EofError()) {
-            // Calling parser->on_eof() could trigger parser->on_end() and
-            // we don't want this function to call ->emit_error()
-            *prevent_emit = true;
             try {
                 logger->debug("Now passing EOF to parser");
                 parser->eof();
```

Before applying the diff, `on_end()` did its job through `on_error()`. Now, the `=` capture only captures the variables that are used and specifically `parser` is used by `on_error()`.

After the diff, `on_end()` does not call `on_error()` and does not use `parser` which is therefore not kept safe by the lambda; hence, the `parser` &mdash; which is what calls its `on_end()` &mdash; method is destroyed in the new code by setting `txp` callbacks to `nullptr` and then, when the stack unwinds, we operate on code that was freed.

I'll think of a way to remove this issue from the code; for now reverting the diff solved the problem and thus I think we can continue to test this branch.

Let me know if the code in here satisfies the requirements of running a full web connectivity test and, if so, I shall enter into the final stage of code review and prepare for merging.

include/measurement_kit/http/http.hpp
bassosimone: Done!

bassosimone:
Note: merged `master` in `3d2da3c` and handled there the conflict.

hellais:
I have tested this branch with the latest changes and it works like a charm!

I will integrate the minor fixes you suggested and implement some of the notes I had for things to do on my side and then I think it's good to go!

hellais:
This is ready for a final review now.

bassosimone:
@hellais thank you, I will begin reviewing the code and report back!

Also merging with branch right now to see whether the breakage of unit tests is transient or permanent (I suspect one of the errors still has unfortunately to do with SSL and premature EOF).

hellais:
Tests all green now!

src/ooni/web_connectivity.cpp
bassosimone: ack!

include/measurement_kit/report/entry.hpp
bassosimone: @hellais if you don't have objections here I'd say "yolo" and "let's expose nlohmann::json" publicly

src/ooni/constants.hpp
```diff
@@ -0,0 +1,63 @@
...
+  "age",
+  "via",
+  "p3p",
+  "x-xss-protection",
+  "content-language",
+  "cf-ray",
+  "strict-transport-security",
+  "link",
+  "x-varnish"
+};
```
bassosimone: *TODO* I think code should be added so to have case insensitive headers!

src/ooni/utils.cpp
```diff
@@ -13,75 +18,139 @@ void ip_lookup(Callback<Error, std::string> callback, Settings settings,
...
-        // Set sane values such that we should not have errors accessing
-        // output even when lookup may fail for current database
-        {"country_code", "ZZ"},
-        {"country_name", "ZZ"},
-        {"asn", "AS0"},
-    };
-    Error err = geoip_resolve_country (ip, path_country, node);
-    if (err) {
-        return err;
+    IPLocation ip_location(path_country, path_asn);
```
bassosimone: The above line means that we're creating instances of GeoIP databases every time we need to lookup a new entry, but my understanding was that the refactoring purpose was to avoid that, correct?

src/ooni/utils.cpp
```diff
@@ -13,75 +18,139 @@ void ip_lookup(Callback<Error, std::string> callback, Settings settings,
...
+    std::string asn = res;
+    asn = split(asn).front(); // We only want ASXX
     free(res);
-    GeoIP_delete(gi);
-    return NoError();
+    return asn;
 }
 
 ErrorOr<json> geoip(std::string ip, std::string path_country,
                     std::string path_asn) {
```
bassosimone: Having a `json` as return value looked like a good idea originally, but now I'm not so sure about this point. I think a better way of dealing with this would be that we have a class wrapping the geoip databases and that we have methods that allow to query each separate aspect (i.e. country) and return `ErrorOr<std::string>`.

src/ooni/utils.cpp
```diff
@@ -13,75 +18,139 @@ void ip_lookup(Callback<Error, std::string> callback, Settings settings,
...
-    if (gi == nullptr) {
+void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
+                     Var<Reactor> reactor, Var<Logger> logger) {
+    resolver_lookup_impl(callback, settings, reactor, logger);
+}
+
+IPLocation::IPLocation(std::string path_country, std::string path_asn) {
+    if (path_asn != "") {
+        gi_asn = GeoIP_open(path_asn.c_str(), GEOIP_MEMORY_CACHE);
+    }
```
bassosimone: I'd prefer it if we could only cache the paths in the constructor and then lazily open the databases when requested to do so by access APIs. This would allow us to return the proper `ErrorOr<>` error in case it was not possible to open a database, and would also put open and check-after-open in the same place, which I think makes the code more robust to refactoring and easier to defend.

src/ooni/utils.hpp
```diff
@@ -20,9 +20,39 @@ void ip_lookup(Callback<Error, std::string> callback, Settings settings = {},
...
+      ErrorOr<std::string> resolve_country_code(std::string ip);
+
+      ErrorOr<std::string> resolve_country_name(std::string ip);
+
+      ErrorOr<std::string> resolve_asn(std::string ip);
+
+    private:
+      GeoIP *gi_asn = nullptr;
+      GeoIP *gi_country = nullptr;
+      Var<Logger> logger = Logger::make();
```
bassosimone: I think it would be interesting to log failures in the context of a specific test; as such, I think the `logger` should not be a class member, rather it should be passed to each of the `resolve_xx` methods as the last optional argument otherwise initialized to the default logger.

src/ooni/web_connectivity.cpp
```diff
@@ -0,0 +1,612 @@
...
+    };
+  };
+
+  for (auto socket : sockets) {
+      std::string address = socket.first;
+      int port = socket.second;
+      Settings connect_options;
+      connect_options["host"] = address;
+      connect_options["port"] = port;
+      connect_options["net/timeout"] = 0.2;
```
bassosimone: Isn't this timeout too small? Especially on mobile and eastern countries, I think 0.2 seconds of RTT can happen depending on where one is and on the cell coverage.

src/report/entry.cpp
```diff
@@ -25,21 +25,6 @@ namespace report {
...
-Entry &Entry::operator[](std::string key) {
-    // Note: out_of_range could not happen because operator[] returns
-    // a null json object if the accessed key is missing
-    try {
-        return static_cast<Entry &>(nlohmann::json::operator[](key));
-    } catch (std::domain_error &) {
-        throw JsonDomainError();
-    }
-}
-
```
bassosimone: *TODO* for self: understand the implication of this hunk &mdash; I have forgotten!

test/ooni/utils.cpp
```diff
@@ -28,3 +29,31 @@ TEST_CASE("geoip works") {
     REQUIRE(((*json)["country_code"] == std::string{"US"}));
     REQUIRE(((*json)["country_name"] == std::string{"United States"}));
 }
+
+TEST_CASE("is_ip_addr works on ipv4") {
+    REQUIRE(mk::ooni::is_ip_addr("127.0.0.1") == true);
+}
+
+TEST_CASE("is_ip_addr works on ipv6") {
+    REQUIRE(mk::ooni::is_ip_addr("::42") == true);
```
bassosimone: :-)

src/ooni/utils.cpp
hellais: What you say is true on both points. The main place where I wanted to avoid though creating a bunch of GeoIP databases was inside of the web_connectivity test (https://github.com/measurement-kit/measurement-kit/pull/690/files/0d68f3f5613e4c5c98b25632b9eb6d5a463d3b69#diff-4d644e144323880fdd210d881591233dR192), but I didn't want to drastically change the rest of the code (not web connectivity related) to be less wasteful.

I think it's probably a good idea to flag this in the points where it's called that the instance of `IPLocation` should be cached, but I would implement this in a separate PR.

src/ooni/utils.cpp
hellais: Yeah to be honest I also think it's a bit weird to have this return a JSON structure when you actually are only returning 3 values. Also we don't really do anything with the country_name so that could potentially even be dropped leaving only the country code as a useful value.

I would also take note of this and suggest we do it as part of another PR.

src/ooni/web_connectivity.cpp
hellais: Oh yes indeed this timeout is definitely too low.

On ooniprobe we are even using a timeout of 30 seconds here (that I think may even be too much, but isn't too much of an issue there due to it's concurrency).

I think changing this to 10 seconds is appropriate.

src/report/entry.cpp
hellais: The reason for doing this is that now that we made `nlohmann::json` public it's no longer necessary to manually expose the methods that we need to support. Having these two methods in place lead to, in particular, the issue that it was not possible to access keys that were not of string type (for example to access the index of an array).

include/measurement_kit/report/entry.hpp
hellais: 187hby

src/ooni/utils.cpp
hellais: Yes I think this is a very valid point.

src/http/request.cpp
bassosimone: Ticketized as #734

src/http/request.cpp
bassosimone: Actually this change was later reverted because it introduced further issues.

src/ooni/web_connectivity.cpp
bassosimone: For the records, `return` changed, thread of comments can be considered folded.

src/ooni/constants.hpp
bassosimone: Ticketized as #735.

src/ooni/utils.cpp
bassosimone: Ticketized as #736

src/ooni/utils.cpp
bassosimone: Ticketized as #737

src/ooni/utils.cpp
bassosimone: Ticketized as #738

src/ooni/utils.hpp
bassosimone: Ticketized as #739

src/ooni/web_connectivity.cpp
bassosimone: Done in 825296bde87f0

bassosimone:
- [ ] read one more time the whole diff
- [ ] decide whether we need more regress tests
- [ ] test locally
- [ ] merge

src/http/response_parser.hpp
bassosimone: Ticketized as #741

src/ooni/web_connectivity.cpp
```diff
@@ -0,0 +1,612 @@
...
+
+  // Check if the headers match
+  logger->debug("web_connectivity: checking headers");
+  std::set<std::string> lowercase_ctrl_headers;
+  std::set<std::string> lowercase_exp_headers;
+
+  for (Entry::iterator it = control["headers"].begin();
+       it != control["headers"].end(); ++it) {
+    //lowercase_ctrl_headers.insert(std::tolower(it.key()));
+    lowercase_ctrl_headers.insert(it.key());
```
bassosimone: @hellais here the idea is to convert all the key as tolower, right?

bassosimone: Confirmed! Ticketized as #743

bassosimone:
Well, then, let's merge this super important bit of work and continue its refinement in tree by addressing all the sub issues that have been opened! 

giphy

hellais:
party-time

PR #691, opened 2016-06-24
Feature/http redirect

bassosimone:


hellais:
Note: I have integrated this patch as part of the web_connectivity branch. Should we perhaps consider merging it all together as part of the merging of `web_connectivity`?

bassosimone:
> Should we perhaps consider merging it all together as part of the merging of `web_connectivity`?

@hellais yes, that makes sense!

PR #697, opened 2016-06-30
Add easy interface and use it from Python

bassosimone:
Proof of concept of async execution of OONI tests from Python.

cc: @hellais @willscott

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -3,7 +3,7 @@
 set -e
 export LC_ALL=C  # Stable sorting regardless of the locale
 
-no_download=
+no_download=""
 if [ "$1" = "-n" ]; then
     no_download=1
 fi
@@ -81,7 +81,7 @@ gen_executables() {
 }
 
 get_repo() {
-    if [ "$no_download" == "1" ]; then
+    if [ "$no_download" = "1" ]; then
         return
     fi
     echo ""
--- a/include/measurement_kit/common.hpp
+++ b/include/measurement_kit/common.hpp
@@ -4,6 +4,7 @@
 #include <measurement_kit/common/callback.hpp>
 #include <measurement_kit/common/continuation.hpp>
 #include <measurement_kit/common/delegate.hpp>
+#include <measurement_kit/common/easy.h>
 #include <measurement_kit/common/error.hpp>
 #include <measurement_kit/common/error_or.hpp>
 #include <measurement_kit/common/lexical_cast.hpp>
--- a/include/measurement_kit/common/easy.h
+++ b/include/measurement_kit/common/easy.h
@@ -1,0 +1,44 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+#ifndef MEASUREMENT_KIT_COMMON_EASY_H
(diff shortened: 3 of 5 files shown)


willscott:
This is great!

I'm not convinced the name 'Easy' is the descriptor that we want. Something more descriptive that lets people understand they're dealing with the 'measurement kit python binding' seems possible?

The structure itself seems fine.

bassosimone:
> I'm not convinced the name 'Easy' is the descriptor that we want.

Yes, perhaps it's not the best choice ever. (FTR, I chose it for similarity with `curl` where the descriptor you manipulate for running stuff is called `curl_easy_t` or something like that.)

> Something more descriptive that lets people understand they're dealing with the 'measurement kit python binding' seems possible?

What do you suggest?

To further elaborate on my objective here,`easy.h` should be a simplified C API for driving MK tests from languages that support foreign-function-interface (FFI) like functionality. That is, with this simplified C API in MK, one should probably be able to write, say, `Ruby` bindings quite easily by writing a Ruby glue around hooks for calling the "easy" functions via FFI. Does this concept make sense?

bassosimone:
I also discussed this pull request with @hellais in chat. Trying to summarize what we said (please, @hellais, correct me if I am wrong). It would be great to have a Python extension that users can `pip install` rather than a Python file distributed along MK. It would not be practical, however, to compile MK directly using `setup.py`, thus it would be acceptable to require users to install MK in order to correctly use the Python extension. The Python extension should live in another repository.

Now, as my personal follow up, I believe that the Python extension could also be implemented using `ctypes` rather than in C; i.e., the most important thing is to have something you can `pip install`. Is that correct? If so, I'd like to advocate for this solution, because I see some advantages of having a FFI like interface (easy to write bindings for other languages) and in using `ctypes` (namely that the cost of having an extension for both Python2 and Python3 is zero). Does it make sense?

willscott:
I feel like easy isn't a super descriptive name, and I don't think there's a more complex `hard` API that it is in contrast to.

Part of me feels like what we're talking about here is the `mk_api`?

bassosimone:
Thanks @willscott! I will think a bit more about `mk_api` as the optimal name; in a certain sense all the headers are the API of MK and that is just a subset for a specific purpose?

I've tried to follow @hellais's suggestion to see what would be the complexity of writing an extension in C for the Python module and made some interesting progresses here:

- https://github.com/bassosimone/measurement-kit_python

Working on this, I also created a repository to make `.deb` packages and a `tap` for MK, because I believe the use case is that you install MK using a package manager and then you `setup.py`:

- https://github.com/bassosimone/measurement-kit_debian
- https://github.com/bassosimone/homebrew-measurement-kit

hellais:
Doing a bit of research into this I came across this: https://github.com/pybind/pybind11.

It seems like it's something worthwhile looking into to make the python bindings and avoid using `Python.h` and handling reference counting by hand.

It's API also seems very intuitive and easy to use. Probably something worth playing around with.

bassosimone:
I am going to close this pull request, which is deprecated in favor of [bassosimone/measurement-kit-python](https://github.com/bassosimone/measurement-kit-python) where I have ported the Python code developed in here along with a pure C extension working like what was done using `ctypes` as part of this pull request. I am also going to transfer the ownership of the repository to measurement-kit, so that more people could easily contribute. Thanks for all the input!

PR #700, opened 2016-07-01
Add lua engine

bassosimone:


bassosimone:
This was a nice example. But interacting at low level with lua API is too complex. I think, if we're even going to use a scripting language like lua, we'll be using the approach in #822.

PR #711, opened 2016-07-10
Fix TLS timeout bug

bassosimone:
When the HTTPS request is sent, the write timeout is never actually removed and, when it expires, we receive a TimeoutError error. We observed this behavior (and we were not able to explain it) when we were debugging the web_connectivity branch. I guess this diff should be applied to the web_connectivity branch to see whether it is enough to repair the test or more is needed (I think so).

The fix is not particularly clever. Timeouts seems to work much better for the socket layer than for the SSL layer in bufferevents. So, rather than setting the timeout at high level, set it at low level. When a timeout would eventually expire at socket layer, it would be passed upstream (added test for this).


doc/api/net/connect.md
```diff
@@ -58,8 +58,7 @@ modified using the following `settings`:
 
 - *"net/ssl"* (bool): whether to establish a SSL connection (default: false).
 
-- *"net/timeout"* (double): timeout for connect and I/O operations (default: `5.0` for
-  connecting and `30.0` for the established connection).
+- *"net/timeout"* (double): timeout for connect and I/O operations (default: `5.0` seconds).
```
bassosimone: Here I noticed documentation was not up to date and fixed it

bassosimone:
Setting as on hold because unfortunately this pull request does not seem to be enough

bassosimone:
Basically, this approach is not viable because now we incur into the risk of receiving read timeouts. In fact, now the code sometimes fails due to this reason. Damn.

bassosimone:
Reopening because I think I sorted it out. Will force push.

hellais:
I have tested this with the web-connectivity branch that includes redirect support and the bug still seems to be present.

The reason for the failure is, as we were seeing previously:

```
libc++abi.dylib: terminating with uncaught exception of type std::__1::bad_function_call: std::exception
```

This means that there is still the problem of the errors being triggered twice.

The minimal way of reproducing the issue is by running the http/request example like so:

```
./example/http/request -vvvR 2 http://fsrn.org
```

bassosimone:
@hellais okay, this probably means there was *one more* bug, because this bug here was clearly in place as showed by the simpler example. Will test this in the context of `web_connectivity` branch to see if I can better understand what this additional problem is.

hellais:
Although this is not the only fix required for getting #690 to work properly, it is still an important fix, so I am going to merge this into master.

PR #716, opened 2016-07-12
ndt as ooni test

bassosimone:
This has been running on a experimental app for quite some time now. The diff is simple enough that I think this can qualify as hotfix.

Since this changes a public header, it leads us to `v0.3.0`-land.

### DIFF
--- a/include/measurement_kit/ooni/collector_client.hpp
+++ b/include/measurement_kit/ooni/collector_client.hpp
@@ -15,7 +15,7 @@ namespace collector {
     To submit a file report, pass it to `submit_report()`:
 */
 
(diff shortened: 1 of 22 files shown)


include/measurement_kit/ooni/collector_client.hpp
```diff
@@ -15,7 +15,7 @@ namespace collector {
     To submit a file report, pass it to `submit_report()`:
 */
 
-#define MK_OONI_DEFAULT_COLLECTOR_URL "https://a.collector.test.ooni.io"
+#define MK_OONI_DEFAULT_COLLECTOR_URL "https://a.collector.ooni.io"
```
bassosimone: cc: @hellais

PR #725, opened 2016-07-21
Copy neubot's implementation of percentile

bassosimone:
This copies Neubot's implementation of percentile. I have verified using python3's median() and Google Docs that the numbers in the regress tests are correct.

It would be nice to have a double check on the formula and on the numbers, tho.

This patch will soon be used by NDT code. The idea is to report to the user specific percentile speeds such that it is possible to roughly estimate connection's quality.

### DIFF
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -284,7 +284,7 @@ void dump_settings(Settings &s, std::string prefix, Var<Logger> logger) {
     logger->debug("%s: }", prefix.c_str());
 }
 
-// Adpated from <http://code.activestate.com/recipes/511478/>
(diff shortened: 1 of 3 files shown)


src/common/utils.cpp
```diff
@@ -284,4 +284,21 @@ void dump_settings(Settings &s, std::string prefix, Var<Logger> logger) {
     logger->debug("%s: }", prefix.c_str());
 }
 
+// Adpated from <http://code.activestate.com/recipes/511478/>
```
AntonioLangiu: typo

bassosimone: fixed, thank you!

PR #728, opened 2016-07-21
Allow logging also on log file

### LINKED ISSUE #706: Logger: allow to write logs onto a file
Something like `logger->set_logfile("test.log")`. I think it should be valid to both specify a logfile and to receive logs using `logger->on_log()` because I see the use case where you want to see logs as they are emitted and at the same time to have the logs saved.

Discussed with @lorenzoPrimi, required by NetProbe

bassosimone:
This logging is orthogonal from logging specified through the
`on_log()` method such that both can be employed.

Closes #706.

- - - 

While there, also make sure logger's documentation is up to date.

### DIFF
--- a/doc/api/common/logger.md
+++ b/doc/api/common/logger.md
@@ -177,6 +177,11 @@ int main() {
 }
 ```
 
+# BUGS
+
+If the logfile could not be open, or written, the error is silently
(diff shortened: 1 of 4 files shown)


bassosimone:
Looking more closely, this can be merged as a hotfix, given its simplicity.

PR #732, opened 2016-07-30
Use OONI's default collector

### LINKED ISSUE #717: Using default OONI collector
We should use the default OONI collector. I briefly enabled it during #716 but it was not working out of the box (guess server-side configuration?) and hence postponed its adoption.

bassosimone:
Must have this in to release v0.2.6.

Closes #717.

### DIFF
--- a/include/measurement_kit/ooni/collector_client.hpp
+++ b/include/measurement_kit/ooni/collector_client.hpp
@@ -15,7 +15,7 @@ namespace collector {
     To submit a file report, pass it to `submit_report()`:
 */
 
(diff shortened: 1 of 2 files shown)


hellais:
So it seems like this is now leading to travis measurements being submitted to the OONI canonical collector.

https://measurements.ooni.torproject.org/2016-08-01/20160731T122014Z-US-AS14618-dns_injection-no_report_id-0.2.0-probe.json

If we want to run end-to-end tests we should use the testing one.

bassosimone:
Ack, will submit a patch for this by end of today!

PR #751, opened 2016-08-01
Make sure we copy on_entry()

bassosimone:
For now this patch will do but we'd better properly use copy
construction to avoid losing pieces when we copy.

Also, it would probably make sense to second think the current
approach where we schedule a Var copy of the original object to
see whether there is a better approach for doing the same.


bassosimone:
Created #752 to remind to really fix the issue rather than just providing a temporary patch as I am doing through this hotfix pull request.

Issue #738, opened 2016-07-31
geoip: lazily open databases


bassosimone:
Quoting myself in #690:

> I'd prefer it if we could only cache the paths in the constructor and then lazily open the databases when requested to do so by access APIs. This would allow us to return the proper ErrorOr<> error in case it was not possible to open a database, and would also put open and check-after-open in the same place, which I think makes the code more robust to refactoring and easier to defend.

bassosimone:
Implemented in #759

Issue #737, opened 2016-07-31
geoip: remove API returning a JSON


bassosimone:
Quoting myself in #690:

>> Having a json as return value looked like a good idea originally, but now I'm not so sure about this point. I think a better way of dealing with this would be that we have a class wrapping the geoip databases and that we have methods that allow to query each separate aspect (i.e. country) and return ErrorOr<std::string>.

and @hellais:

> Yeah to be honest I also think it's a bit weird to have this return a JSON structure when you actually are only returning 3 values. Also we don't really do anything with the country_name so that could potentially even be dropped leaving only the country code as a useful value.
> 
> I would also take note of this and suggest we do it as part of another PR.

Issue #736, opened 2016-07-31
geoip: memoize previously opened dbs to allow reuse


bassosimone:

AntonioLangiu:
If the meaning of this issue was to be able to handle multiple queries with an open database I think it is done in #759 (if not, @bassosimone can you give me more details about it)

bassosimone:
Uhm, yes and no. The main problem, I guess, is that I incorrectly worded the issue. I agree with you that the class already allows to reuse existing connections. However, in practice, the current code instances the class and destroys it every time. This is not totally incorrect, because each test could in principle receive different paths for the databases. What I'd like to see is that, if we already did open a database for a specific set of paths, we re-use it. Otherwise, we close the previously opened database and open a new one. A sort of one-off memoization.

Is the issue clearer now?

bassosimone:
This has now been merged into master.

PR #759, opened 2016-08-02
open database just when needed

AntonioLangiu:
This PR is related to #738 
I did the following changes:

1. Add a specific error to handle an invalid db. 
2. The constructor will just save the path and nothing more
3. Move the db opening phase to the functions that needs it and leave it open until the destructor is called.
This makes us able to do multiple query to the same db without having to reopen it.
4. Adapt the function `geoip` to return the correct error. I will remove it in future PR  according to #737

I think this will also complete #736 if I interpreted it right.

### DIFF
--- a/include/measurement_kit/ooni/error.hpp
+++ b/include/measurement_kit/ooni/error.hpp
@@ -17,6 +17,7 @@ MK_DEFINE_ERR(MK_ERR_OONI(4), MissingMandatoryKeyError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(5), InvalidMandatoryValueError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(6), MissingRequiredHostError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(7), MissingRequiredUrlError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(8), InvalidGeoIPDatabase, "")
 
 } // namespace mk
 } // namespace ooni
--- a/include/measurement_kit/ooni/error.hpp
+++ b/include/measurement_kit/ooni/error.hpp
@@ -17,7 +17,8 @@ MK_DEFINE_ERR(MK_ERR_OONI(4), MissingMandatoryKeyError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(5), InvalidMandatoryValueError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(6), MissingRequiredHostError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(7), MissingRequiredUrlError, "")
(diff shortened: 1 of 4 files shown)


include/measurement_kit/ooni/error.hpp
```diff
@@ -17,6 +17,7 @@ MK_DEFINE_ERR(MK_ERR_OONI(4), MissingMandatoryKeyError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(5), InvalidMandatoryValueError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(6), MissingRequiredHostError, "")
 MK_DEFINE_ERR(MK_ERR_OONI(7), MissingRequiredUrlError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(8), InvalidGeoIPDatabase, "")
```
bassosimone: Could this be `InvalidGeoIpDatabase`? I am old and find it difficult to split words when acronyms and CamelCase are mixed, so I prefer when just one word is capitalized in a acronym.

src/ooni/utils.cpp
```diff
@@ -24,12 +24,8 @@ void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
...
 
 IPLocation::IPLocation(std::string path_country, std::string path_asn) {
-    if (path_asn != "") {
-        gi_asn = GeoIP_open(path_asn.c_str(), GEOIP_MEMORY_CACHE);
-    }
-    if (path_country != "") {
-        gi_country = GeoIP_open(path_country.c_str(), GEOIP_MEMORY_CACHE);
-    }
+    _path_asn = path_asn;
+    _path_country = path_country;
```
bassosimone: Could you use, instead, `path_asn_`? The leading underscore is a Pythonism, but in C++ it is more common either to put a trailing underscore or to prefix `m_` to the variable. (As an aside, I think it would be much better to do `this->variable`, but this seems to be a strange thing to do in C++, so not using it.)

bassosimone: (Or, given that no underscore is currently used for the class attributes, just don't use any underscore.)

test/ooni/utils.cpp
```diff
@@ -30,6 +30,12 @@ TEST_CASE("geoip works") {
     REQUIRE(((*json)["country_name"] == std::string{"United States"}));
 }
 
+TEST_CASE("geoip can handle an invalid database") {
```
bassosimone: As above, it's not invalid rather you cannot open it (I think invalid is a narrower concept)

bassosimone:
Thank you! This pull request looks good to me! Will be happy to merge once the small mainly-cosmetic changes that I requested have been applied.

bassosimone:
Thank you, merging!

PR #765, opened 2016-08-05
Feature/incremental ooni collector

bassosimone:


bassosimone:
Keeping this on hold for a while more: will see whether bits of it could be submitted as hotfix pull requests to simplify the task of reviewing it all.

bassosimone:
@hellais fyi, this is the most high-priority MK pull request requiring review in order to move forward with the release of MK v0.3.0.

src/libmeasurement_kit/ooni/ooni_reporter.cpp
```diff
@@ -0,0 +1,89 @@
...
+    reactor = parent.reactor;
+    if (settings.find("collector_base_url") == settings.end()) {
+        // Note: by default we use the testing collector URL because otherwise
+        // testing runs would be collected creating noise and using resources
+        settings["collector_base_url"] = collector::testing_collector_url();
+    }
+    logger->info("Using collector: %s", settings["collector_base_url"].c_str());
+}
+
+/* static */ Var<BaseReporter> OoniReporter::make(const OoniTest &parent) {
```
hellais: Why is `static` commented out? I think it makes sense to make this method static.

bassosimone: The method is declared static in the class. Here in the implementation I've prefixed a commented out `static` to remind the reader that the method is static.

It's a convention I've seen used somewhere, and thought it was useful as a reminder; now I also see that it can be confusing and I'm less sympathetic towards it than before.

src/libmeasurement_kit/report/report.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+#define FMAP mk::fmap<Var<BaseReporter>, Continuation<Error>>
+
+void Report::open(Callback<Error> callback) {
+    if (openned_) {
+        callback(ReportAlreadyOpen());
+        return;
+    }
+    openned_ = true;
```
hellais: I can think of some issues with setting this attribute before you actually call the open report function.

If for some reason one of the two report opens fail an error message will be returned to the caller however the called doesn't have the ability to re-issue the open since it's guarded by the openned_ flag.

It's probably also not wise to put it after the call, though, since that would mean that if you call `open` while the open is in progress you will still get an error.

A possible way out of this would be to move the `openned_` and `closed_` flags into the report functions and make them a state variable instead of a boolean flag.

The states could be:

* INITIAL

* OPENNED

* IN_PROGRESS

* CLOSED

Where IN_PROGRESS is used if either an open or a close is in progress.

src/libmeasurement_kit/ooni/ooni_test.cpp
```diff
@@ -46,17 +46,19 @@ void OoniTest::run_next_measurement(Callback<Error> cb) {
...
-            entry_cb(entry.dump());
-        }
-        logger->debug("net_test: written entry");
+        report.fill_entry(entry);
+        report.write_entry(entry, [=](Error error) {
+            if (error) {
+                cb(error);
+                return;
+            }
+            if (entry_cb) {
```
hellais: I think it's better to call the `entry_cb` function before you handle the error callback. Actually it may even make sense to call it before you write the entry itself. I would imagine a user of this feature may want to know the result of the measurement independently from wether or not reporting succeeded.

hellais:
This branch overall looks good to me. I made a couple of comments to some things I think should be adjusted before the merge. After that I will proceed with testing it and merging!

Good work!

👍

src/libmeasurement_kit/report/report.cpp
bassosimone: I expanded on your suggestion in 7070663b4ea0 by moving the logic to allow retry into the common layer (`BaseReporter`) and allowing to retry partially failed operations with idempotent semantic (e.g. you write an entry and the overall operation fails because the network failed but the entry was written on disk, you retry and the entry is not written twice onto the disk, but only the network sub-operation is retried).

include/measurement_kit/ooni/ooni_reporter.hpp
```diff
@@ -13,8 +13,8 @@ class OoniReporter : public BaseReporter {
   public:
     static Var<BaseReporter> make(const OoniTest &ooni_test);
 
-    Continuation<Error> open(const Report &report) override;
-    Continuation<Error> write_entry(const Entry &entry) override;
+    Continuation<Error> open(Report report) override;
+    Continuation<Error> write_entry(Entry entry) override;
```
bassosimone: Since now we have a chain of continuations, better to copy report and entry because there are many lambdas and it is not wise to pass references and then have to remember to make copies to keep safe objects. Better to change the API and take the safe side in one single place and for good.

include/measurement_kit/report/base_reporter.hpp
```diff
@@ -5,22 +5,46 @@
...
+#include <measurement_kit/report/report.hpp>
 #include <measurement_kit/report/entry.hpp>
 
 namespace mk {
 namespace report {
 
 class Report;
 
-class BaseReporter {
+class BaseReporter : public NonCopyable, public NonMovable {
```
bassosimone: Just in case

include/measurement_kit/report/error.hpp
```diff
@@ -9,12 +9,13 @@
 namespace mk {
 namespace report {
 
-MK_DEFINE_ERR(MK_ERR_REPORT(0), ReportAlreadyOpen, "")
-MK_DEFINE_ERR(MK_ERR_REPORT(1), ReportNotOpen, "")
-MK_DEFINE_ERR(MK_ERR_REPORT(2), ReportAlreadyClosed, "")
+MK_DEFINE_ERR(MK_ERR_REPORT(0), ReportAlreadyOpenError, "")
+MK_DEFINE_ERR(MK_ERR_REPORT(1), ReportNotOpenError, "")
+MK_DEFINE_ERR(MK_ERR_REPORT(2), ReportAlreadyClosedError, "")
```
bassosimone: Making sure all errors end with the `Error` suffix

src/libmeasurement_kit/report/file_reporter.cpp
```diff
@@ -26,38 +26,37 @@ static Error map_error(std::ofstream &file) {
...
+    });
 }
 
-Continuation<Error> FileReporter::write_entry(const Entry &entry) {
-    std::string s = entry.dump();
-    return [=](Callback<Error> cb) {
-        file << s << std::endl;
+Continuation<Error> FileReporter::write_entry(Entry entry) {
+    return do_write_entry_(entry, [=](Callback<Error> cb) {
+        file << entry.dump() << std::endl;
```
bassosimone: String now unnecessary since entry is now passed by copy to this function

bassosimone:
Thank you for your review, @hellais, I am thrilled to see how your comments allowed me to significantly improve the functionality proposed in this pull request. Unless travis-ci and circle-ci fail, I'd say we're good for another round of review and hopefully for merging.

src/libmeasurement_kit/report/report.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#define FMAP mk::fmap<Var<BaseReporter>, Continuation<Error>>
+
+void Report::open(Callback<Error> callback) {
+    mk::parallel(FMAP(reporters_, [=](Var<BaseReporter> r) {
+        return r->open(*this);
+    }), callback);
+}
+
+void Report::write_entry(Entry &entry, Callback<Error> callback) {
```
bassosimone: Here it should be `entry` not `&entry` to make sure `entry` is properly copied and kept safe by `[=]`.

test/report/report.cpp
```diff
@@ -0,0 +1,250 @@
...
+    Report report;
+    report.add_reporter(BaseReporter::make());
+    Entry entry;
+    report.write_entry(entry, [&](Error err) {
+        REQUIRE(err.code == ParallelOperationError().code);
+        REQUIRE(err.child_errors.size() == 1);
+        REQUIRE(err.child_errors[0]->code == ReportNotOpenError().code);
+        report.open([&](Error err) {
+            REQUIRE(!err);
+            report.write_entry(entry, [&](Error err) {
```
bassosimone: xxx: because I've added caching, this test is now actually not testing what I would like to test

bassosimone:
I did another code review, @hellais, and added two commit to increase code correctness. Now I am fully convinced that this branch is totally ready for the final review and for merging.

hellais:
Did some further review and testing of this.

Just have a couple of comments:

* Now there is no guarding against somebody calling `open()` or `close()` twice before the previous `open()` or `close()` call has reached the final stage. I think this is OK as it's a user error, but it's something worth pointing out.

* Something else to watch out for an implementor that has retry logic in place is that if they issue a retry call for `write_entry` after they have called close on any of the attached reports, the retry call will fail. I also think this is an implementation detail that can be sorted out.

That said I think this is good to be merged. Unless you have some comments on the above comments (🔁 ) I will hit it ‼️

bassosimone:
@hellais thank you for the comments! I've updated the documentation of report in 9e359146af39 to indicate the issues you pointed out (in my description there is just a single issue, actually, i.e. that you cannot overlap any of `open`, `write_entry`, or `close` &mdash; this should cover IMHO both the cases you did indicate above in your comment).

To remind us of these issues, I've also created #799.

doc/api/report/error.md
```diff
@@ -0,0 +1,73 @@
...
+- *ReportIoError*: writing the report on file yields a I/O error
+- *ReportLogicalError*: it is not possible to serialize input data into the report when writing it to file
+- *DuplicateEntrySubmitError*: the previous call to the `write_entry` of a `BaseReporter` received an entry that is exacly equal to the one passed in the current call to `write_entry`
+
+Note that the `BaseReporter` logic may hide many of these errors when it makes
+sense to do so in order to make code idempotent. For example, if you call
+`open` on a `BaseReporter` that is already open, the result would be `NoError`
+rather than `ReportAlreadyOpenError`. This is to guarantee idempotence when
+you a have a `Report` with many `BaseReporter`s registered and you want to
+be able to retry `open` after if failed only on some `BaseReporters.
```
hellais: Missing closing ```

bassosimone: woops, fixed, thank you!

hellais:
LGTM! Merging!

PR #769, opened 2016-08-05
Implement Neubot's DASH test

bassosimone:
Supersedes #572


include/measurement_kit/mlabns/mlabns.hpp
```diff
@@ -19,6 +19,8 @@ MK_DEFINE_ERR(MK_ERR_MLABNS(0), InvalidPolicyError, "")
 MK_DEFINE_ERR(MK_ERR_MLABNS(1), InvalidAddressFamilyError, "")
 MK_DEFINE_ERR(MK_ERR_MLABNS(2), InvalidMetroError, "")
 MK_DEFINE_ERR(MK_ERR_MLABNS(3), InvalidToolNameError, "")
+MK_DEFINE_ERR(MK_ERR_MLABNS(4), TooManyNegotiationsError, "")
+MK_DEFINE_ERR(MK_ERR_MLABNS(5), NegativeTimeError, "")
```
bassosimone: Note to self: these should be moved into the neubot namespace

example/neubot/dash.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <string>
+#include <unistd.h>
+
+using namespace mk;
+using namespace mk::neubot::negotiate;
+using namespace mk::net;
+using namespace mk::http;
+
+static const char *kv_usage =
+    "usage: ./example/http/dash [-vn] [-a address]\n";
```
bassosimone: Note to self: the address should be passed as the first argument

example/neubot/dash.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <measurement_kit/http.hpp>
+#include <measurement_kit/neubot.hpp>
+#include <stdlib.h>
+#include <string>
+#include <unistd.h>
+
+using namespace mk;
+using namespace mk::neubot::negotiate;
+using namespace mk::net;
+using namespace mk::http;
```
bassosimone: Note to self: do not import too much from namespaces to increase readability

example/neubot/dash.cpp
```diff
@@ -0,0 +1,52 @@
...
+            break;
+        default:
+            std::cout << kv_usage;
+            exit(1);
+        }
+    }
+
+    loop_with_initial_event([=]() {
+        run([=](Error error) {
+            debug("Error: %d", (int) error);
```
bassosimone: Note to self: this must be `error.code`

bassosimone:
[There is a memory leak](https://circleci.com/api/v1.1/project/github/measurement-kit/measurement-kit/641/output/9/0?file=true) that needs to be investigated.

include/measurement_kit/mlabns/mlabns.hpp
bassosimone: Actually there are redundant and they should be removed.

bassosimone:
Hmm, there is this memory leak that I really don't understand and cannot reproduce locally using a xenial Vagrant box. I have enabled build for this branch using gitlab to see whether this helps me understand what is actually going on. I am reluctant to disable checking for leaks and other memory errors using Valgrind or to create a suppression for this issue. For now, I'll put this pull request again "on hold" and remove it from the things to be implemented in 0.3.0. We'll see.

src/libmeasurement_kit/neubot/negotiate_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+template <MK_MOCK_NAMESPACE(http, request_sendrecv)>
+void loop_negotiate(Var<Transport> transport, Callback<Error> cb,
+                    Settings settings, Var<Reactor> reactor, Var<Logger> logger,
+                    int iteration = 0) {
+
+    std::array<int, 20> DASH_RATES{{100,  150,  200,  250,  300,   400,  500,
+                                    700,  900,  1200, 1500, 2000,  2500, 3000,
+                                    4000, 5000, 6000, 7000, 10000, 20000}};
```
bassosimone: Note to self: the array is duplicate.

src/libmeasurement_kit/neubot/negotiate_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+    settings["http/path"] = "/collect/dash";
+
+    request_sendrecv(
+        transport, settings,
+        {
+            {"Content-Type", "application/json"}, {"Authorization", auth},
+        },
+        body,
+        [=](Error error, Var<Response> res) {
+            if (error || (res->status_code != 200)) {
```
bassosimone: Note to self:

- `res` in theory *may* be `nullptr`
- we don't close the connection in this code path

src/libmeasurement_kit/neubot/negotiate_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+
+    request_sendrecv(
+        transport, settings,
+        {
+            {"Content-Type", "application/json"}, {"Authorization", ""},
+        },
+        body,
+        [=](Error error, Var<Response> res) {
+            if (error || (res->status_code != 200)) {
+                std::cout << "Error: " << (int)error;
```
bassosimone: Also here we should deal with:

- res that might be nullptr
- making sure we close connection
- logging on stdout

src/libmeasurement_kit/neubot/negotiate_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+                cb(HttpRequestFailedError());
+                return;
+            }
+
+            auto respbody = nlohmann::json::parse(res->body);
```
bassosimone: Here this function could fail, tearing down the whole library in such case.

src/libmeasurement_kit/neubot/negotiate_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+            std::string auth = respbody.at("authorization");
+            auto queue_pos = respbody.at("queue_pos");
+            auto real_address = respbody.at("real_address");
+            int unchoked = respbody.at("unchoked");
```
bassosimone: More code that might fail

src/libmeasurement_kit/neubot/negotiate_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+
+            // XXX
+            if (unchoked == 0) {
+                reactor->call_soon([=]() {
+                    loop_negotiate(transport, cb, settings, reactor, logger,
+                                   iteration + 1);
+                });
```
bassosimone: TODO: refactor this to avoid the else.

src/libmeasurement_kit/neubot/negotiate_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+            });
+            return;
+        }
+
+        request_connect(settings,
+                        [=](Error error, Var<Transport> transport) {
+
+                            if (error) {
+                                logger->warn("Error: %d", (int)error);
+                                transport->close([=]() { cb(error); });
```
bassosimone: I am not sure here we should close the transport. It will most likely be `nullptr`.

src/libmeasurement_kit/neubot/negotiate_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+        s["http/url"] = reply.url;
+
+        if (s["negotiate"] == "false") {
+            dash::run(s, [=](Error error, Var<Entry>) {
+                if (error) {
+                    logger->warn("Error: %d", (int)error);
+                    cb(error);
+                    return;
+                }
+            });
```
bassosimone: This is duplicate code. We can probably avoid it using a lambda.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+namespace mk {
+namespace neubot {
+namespace dash {
+
+const std::vector<int> &dash_rates(); // Implemented in dash.cpp
+
+static inline size_t select_lower_rate_index(int speed_kbit) {
+    size_t rate_index = 0;
+    while (dash_rates()[rate_index] < speed_kbit &&
+           rate_index < dash_rates().size()) {
```
bassosimone: Note to self: maybe here we can cache the return value of `dash_rates()`.

bassosimone: Or maybe we can declare `dash_rates()` to be `constexpr`?

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+        rate_index++;
+    }
+    if (rate_index > 0) { // Make sure we don't underflow
+        rate_index -= 1;
+    }
```
bassosimone: Here we can maybe add an assert on the value of `rate_index`, just in case.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+template <MK_MOCK_AS(http::request_send, http_request_send),
+          MK_MOCK_AS(http::request_recv_response, http_request_recv_response)>
+void run_loop_(Var<net::Transport> txp, int speed_kbit, std::string auth_token,
+               Var<report::Entry> entry, Settings settings,
+               Var<Reactor> reactor, Var<Logger> logger, Callback<Error> cb,
+               int iteration = 1) {
+
+    if (iteration > DASH_MAX_ITERATIONS) {
+        logger->debug("dash: completed all iterations");
```
bassosimone: Note to self: I should probably use the logger's progress mechanism for the dash test.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+#ifndef SRC_LIBMEASUREMENT_KIT_NEUBOT_DASH_IMPL_HPP
+#define SRC_LIBMEASUREMENT_KIT_NEUBOT_DASH_IMPL_HPP
+
```
bassosimone: Note to self: here we should probably add a reference to the paper with the methodology.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+    if (iteration > DASH_MAX_ITERATIONS) {
+        logger->debug("dash: completed all iterations");
+        cb(NoError());
+        return;
+    }
+
+    size_t rate_index = select_lower_rate_index(speed_kbit);
+    int rate_kbit = dash_rates()[rate_index];
+    int count = ((rate_kbit * 1000) / 8) * DASH_SECONDS;
```
bassosimone: Note to self: it would probably be best to explain what these three lines do.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+            {"Authorization", auth_token},
+        },
+        "", [=](Error error, Var<http::Request> req) {
+            if (error) {
+                logger->warn("dash: request failed: %s",
+                             error.explain().c_str());
+                cb(error);
+                return;
+            }
+            assert(!!req);
```
bassosimone: TODO: make sure that it cannot happen that `req` is `nullptr`. And possibly also document which is the expected behaviour in the HTTP library. I've added this assert because that is what I would expect but have not checked yet.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+                                     error.explain().c_str());
+                        cb(http::HttpRequestFailedError());
+                        return;
+                    }
+                    res->request = req;
+                    double length = res->body.length();
+                    double time_elapsed = mk::time_now() - saved_time;
+                    if (time_elapsed <= 0) { // For robustness
+                        logger->warn("dash: invalid time error");
+                        cb(InvalidTimeError());
```
bassosimone: This error can probably have a better name

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+                        {"iteration", iteration},
+                        //{"platform", sys.platform}
+                        {"rate", rate_kbit},
+                        //{"real_address", self.parent.real_address}
+                        //{"received", received}
+                        //{"remote_address", stream.peername[0]}
+                        //{"request_ticks", self.saved_ticks}
+                        //{"timestamp", mk::time_now}
+                        //{"uuid", self.conf.get("uuid")}
+                        {"version", MEASUREMENT_KIT_VERSION}});
```
bassosimone: We should probably add a field in the spec that says we're running MK.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+                    res->request = req;
+                    double length = res->body.length();
+                    double time_elapsed = mk::time_now() - saved_time;
+                    if (time_elapsed <= 0) { // For robustness
+                        logger->warn("dash: invalid time error");
+                        cb(InvalidTimeError());
+                        return;
+                    }
+                    // TODO: we should fill all the required fields
+                    (*entry)["receiver_data"].push_back(report::Entry{
```
bassosimone: I should probably write a OONI specification for this test saying exactly what should be saved by the OONI collector (the entry here should be what is saved by Neubot).

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+                        //{"timestamp", mk::time_now}
+                        //{"uuid", self.conf.get("uuid")}
+                        {"version", MEASUREMENT_KIT_VERSION}});
+                    double speed = length / time_elapsed;
+                    double s_k = (speed * 8) / 1000;
+                    logger->info("[%d/%d] rate: %d kbit/s, speed: %.2f "
+                                 "kbit/s, elapsed: %.2f s",
+                                 iteration, DASH_MAX_ITERATIONS, rate_kbit, s_k,
+                                 time_elapsed);
+                    if (time_elapsed > DASH_SECONDS) {
```
bassosimone: This part where we scale down the speed significantly when we think that there was congestion had been criticised during the review and was introduced under the assumption that Neubot was a background tool, but there are also use cases where we run in the foreground. I think it should be on by default to be compatible with Neubot but, moving forward, it should probably be off by default and we should have a bit into the results that indicates whether this was ON or OFF.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+              Settings settings, Var<Reactor> reactor, Var<Logger> logger,
+              Callback<Error> cb) {
+    settings["http/url"] = url;
+    settings["http/method"] = "GET";
+    logger->info("Start dash test with: %s", url.c_str());
+    http_request_connect(
+        settings,
+        [=](Error error, Var<net::Transport> txp) {
+            if (error) {
+                assert(!txp);
```
bassosimone: Verify that this assertion is what we really expect and possibly document this.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+                logger->warn("dash: cannot connect to server: %s",
+                             error.explain().c_str());
+                cb(error);
+                return;
+            }
+            // Note: from now on, we own `txp`
+            assert(txp);
```
bassosimone: Maybe write this as `assert(!!txp)`?

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+            callback(NoError(), auth);
+        },
+        reactor, logger);
+}
+
+template <MK_MOCK_AS(http::request_sendrecv, http_request_sendrecv)>
+void collect_(Var<net::Transport> txp, Var<report::Entry> entry,
+              std::string auth, Settings settings, Var<Reactor> reactor,
+              Var<Logger> logger, Callback<Error> cb) {
+    std::string body = entry->dump();
```
bassosimone: Problem: the original format of the entry was exactly the one of Neubot. Then I have basically modified it such that, now, what we send is not what Neubot would except. I guess here what we should dump is the specific key that contains the list of results rather than the whole object.

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+}
+
+template <MK_MOCK_AS(http::request_connect, http_request_connect),
+          MK_MOCK_AS(http::request_sendrecv, http_request_sendrecv_negotiate),
+          MK_MOCK_AS(http::request_sendrecv, http_request_sendrecv_collect)>
+void negotiate_with_(std::string url, Var<report::Entry> entry,
+                     Settings settings, Var<Reactor> reactor,
+                     Var<Logger> logger, Callback<Error> cb) {
+    logger->info("Negotiating with: %s", url.c_str());
+    settings["http/url"] = url;
```
bassosimone: For robustness here we should also probably set the method explicitly

src/libmeasurement_kit/neubot/dash_impl.hpp
```diff
@@ -0,0 +1,345 @@
...
+}
+
+template <MK_MOCK_AS(mlabns::query, mlabns_query)>
+void negotiate_impl(Var<report::Entry> entry, Settings settings,
+                    Var<Reactor> reactor, Var<Logger> logger,
+                    Callback<Error> cb) {
+    if (settings.find("url") != settings.end()) {
+        negotiate_with_(settings["url"], entry, settings, reactor, logger, cb);
+        return;
+    }
```
bassosimone: Note: I think that, on Neubot, we usually specify address and port rather than a URL, so perhaps here it would be much better to have in input address and port rather than URL as well. (Also, the command line should be modified such that it is possible to pass address and port.)

src/measurement_kit/cmd/dash.cpp
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+    argc -= optind, argv += optind;
+    if (argc > 1) {
+        fprintf(stderr, "%s\n", USAGE);
+        exit(1);
+        /* NOTREACHED */
+    }
+    if (argc == 1) {
+        test.set_options("url", argv[0]);
```
bassosimone: This usage is probably broken because the code expects a URL not a hostname (to be tested).

test/neubot/dash.cpp
```diff
@@ -0,0 +1,39 @@
+#define CATCH_CONFIG_MAIN
```
bassosimone: TODO: make sure that tests cover all possible errors and are up-to-date with respect to changes in the code (I am not even sure whether currently these tests compile cleanly).

bassosimone:
This diff now looks good to me. I'll merge it into the integration branch as soon as travis is okay.

Cc: @hellais

PR #789, opened 2016-08-24
ssl: handle errors using callbacks not exceptions

bassosimone:
Rather than letting the application die because the CA bundle is
missing, let the connect() attempt fail gracefully.

It occurred to me to think that this was needed when debugging
a crash in netprobe-android with @lorenzoPrimi.


src/libmeasurement_kit/net/ssl-context.cpp
```diff
@@ -42,44 +37,42 @@ void SslContext::init(std::string ca_bundle_path) {
...
-    #endif
+/*static */ ErrorOr<Var<SslContext>> SslContext::make(std::string path) {
+    static Var<SslContext> singleton;
+    // We basically cache the last created context and we reuse it if the path
+    // has not been changed, otherwise we create a new context
+    if (!singleton or singleton->ca_bundle_path != path) {
+        singleton.reset(new SslContext);
+        Error err = singleton->init(path);
+        if (err) {
+            singleton.reset(); // Basically: "this is Sparta!"
```
AntonioLangiu: lol

PR #795, opened 2016-08-26
Feature/android.mk

bassosimone:


bassosimone:
Closing because a better approach is implemented in measurement-kit/measurement-kit-android#16

PR #797, opened 2016-08-27
Teach acinclude.m4 to use brew's OpenSSL

bassosimone:


bassosimone:
cc: @AntonioLangiu and @hellais &mdash; from now on you should not pass the annoying flag to point to the brew's version of OpenSSL anymore :-)

PR #809, opened 2016-09-06
geoip: memoize latest databases, remove json-based api

### LINKED ISSUE #737: geoip: remove API returning a JSON

bassosimone:
Closes #737 #736


bassosimone:
Setting "on hold" again: I've figure out this scheme will break when we update databases on the fly, a case when we do not want to keep the previous database openned.

bassosimone:
@hellais this avoids opening GeoIP database whenever we use GeoIP services. It started as a candidate hotfix but then grow more complex, since I also considered the case where you want to close the database, update the files, and then reopen the database. Bottom line is that I do not think I can merge this directly and a review of this diff would be very welcome!

src/libmeasurement_kit/ooni/ooni_test.cpp
```diff
@@ -69,6 +69,7 @@ void OoniTest::geoip_lookup(Callback<> cb) {
     probe_cc = "ZZ";
     ip_lookup(
         [=](Error err, std::string ip) {
+            // TODO: refactor to use the if-err-then-cb-and-return style
```
hellais: Also the if else nesting is quite nasty here.

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -23,14 +23,46 @@ void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
     resolver_lookup_impl(callback, settings, reactor, logger);
 }
 
-IPLocation::IPLocation(std::string path_country_, std::string path_asn_,
+/* static */ Var<IpLocationProxy> IpLocationProxy::global() {
+    static Var<IpLocationProxy> singleton(new IpLocationProxy);
+    return singleton;
+}
+
+Var<IpLocation> IpLocationProxy::open(std::string country, std::string asn,
```
hellais: I find it a bit misleading that this method name is called `open`. I would expect that an open method would actually create open handles to the various geoip files, but instead what it does is that it just creates an instance of `IPLocation`. The files are actually opened lazily when you first attempt to resolve something.

At the minimum I would add a line stating that this is the behaviour.

src/libmeasurement_kit/ooni/ooni_test.cpp
```diff
@@ -77,30 +78,27 @@ void OoniTest::geoip_lookup(Callback<> cb) {
...
+                    ErrorOr<std::string> s;
+                    s = gi->resolve_asn(ip);
+                    if (!!s) {
+                        logger->info("probe_asn: %s", s->c_str());
                         if (options.get("save_real_probe_asn", true)) {
-                            probe_asn = (*res)["asn"];
+                            probe_asn = *s;
                         }
-                        logger->info("probe_asn: %s", probe_asn.c_str());
+                    }
```
hellais: Maybe we should write a log message when `!s`.

src/libmeasurement_kit/ooni/ooni_test.cpp
```diff
@@ -77,30 +78,27 @@ void OoniTest::geoip_lookup(Callback<> cb) {
...
-                        // better to let the exception unwind than just print
-                        // a warning, because the former is easier to notice
-                        // and therefore fix during development
+                    auto gi = IpLocationProxy::global()->open(
+                                cntry_p, asn_p, "", logger);
+                    ErrorOr<std::string> s;
+                    s = gi->resolve_asn(ip);
+                    if (!!s) {
+                        logger->info("probe_asn: %s", s->c_str());
                         if (options.get("save_real_probe_asn", true)) {
```
hellais: Since we actually don't do anything with the lookup when it fails, we may want to just skip doing it if `save_real_probe_XXX` is false.

src/libmeasurement_kit/ooni/ooni_test.cpp
```diff
@@ -77,30 +78,27 @@ void OoniTest::geoip_lookup(Callback<> cb) {
...
                     probe_ip = ip;
                 }
-                std::string country_p =
-                    options.get("geoip_country_path", std::string{});
-                std::string asn_p =
-                    options.get("geoip_asn_path", std::string{});
-                if (country_p == "" or asn_p == "") {
+                auto cntry_p = options.get("geoip_country_path", std::string{});
+                auto asn_p = options.get("geoip_asn_path", std::string{});
+                if (cntry_p == "" or asn_p == "") {
```
hellais: Maybe we can check only for the paths we actually need. That is `cntry_p` if `save_real_probe_country` is true and `asn_p` if `save_real_probe_asn` is true.

src/libmeasurement_kit/ooni/utils.cpp
hellais: Maybe a more appropriate name would be `set_paths`.

hellais:
Ok I took a look at this branch and sprinkled some comments on it.

This is what I would do to proceed in merging:

* [x] First pass at reviewing the branch

* [ ] merge/rebase it onto master (can you do this @bassosimone ?)

* [ ] second pass at code review

* [ ] local testing

* [ ] Merge.

src/libmeasurement_kit/ooni/ooni_test.cpp
bassosimone: Yes, indeed, that's why I'd like to refactor! I've created #815 to take care of this.

bassosimone:
okay @hellais I've done another pass at improving this pull request trying to take into account your suggestions plus trying to further simplify the code (i.e. rather than have one class managing three geoip databases at a time, now there's just one class per geoip database). I think this is is again ready for review!

include/measurement_kit/common/error.hpp
```diff
@@ -45,7 +45,7 @@ class Error : public std::exception {
         child_errors.push_back(container);
     }
 
-    std::string explain() {
+    std::string explain() const {
```
bassosimone: In a couple of places I was deadling with `const Error` and this change helped

src/libmeasurement_kit/ooni/ooni_test.cpp
```diff
@@ -74,47 +74,55 @@ void OoniTest::run_next_measurement(size_t thread_id, Callback<Error> cb,
...
+                try {
+                    probe_asn = *GeoipCache::global()
+                        ->resolve_asn(asn_path, ip, logger);
+                } catch (const Error &err) {
+                    logger->warn("cannot lookup asn: %s",
+                                 err.explain().c_str());
                 }
+            } else if (asn_path == "") {
+                logger->warn("geoip_asn_path is not set");
             }
```
bassosimone: In this hunk I refactored the code in here to reduce cyclomatic complexity

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -23,141 +23,103 @@ void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
...
-    const char *result;
-    result = GeoIP_country_name_by_name_gl(gi_country, ip.c_str(), &gl);
-    if (result == nullptr) {
-        return GenericError();
-    }
-    std::string country_name = result;
-    return country_name;
+        std::string country_code = result;
+        return country_code;
+    }, logger);
```
bassosimone: Most of the hunk above disappears if we don't consider space changes

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -23,141 +23,103 @@ void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
...
-    }
-    std::string result;
-    if (gir->city != nullptr) {
-        result = gir->city;
-    }
-    GeoIPRecord_delete(gir);
-    return result;
+        std::string country_name = result;
+        return country_name;
+    }, logger);
```
bassosimone: Most of the hunk above disappears if we don't consider space changes

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -23,141 +23,103 @@ void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
...
-    asn = split(asn).front(); // We only want ASXX
-    free(res);
-    return asn;
+        std::string result;
+        if (gir->city != nullptr) {
+            result = gir->city;
+        }
+        GeoIPRecord_delete(gir);
+        return result;
+    }, logger);
```
bassosimone: Most of the diff related to this function above disappears if we don't consider space changes

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -23,141 +23,103 @@ void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
...
+    return action();
 }
 
-ErrorOr<std::string> IPLocation::resolve_country_name(std::string ip,
-                                                      Var<Logger> logger) {
-    if (gi_country == nullptr) {
-        gi_country = GeoIP_open(path_country.c_str(), GEOIP_MEMORY_CACHE);
-        if (gi_country == nullptr) {
-            logger->warn("IPLocation: cannot open geoip country database");
-            return CannotOpenGeoIpCountryDatabaseError();
```
bassosimone: This code that opens the database has been refactored inside the method called `with_open_database_do`

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -23,141 +23,103 @@ void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
...
-    }
-    ErrorOr<std::string> country_name = ip_location.resolve_country_name(ip);
-    if (!country_name) {
-        return country_name.as_error();
-    }
-    ErrorOr<std::string> city_name;
-    if (path_city != "") {
-        city_name = ip_location.resolve_city_name(ip);
-        if (!city_name) {
-            return city_name.as_error();
```
bassosimone: This method returning a json has been removed

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -23,141 +23,103 @@ void resolver_lookup(Callback<Error, std::string> callback, Settings settings,
...
-    node["asn"] = asn.as_value();
-    if (path_city != "") {
-        node["city_name"] = city_name.as_value();
-    }
-    return node;
+        std::string asn = res;
+        asn = split(asn).front(); // We only want ASXX
+        free(res);
+        return asn;
+    }, logger);
```
bassosimone: Also in this case most changes are space changes

hellais:
This looks good to me!

🌐

PR #812, opened 2016-09-09
Port measurement-kit to Windows

bassosimone:


CMakeLists.txt
```diff
@@ -7,22 +7,55 @@ cmake_minimum_required(VERSION 3.6)
...
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
 file(
      GLOB
      sources
      "src/libmeasurement_kit/*/*.c"
      "src/libmeasurement_kit/*/*.cpp"
     )
 
+add_definitions(-DMK_CA_BUNDLE="mk-ca-bundle.pem")
```
bassosimone: This is to tell the compiled code where to look for a CA bundle. We can change this in the future, but for now it works well having `measurement_kit.exe` and the CA bundle in the same directory.

CMakeLists.txt
```diff
@@ -7,22 +7,55 @@ cmake_minimum_required(VERSION 3.6)
 # Note: CMake is mainly meant for usage under Windows, while under Unix we
 # prefer to use autotools; hence assume for Unix this file has quirks.
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
bassosimone: Not strictly needed, perhaps it would be better to lower it back to 11.

CMakeLists.txt
```diff
@@ -7,22 +7,55 @@ cmake_minimum_required(VERSION 3.6)
...
      "src/libmeasurement_kit/*/*.c"
      "src/libmeasurement_kit/*/*.cpp"
     )
 
+add_definitions(-DMK_CA_BUNDLE="mk-ca-bundle.pem")
+
 if(APPLE)
   set(openssl_include_path /usr/local/Cellar/openssl/1.0.2h_1/include)
+  set(osx_extra_libs event_pthreads event_openssl)
+  set(osx_link_path /usr/local/Cellar/openssl/1.0.2h_1/lib)
```
bassosimone: The above is quite *ghetto*, but the point of this file is to enable compilation on Windows and building on macOS for now is just something I do to make sure I haven't broken anything.

Makefile.am
```diff
@@ -26,7 +26,6 @@ noinst_PROGRAMS    = # Empty
 ALL_TESTS          = # Empty
 
 include include.am
-include src/libmeasurement_kit/ext/include.am
```
bassosimone: Now dependencies are explicitly committed, to simplify cross platform development. As such, they can be computed automatically and some other complexity is gone.

acinclude.m4
```diff
@@ -36,6 +36,21 @@ AC_DEFUN([MK_AM_DISABLE_INTEGRATION_TESTS], [
...
+    errx \
+    warn \
+    warnx \
+    getopt \
+    getopt_long \
+    getopt_long_only \
+    gmtime_r \
+    strcasecmp \
+    strtonum \
+  ])
```
bassosimone: For each of these functions, if the function is not present on the system, we supply our own.

include/measurement_kit/common.hpp
```diff
@@ -1,6 +1,7 @@
 // File autogenerated by `./autogen.sh`, do not edit
 #ifndef MEASUREMENT_KIT_COMMON_HPP
 #define MEASUREMENT_KIT_COMMON_HPP
+#include <measurement_kit/common/aaa_base.hpp>
```
bassosimone: All system specific headers are now included by this one.

include/measurement_kit/common/utils.hpp
```diff
@@ -29,6 +29,7 @@ T split(std::string s, std::string pattern = "\\s+") {
     return {first, last};
 }
 
+int strcasecmp_wrapper(const char *, const char *);
```
bassosimone: Need this wrapper because previously the public http header was pulling directly `strcasecmp` which is a function that is not available on Windows. Better to pull the wrapper and use it to redirect to impl.

include/measurement_kit/dns/dns.hpp
```diff
@@ -36,132 +32,10 @@ MK_DEFINE_ERR(MK_ERR_DNS(16), UnsupportedTypeError, "")
 
 Error dns_error(int code); ///< Map evdns code to proper error
 
-/// Available query classes id
```
bassosimone: This chunk of code was ugly and also was getting in the way complicating the porting. I prefer strings and also @hellais expressed the same preference. Now I have an excuse to nuke it.

include/measurement_kit/report/entry.hpp
```diff
@@ -20,7 +20,7 @@ class Entry : public nlohmann::json {
 
     template <typename T> operator ErrorOr<T>() {
         try {
-            return nlohmann::json::operator T();
+            return get<T>();
```
bassosimone: MSVC++ was not able to compile the code, so I implemented it another way.

src/libmeasurement_kit/cmdline/dns_injection.cpp
```diff
@@ -19,10 +20,10 @@ int main(const char *, int argc, char **argv) {
     mk::ooni::DnsInjection test;
     int ch;
 
-    while ((ch = getopt(argc, argv, "b:nv")) != -1) {
+    while ((ch = mkp_getopt(argc, argv, "b:nv")) != -1) {
```
bassosimone: For all the portable standard library symbols, we always use the `mkp_foo` version. A set of macros in `"../portable/api.h"` would determine whether we use our own implementation (the one starting with `mkp_`) or the one provided by the system. In the latter case, macros are used to turn `mkp_foo` into `foo`.

src/libmeasurement_kit/common/poller.hpp
```diff
@@ -21,24 +21,54 @@ void mk_do_periodic_cb(evutil_socket_t, short, void *ptr);
...
             throw std::runtime_error("sigaction() failed");
         }
+#else
+        WSADATA wsa_data;
+        WORD version = MAKEWORD(2, 2);
+        memset(&wsa_data, 0, sizeof (wsa_data));
+        if (WSAStartup(version, &wsa_data) != 0) {
+            throw std::runtime_error("WSAStartup() failed");
+        }
+#endif
```
bassosimone: We need to initialize winsock2 or the programs linked with MK won't work.

src/libmeasurement_kit/common/poller.hpp
```diff
@@ -21,24 +21,54 @@ void mk_do_periodic_cb(evutil_socket_t, short, void *ptr);
...
+#else
+# define mk_use_platform_threads evthread_use_pthreads
+#endif
+
+template <
+    MK_MOCK(mk_use_platform_threads)
+#ifndef _WIN32
+    , MK_MOCK(sigaction)
+#endif
+> class MkLibrarySingleton {
```
bassosimone: Some conditional inclusion and macro magic to use windows threads on windows.

src/libmeasurement_kit/common/utils.cpp
```diff
@@ -57,8 +48,8 @@ timeval *timeval_init(timeval *tv, double delta) {
     if (delta < 0) {
         return nullptr;
     }
-    tv->tv_sec = (time_t)floor(delta);
-    tv->tv_usec = (suseconds_t)((delta - floor(delta)) * 1000000);
```
bassosimone: These cases were creating portability problems (suseconds_t not defined on Windows and struct timeval has all fields as long), so I decided to sacrifice them to move forward, without losing pretty much anything.

src/libmeasurement_kit/traceroute/android.cpp
```diff
@@ -33,32 +33,21 @@
 /// Android implementation of prober
 
 // This is meant to run on Android but can run on all Linux systems
-#ifdef __linux__
+#ifndef __linux__
+void mk_traceroute_android_unused();
+void mk_traceroute_android_unused() {}
```
bassosimone: To avoid compiler warning

bassosimone:
@hellais this is now ready for review: the code compiles on Windows and the executable works (I tested NDT and web_connectivity). I'll leave to you to decide whether this could go in for 0.3.0 as I originally wanted to do, or whether to postpone: I reckon this is not super small.

bassosimone:
We decided to postpone this to v0.4.0. I will shortly close this and reissue it as a pull request with respect to the master branch.

Issue #777, opened 2016-08-12
ooni: add support for passing ASN and country code when starting a test


hellais:
In MK for android and iOS it's necessary to lookup the country code of the user in order to generate the test list specific to the country of the user.

To avoid doing a geo ip lookup twice we should support passing the ASN and country code to the `begin()` method and this will bypass the geoip lookup.

bassosimone:
This has now been merged into master.

PR #816, opened 2016-09-19
ooni_test.cpp: allow caller to set probe_{ip,asn,cc}

bassosimone:
This pull request addresses #777


hellais:
LGTM.

hellais:
Note: travis is failing because the cert for the test collector expired. I would switch over to using b.collector.test.ooni.io, since that one is going to be more stable for testing.

bassosimone:
> Note: travis is failing because the cert for the test collector expired. I would switch over to using b.collector.test.ooni.io, since that one is going to be more stable for testing.

Will commit this soon, thank you!

PR #820, opened 2016-09-20
Port measurement-kit to Windows

bassosimone:


include/measurement_kit/portable/strings.h
```diff
@@ -8,4 +8,16 @@
 #include <strings.h>
 #endif
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int mkp_strcasecmp(const char *s1, const char *s2);
+
+int mkp_strncasecmp(const char *s1, const char *s2, size_t n);
```
bassosimone: Here we should add translation macros for unix

include/measurement_kit/portable/sys/time.h
```diff
@@ -4,7 +4,20 @@
...
 
-#ifndef _WIN32
+#ifdef _WIN32
+#include <measurement_kit/portable/_windows.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int gettimeofday(struct timeval *tvp, void *tzp);
```
bassosimone: Here we should explain what is going on on Windows, i.e. why we need to declare the prototype

src/libmeasurement_kit/common/utils.cpp
```diff
@@ -279,4 +281,8 @@ double percentile(std::vector<double> v, double percent) {
     return val0 + val1;
 }
 
+int strcasecmp_wrapper(const char *s, const char *r) {
+    return mkp_strcasecmp(s, r);
+}
```
bassosimone: This should not be here

bassosimone:
MK now compiles for MSVC and Mingw-w64. Some bits of MK do not compile yet but they're easy and I have addressed this in more than one branch, so this branch can finally be closed. 🎉

PR #822, opened 2016-09-27
Lua integration using sol

bassosimone:


bassosimone:
I guess we can also use this. We will start off with Python, and this is just a simple PoC. If and when we really want to use lua, we can dig back into the history and learn from this.

PR #823, opened 2016-09-30
Allow to run tests using a string as input

### LINKED ISSUE #308: Tests should run with a string as input
It would be nice while defining the test to use a string as input. 

For example [here](https://github.com/measurement-kit/measurement-kit/blob/master/include/measurement_kit/ooni/dns_injection_test.hpp#L30)  we should enable the user to pass a string to the test.

bassosimone:
Closes #308.


bassosimone:
Putting on hold and removing milestone. This branch does not follow the correct approach, which would be instead of putting the input into the std::istream form which we read usually.

hellais:
I seem to recall you mentioning that you added support for this. I guess not as part of this branch. Can this be closed?

hellais:
Edit: found it!
Does https://github.com/measurement-kit/measurement-kit/pull/1132 superseed this?

bassosimone:
Yes, closing the branch right now, thanks!

PR #824, opened 2016-10-02
Simplify driving MK from a scripting language

bassosimone:
1. runner: add generic function to run background thread task
2. add scriptable API for OONI and NDT
3. stop background thread when no tasks are scheduled
4. reliability fixes and improvements


src/libmeasurement_kit/common/runner.cpp
```diff
@@ -30,27 +42,42 @@ void Runner::run_test(Var<NetTest> test, Callback<Var<NetTest>> fn) {
...
+        kickoff([=](Callback<> end) {
+            debug("runner: ending %d", task_id);
+            // For robustness, delay the final callback to the beginning of
+            // next I/O cycle to prevent possible user after frees. This
+            // could happen because, in our current position on the stack,
+            // we have been called by `NetTest` code that may use `this`
+            // after calling the callback. But this would be a problem
+            // because `test` is most likely to be destroyed after `fn()`
+            // returns. This, when unwinding the stack, the use after free
+            // would happen.
```
bassosimone: Uhm, this comment should be updated since the code changed. In particular here we should not reference `NetTest` directly since we're not dealing with this class directly anymore.

bassosimone:
The gist of this PR is very simple. An API that inputs and outputs strings. I discussed this with @hellais in Rome some days ago. We concluded it was smarter to just wrap existing APIs, like we currently do on Java. We will discuss wrapping basic primitives further with @joelanders next week. In any case, this is so simple that it can be redone based on master. Merging forward this PR does not seem obvious.

PR #885, opened 2016-10-29
Little fix for ndt json

AntonioLangiu:
While checking the output of NDT with @bassosimone we discovered this bug, where the JSON doesn't have a valid type entry.


AntonioLangiu:
Sorry, I need to close this PR because I had to open it with `stable`

PR #886, opened 2016-10-29
Bug in ndt JSON

AntonioLangiu:
While checking the output of NDT with @bassosimone we discovered this bug, where the JSON doesn't have a valid type entry.


bassosimone:
Thank you!

PR #898, opened 2016-11-01
Implement and use reactor->pollfd()

bassosimone:
One of the v0.4.x objectives is to integrate c-ares and use it for DNS. Now, to integrate c-ares we need to either use directly libevent's `event_once()` API or to write a wrapper for it. Since I'm working towards moving libevent code all into a specific place, I've opted for the latter. This diff exposes a poll(2)-like interface that polls "just once" a specific file descriptor and possibly returns timeout error if the specified timeout has expired. It also checks this API works by modifying Android traceroute to use it.


AntonioLangiu:
It looks good to me! Merging

PR #710, opened 2016-07-09
ndt: multi stream test

bassosimone:

PR #899, opened 2016-11-02
net: measure connect time during connect

bassosimone:
This diff has been extracted from #710 to ease review.

The plan is to merge this into master and then merge master in #710.


AntonioLangiu:
It looks good to me!

PR #902, opened 2016-11-02
Prepare for required successful build only on travis

bassosimone:
So, circle-ci was taking too much time to compile. There are ways to speed up the build but they seem to offer more in-build parallelism than parallel builds (unlike travis). This slows me down too much when I have more than one branch that I am working on. Thus prepare to return to a situation where only travis is run after every push. To do this, we need to have an acceptable (in terms of correctness and speed) valgrind run also on travis. In line I explain how I am going to do that.

(Needless to say, dealing with travis & co was - as always - a huge blood bath...)


.travis.yml
```diff
@@ -1,64 +1,60 @@
...
+        --with-openssl=builtin --with-libevent=builtin $COVERAGE
+    - make $pkg_make_flags
+    - make $pkg_make_flags check-am$VALGRIND
+    - if [ -f test-suite.log ]; then cat test-suite.log; fi
 
 # Silence gcov standard output since it produces way too much output
 after_success:
     - if [ "$COVERAGE" = "--enable-coverage" ]; then
-        coveralls --gcov /usr/bin/gcov-5 --exclude src/libmeasurement_kit/ext
+        coveralls --gcov /usr/bin/gcov-6 --exclude src/libmeasurement_kit/ext
```
bassosimone: Migrate to gcc-6

acinclude.m4
```diff
@@ -172,20 +176,21 @@ AC_DEFUN([MK_AM_OPENSSL], [
...
-    fi
-    mk_not_found=1
-  ],
-  [AC_MSG_RESULT([Skip the test because we are cross-compiling])])
+      ])
+    ],
+    [AC_MSG_RESULT([yes])],
+    [
+      AC_MSG_RESULT([no])
+      if test -d /usr/local/Cellar/openssl; then
```
bassosimone: Here, btw, I fixed the test adding `-d` otherwise the test was always true :)

configure.ac
```diff
@@ -29,9 +37,11 @@ MK_AM_DISABLE_INTEGRATION_TESTS
...
 
 # Must set -pthread before testing for -levent_pthreads
-CFLAGS="$CFLAGS -pthread"
-CXXFLAGS="$CXXFLAGS -pthread"
-LDFLAGS="$LDFLAGS -pthread"
+if test "`uname`" != "Darwin"; then
+  CFLAGS="$CFLAGS -pthread"
+  CXXFLAGS="$CXXFLAGS -pthread"
+  LDFLAGS="$LDFLAGS -pthread"
+fi
```
bassosimone: Get rid of annoying warnings about `-pthread` on macOS.

PR #907, opened 2016-11-09
Add support for submitting reports via cloudfronted collectors

hellais:


bassosimone:
@hellais tests have been broken by this commit, probably because the template proto changed

bassosimone:
Thank you!

PR #811, opened 2016-09-08
Implement ooni resources downloader

bassosimone:
Closes #649 

# TODO list

- [x] repair build
- [x] address TODO comments
- [x] add comment suggesting possible improvement: use atomic integer to speed up in endgame mode
- [x] add specific errors
- [x] add regress tests
- [x] address comments by @hellais

PR #908, opened 2016-11-09
http/request: don't redirect if *max_request <= 0

bassosimone:
If the user has not asked to follow redirect (i.e. the max number of
redirects is zero or less than that), do not enter into the branch
where we follow the redirect even if there is a `location` header in
the received headers set.

Fix originally implemented to #811 and backported as hotfix.


bassosimone:
I am going to merge this: the reason why [circle-ci fails](https://circleci.com/gh/measurement-kit/measurement-kit/1125) can be traced back to ipv6 domain resolution issues for google.com.

PR #912, opened 2016-11-10
Log dns debug to the correct logger

AntonioLangiu:
We noticed using MK on Android that some logs were not on the correct logger!

A solution that I found to pass the logger to the libevent callback was to put the logger inside the context structure that we pass to evdns_resolver_*


src/libmeasurement_kit/libevent/dns_impl.hpp
```diff
@@ -40,12 +40,16 @@ class QueryContext : public NonMovable, public NonCopyable {
     Var<Message> message;
     Callback<Error, Var<Message>> callback;
 
+    Var<Logger> logger;
```
bassosimone: For robustness I'd initialize also this using `Logger::global()`

bassosimone:
Thank you!

Issue #904, opened 2016-11-07
dns: support multi dns settings


AntonioLangiu:
With the actual api we are able to set just one DNS resolver, due to the fact that the settings doesn't support multiple values. I discovered this trying measurement-kit on Android on a network where the first DNS resolver (that was the only one setted on measuremen-kit) was not working. 

This depends on #837

AntonioLangiu:
A possible solution for this problem if we are not able to use json as settings is to have more fields for dns nameserver, like 

    dns/nameserver1
    dns/nameserver2
    dns/nameserver3
    dns/nameserverN

bassosimone:
I think it's important to enable this in v0.9.0 where, in particular, we'll also have `c-ares`.

bassosimone:
Will consider doing something like that in https://github.com/ooni/probe-engine/issues/18

PR #913, opened 2016-11-10
dns: take 4 resolvers

AntonioLangiu:
This fast fix help us using the library in mobile applications where we need to set more than one dns resolver.

This is an hotfix to solve #904 until we refactor the settings class to support string vectors.


bassosimone:
Superseded by #1218.

PR #915, opened 2016-11-11
Support IPv6 link-local resolver

darkk:
Works-on-my-machine.jpeg (Ubuntu Linux), should work for Android as well.


src/libmeasurement_kit/libevent/dns_impl.hpp
```diff
@@ -53,37 +54,66 @@ class QueryContext : public NonMovable, public NonCopyable {
...
-    if (settings.find("dns/nameserver") != settings.end()) {
-        if ((base = evdns_base_new(evb, 0)) == nullptr) {
-            throw std::bad_alloc();
+    if (settings.count("dns/nameserver")) {
+        // libevent can't handle link-local IPv6 nameserver in
+        // evdns_base_nameserver_ip_add, and there is no way to parse alike
+        // addresses in platform-independent way, so that's why getaddrinfo().
+        // fe80::227:22ff:fe45:3a92%wlan0 -- Android
+        // fe80::227:22ff:fe45:3a92%wlp1s0 -- Ubuntu Linux
+        evutil_addrinfo hints = { };
```
bassosimone: I assume here you're assigning to zero the POD. I seem to remember that assigning to `{}` has this effect. Is my understanding correct? Do you have a reference explaining that behavior?

src/libmeasurement_kit/libevent/dns_impl.hpp
```diff
@@ -53,37 +54,66 @@ class QueryContext : public NonMovable, public NonCopyable {
...
+};
+
+struct evaddrinfo_deleter {
+    void operator()(evutil_addrinfo* p) {
+        evutil_freeaddrinfo(p);
+    }
+};
+
+using evdns_base_uptr = std::unique_ptr<evdns_base, evdns_base_deleter>;
+using evaddrinfo_uptr = std::unique_ptr<evutil_addrinfo, evaddrinfo_deleter>;
```
bassosimone: I appreciate the usage of RAII here. Note to self: As part of more significant refactoring of this file, I'd like the function modified by this diff to return a `Var<event_base>` (where `Var<>` is a `shared_ptr<>` wrapper will null pointer exception checks) with the proper deleter set (I've typically used a lambda rather than a struct for this).

src/libmeasurement_kit/libevent/dns_impl.hpp
```diff
@@ -53,37 +54,66 @@ class QueryContext : public NonMovable, public NonCopyable {
...
         MK_MOCK(evdns_base_free), MK_MOCK(evdns_base_set_option)>
 static inline evdns_base *create_evdns_base(
         Settings settings, Var<Reactor> reactor = Reactor::global()) {
 
-    evdns_base *base;
     event_base *evb = reactor->get_event_base();
+    const int initialize_nameservers = settings.count("dns/nameserver") ? 0 : 1;
+    evdns_base_uptr base(evdns_base_new(evb, initialize_nameservers));
+    if (!base) {
+        throw std::bad_alloc();
```
bassosimone: Note to self: as part of future refactoring, I'd like this function to return `Var<event_base>` or possibly even better if it would return a `ErrorOr<Var<event_base>>` rather than throwing.

src/libmeasurement_kit/libevent/dns_impl.hpp
darkk: We have same understanding. I assume, that's what [zero initialization](http://en.cppreference.com/w/cpp/language/zero_initialization) is, but I'm not a language lawyer.

src/libmeasurement_kit/libevent/dns_impl.hpp
bassosimone: @darkk, I see that the regress tests fail. This is because `create_evdns_base()` is a template that allows to mock `evdns_base_free()` and the regress tests mock it and make sure that, for each error path, the function is called.

The mocking itself is implemented via `MK_MOCK` in `<measurement_kit/common/mock.hpp>`.

I suggest to pass the deleter as an explicit lambda to the `std::unique_ptr` such that the `evdns_base_free` that is picked up inside `create_evdns_base()` could be mocked in the unit test.

I'd also check the coverage and, if it's reduced after this commit, it would be nice to add more tests.

bassosimone:
@darkk thanks for the diff and your help. Overall the diff looks good to me. I've annotated it with explanatory comments. The only show stopper I see is that regress tests fails, as commented inline.

src/libmeasurement_kit/libevent/dns_impl.hpp
darkk: I think, same logic applies like _do not zero the memory if it's not used_. That's why we have various _secure_memset_ functions.

src/libmeasurement_kit/libevent/dns_impl.hpp
darkk: > evdns_base_free that is picked up inside create_evdns_base() could be mocked in the unit test

I'd rather say, that checking for exception being thrown is enough for tests and validation of memory being cleaned up should be done by `valgrind` or alike stuff. Actually, `make check-am-valgrind` highlights several tests being red :)

darkk:
@bassosimone there is another stuff that worries me. See, I've hardcoded "53" as a port number :)

That's bad thing to do for sure, but the question is — should we pass `:53` as a string to `dns/nameserver` or should we set the port as a separate variable.

Android code was having code like `nameserver.as_string() + ":53"` and that's wrong thing to do with IPv6 as you can convert `fd00::aaaa:bbbb:cccc:dddd, port=53` to `fd00::aaaa:bbbb:cccc:dddd:53` that's valid IPv6 address as well.

So we have several options:
- set resolver port as a separate options like `dns/nameserver-port`
- fix all MK users that deal with IPv6 so they'll properly escape IPv6 address like `[fd00:aaaa:bbbb:cccc:dddd%en-wtf]:53` and distinguish IPv4 from IPv6 as IPv4 should not be escaped.

What do you consider a way to go?

src/libmeasurement_kit/libevent/dns_impl.hpp
bassosimone: > [...] checking for exception being thrown is enough for tests and validation of memory being cleaned up should be done by valgrind or alike stuff

On macOS there is no Valgrind (and even when there was a Valgrind it was not working correctly). This is the reason why there are also cases where we mock the destructor. For this reason, I do see a case for having that.

Regarding how it will be possible to implement that, you can either enhance the structure you already created to become a template receiving a deleter, or do use a lambda. The latter would be my preferred choice because it is more compact han defining a custom type for the deleter. I've written the following proof of concept that compiles with clang++ on macOS and shows how to do that:

```C++
#include <memory>
#include <iostream>

int main() {
    std::unique_ptr<int, void (*)(int *)> foo{new int{21}, [](int *x) {
        std::cout << "called\n";
        delete x;
    }};
}
```

By the way, when I mentioned to you that `shared_ptr` was more cool I actually meant that you can rewrite the above example with `std::shared_ptr<int>` rather than `std::unique_ptr<int, /*bla bla*/>`. It would be interesting to understand why, but for now I have not spent time trying to understand that.

> Actually, make check-am-valgrind highlights several tests being red :)

Which ones? Should I check this myself? Tests work and Valgrind does pass on master, and looking at the output of Valgrind as I see it on travis-ci, it seems to me the `evbuffer_free()` change in this patch broke the tests.

bassosimone:
> @bassosimone there is another stuff that worries me

Yep, I agree that this is a source of concern.

> What do you consider a way to go?

On top of my head, I'd say that, whenever possible, we should have separate hostname (or address and protocol family) and port and pass them firstly to some DNS resolver &mdash; if needed &mdash; to obtain the corresponding address and protocol family (possibly addresses and protocol families) and then pass the address+protocol_family tuple to getaddrinfo as you are doing in this patch to obtain a `struct sockaddr` (actually a `sockaddr_storage`) and pass it to the relevant socket-level functions.

Doing this kind of restructuring is probably not super easy always, but it seems to me that it could be the most robust way to structure the code, given the things that can go wrong with IPv6 (including, for example, the scope based link-local ipv6 addressing issue you fixed in this pull request).

In the context of your question, I'd say this means passing the port separately.

Does this make sense to you?

src/libmeasurement_kit/libevent/dns_impl.hpp
darkk: You're reading it incorrectly. I mean, that in the following case stack memory will not be zeroed

```
void foo() {
 Struct s;
 mess_with(&s);
 s = {}; // may be optimised out
}
```

The same reasoning applies to compiler dropping `memset`.

bassosimone: Okay, then, this is what I understood in the first place. I see the rationale for that happening, but I am also wondering whether a compiler is allowed to drop the semantic of this statement:

```C++
void foo() {
    Struct s{};
}
```

As I see that this discussion could become bike-shedding easily, I will do my research. Thanks.

src/libmeasurement_kit/cmdline/dns_query.cpp
```diff
@@ -12,18 +12,22 @@ namespace cmdline {
 namespace dns_query {
 
 static const char *kv_usage =
-    "usage: measurement_kit dns_query [-v] [-c class] [-t type] domain\n";
+    "usage: measurement_kit dns_query [-N nameserver] [-v] [-c class] [-t type] domain\n";
```
bassosimone: *TODO*: here we should also add support for specifying the port

src/libmeasurement_kit/libevent/dns_impl.hpp
darkk: > Valgrind does pass on master

`make check-am-valgrind` produces at least following failures for me both on "master" (67441ddee9bc1e6057982efa7757522844b34426) and this branch:

```
FAIL: test/mlabns/mlabns
FAIL: test/ndt/ndt_test
FAIL: test/ndt/run
FAIL: test/http/request
FAIL: test/ooni/web_connectivity
FAIL: test/net/connect
```

All these tests fail with alike assertion failures somewhere near `EC_KEY_generate_key`:
```
vex: the `impossible' happened:
   isZeroU
vex storage: T total 1462491848 bytes allocated
vex storage: P total 640 bytes allocated
valgrind: the 'impossible' happened:
   LibVEX called failure_exit().
```

That's a known bug in valgrind+openssl: [1](https://bugs.kde.org/show_bug.cgi?id=356393), [2](https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/1574437), [3](https://bugzilla.redhat.com/show_bug.cgi?id=1326024).

Other tests are OK. I was using Valgrind-3.11.0 and LibVEX

src/libmeasurement_kit/ooni/templates.cpp
```diff
@@ -21,13 +21,15 @@ void dns_query(Var<Entry> entry, dns::QueryType query_type,
     std::string resolver_hostname, nameserver_part;
     std::stringstream nameserver_ss(nameserver);
 
+    // TODO: Fix IPv6 `nameserver`, use google's [2001:4860:4860::8888]:53 for testing.
```
darkk: IMHO, merging without further fixes is okayish way to go.

That's especially true, given the uncertain future of libevent-based DNS parser that will be replaced with c-ares. Also c-ares will have alike issue as well, as we've deduced from the API that takes only `in_addr` and `in6_addr`. :)

src/libmeasurement_kit/libevent/dns_impl.hpp
bassosimone: I've seen that using `struct addrinfo ainfo = {}` causes a g++ warning where the compiler complains that fields are not initialized (specifically with `-Wextra` also `-Wmissing-field-initializers` is enabled by g++ and the warning says that the initializer is missing). Yet the assembly contains a `req stosq` instruction that IIRC should clear the buffer:

```C++
#include <sys/socket.h>
#include <netinet/in.h>
#include <netdb.h>

int main() {
    addrinfo hints{};
}
```

and (snipped assembly that should mean "for 6 repetitions, store 0 into memory pointed to by rdi incrementing rdi by 8 each time"):

```amd64
        leaq    -48(%rbp), %rdx
        movl    $0, %eax
        movl    $6, %ecx
        movq    %rdx, %rdi
        rep stosq
```

This seems similar to the assembly generated by memset().

Regarding `explicit_memset()`, the issue arises when you want to clear a private key once you have finished using it, because there is an optimization calling "dead store elimination" where the compiler removes write operations to areas of memory that are no longer used after such store.

This is a concern for private keys because they are cleared after use, but not a concern when you are initializing memory for using it, because that does not trigger the optimization.

Interestingly, [explicit memset is implemented by calling a "weak" function after zeroing the memory](https://github.com/libressl-portable/openbsd/blob/master/src/lib/libc/string/explicit_bzero.c), to fool the optimizer and avoid triggering dead store elimination (and there is also [a discussion on when this approach could fail](https://github.com/libressl-portable/openbsd/issues/5)).

Bottom line: let's just use `memset()`.

src/libmeasurement_kit/ooni/templates.cpp
bassosimone: > IMHO, merging without further fixes is okayish way to go.

Okay, then, I'll make a TODO list for merging which should include not forgetting about all the issues raised during this pull request (and most especially the port) and proceed.

> Also c-ares will have alike issue as well [...]

True. My current c-ares prototype however operates at a low level and does not use c-ares code to open the UDP socket, so we have more control. Thus, I can use `getaddrinfo()` like you did in this pull request and that should workaround the issue with scoped IPv6 addresses.

We may want to invest some time to fix the issue for libevent and/or c-ares though, don't we?

src/libmeasurement_kit/libevent/dns_impl.hpp
darkk: > a g++ warning

What g++ version? I can't reproduce that. I wonder if it's the difference between `hints{}` and `hints = {}`.

bassosimone:
Merged, thanks!

src/libmeasurement_kit/libevent/dns_impl.hpp
bassosimone: > What g++ version?

4.9.4 (it's quite old, unfortunately it's the only g++ on VoidLinux).

For completeness, that's what I see:

```
$ g++ --version
g++ (GCC) 4.9.4
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
$ cat prova.cpp
#include <sys/socket.h>
#include <netinet/in.h>
#include <netdb.h>

int main() {
    addrinfo hints{};
}
$ g++ -Wall -Wextra -std=c++11 -S prova.cpp
prova.cpp: In function 'int main()':
prova.cpp:6:20: warning: missing initializer for member 'addrinfo::ai_flags' [-Wmissing-field-initializers]
     addrinfo hints{};
                    ^
prova.cpp:6:20: warning: missing initializer for member 'addrinfo::ai_family' [-Wmissing-field-initializers]
prova.cpp:6:20: warning: missing initializer for member 'addrinfo::ai_socktype' [-Wmissing-field-initializers]
prova.cpp:6:20: warning: missing initializer for member 'addrinfo::ai_protocol' [-Wmissing-field-initializers]
prova.cpp:6:20: warning: missing initializer for member 'addrinfo::ai_addrlen' [-Wmissing-field-initializers]
prova.cpp:6:20: warning: missing initializer for member 'addrinfo::ai_addr' [-Wmissing-field-initializers]
prova.cpp:6:20: warning: missing initializer for member 'addrinfo::ai_canonname' [-Wmissing-field-initializers]
prova.cpp:6:20: warning: missing initializer for member 'addrinfo::ai_next' [-Wmissing-field-initializers]
prova.cpp:6:14: warning: unused variable 'hints' [-Wunused-variable]
     addrinfo hints{};
              ^
```
> I wonder if it's the difference [...]

I've checked, but even with `= {}` I obtain the same warnings.

Issue #742, opened 2016-07-31
dns: add support for CNAME queries


bassosimone:

bassosimone:
Support added through the system resolver in #918

PR #918, opened 2016-11-11
Resolve hosts using the system resolver

AntonioLangiu:
[comment added by @bassosimone: This is a very useful enhancement that allows us to use the system resolver for performing ordinary resolution tasks so to be closer to what an user would see]


include/measurement_kit/dns/dns.hpp
```diff
@@ -26,6 +26,7 @@ MK_DEFINE_ERR(MK_ERR_DNS(13), UnsupportedClassError, "")
 MK_DEFINE_ERR(MK_ERR_DNS(14), InvalidNameForPTRError, "")
 MK_DEFINE_ERR(MK_ERR_DNS(15), ResolverError, "")
 MK_DEFINE_ERR(MK_ERR_DNS(16), UnsupportedTypeError, "")
+MK_DEFINE_ERR(MK_ERR_DNS(17), InvalidDnsEngine, "")
```
bassosimone: <3

src/libmeasurement_kit/dns/query.cpp
```diff
@@ -15,7 +16,14 @@ void query(
...
         Var<Reactor> reactor,
         Var<Logger> logger) {
-    libevent::query(dns_class, dns_type, name, cb, settings, reactor, logger);
+    std::string engine = settings.get("dns/engine", std::string("libevent"));
+    if (engine == "libevent") {
+        libevent::query(dns_class, dns_type, name, cb, settings, reactor, logger);
+    } else if (engine == "system") {
+        system_resolver(dns_class, dns_type, name, cb, settings, reactor, logger);
+    } else {
+        cb(InvalidDnsEngine(), nullptr);
```
bassosimone: (rationale: I hate when libraries call *callbacks* ahead in time. I suppose callbacks should be called later.

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
```
bassosimone: Nitpicking: it seems to me one line from the usual header is missing here.

bassosimone: More nitpicking: I presume it would be more obvious if this file was called `system_resolver.hpp`

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+    QueryClass dns_class;
+    QueryType dns_type;
+    std::string name;
+    Callback<Error, Var<Message>> cb;
+    Settings settings;
+    Var<Reactor> reactor;
+    Var<Logger> logger;
+
+    struct addrinfo hints;
+    struct addrinfo *servinfo = nullptr;
```
bassosimone: Style nit: you can omit `struct` when you're in C++

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+// information on the copying conditions.
+
+#include <measurement_kit/common.hpp>
+#include <measurement_kit/dns.hpp>
+
+#include <thread>
+/// XXX use versions compatible with windows
+#include <netdb.h>
+#include <sys/socket.h>
+#include <sys/types.h>
```
bassosimone: The correct way to go about portability, in my opinion, is the following: you should make sure all Unix-specific headers you want to include are in `<measurement_kit/portable/>` and that they're pulled by `<measurement_kit/aaa_base.hpp>`. Then, you need to check what Windows headers are required to achieve a similar task and &mdash; when we'll merge the Windows patch &mdash; we'll make sure that those headers are also listed and pulled by `aaa_base.hpp` by using `#ifdef WIN32` when required (from a cursory analysis it seems to me that `netdb.h` is not part of `portable` headers and that `winsock2.h` is all you need on Windoze to call getaddrinfo).

As a side note, I also see the possibility of tapping into libevent's `getaddrinfo()` but I'd rather not do so as this level because it seems to me that libevent's `getaddrinfo()` only concerns itself with portability towards old systems of which we do not care about, plus I'd rather keep the surface of code depending on libevent small (and this is the rationale behind having a `libevent` folder).

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+        cb = c;
+        settings = s;
+        reactor = r;
+        logger = l;
+        memset(&hints, 0, sizeof(hints));
+    }
+
+    ~ResolverContext() {
+        if (servinfo != nullptr) {
+            freeaddrinfo(servinfo);
```
bassosimone: In #915, @darkk introduced a pattern based on `std::unique_ptr` to deal with automatic destruction that I think we can also reuse in this case. Can we use this pattern here as well?

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+    Query query;
+    ctx->hints.ai_flags = AI_ADDRCONFIG;
+    ctx->hints.ai_socktype = SOCK_STREAM;
+    ctx->hints.ai_protocol = 0;
+    ctx->hints.ai_canonname = NULL;
+    ctx->hints.ai_addr = NULL;
+    ctx->hints.ai_next = NULL;
+
+    if (dns_class != QueryClassId::IN) {
+        cb(UnsupportedClassError(), nullptr);
```
bassosimone: Here, in case of error, you're leaking `ctx`. I believe the same happens also in other error cases.

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+                            Var<Logger> logger) {
+    ResolverContext *ctx = new ResolverContext(dns_class, dns_type, name, cb,
+                                               settings, reactor, logger);
+    Query query;
+    ctx->hints.ai_flags = AI_ADDRCONFIG;
+    ctx->hints.ai_socktype = SOCK_STREAM;
+    ctx->hints.ai_protocol = 0;
+    ctx->hints.ai_canonname = NULL;
+    ctx->hints.ai_addr = NULL;
+    ctx->hints.ai_next = NULL;
```
bassosimone: Two thoughts here: 1) you should probably use `nullptr` as opposed to `NULL`; 2) when you need to zero a structure and be sure every field is zeroed, I'd rather use `memset` (well, to be honest, I'd rather use `bzero` but unfortunately it's not portable to all systems while `memset` is).

src/libmeasurement_kit/dns/resolver.hpp
bassosimone: A possible way to go about this is to wrap `ctx` into a smart pointer and then explicitly `release()` the smart pointer when we're done with initialization of the object (I find this proposed pattern more safe because forgetting to release in the "usual" path leads us to immediately notice such programming error via Valgrind.)

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+    if (dns_class != QueryClassId::IN) {
+        cb(UnsupportedClassError(), nullptr);
+        return;
+    }
+
+    if (dns_type == QueryTypeId::A) {
+        ctx->hints.ai_family = AF_UNSPEC;
```
bassosimone: Actually here you should use `PF_INET` according to me

bassosimone: (Or `AF_INET`, they're actually the same.)

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+    query.type = dns_type;
+    query.qclass = dns_class;
+    query.name = name;
+
+    ctx->message->queries.push_back(query);
+
+    std::thread res_thread(resolve_async, ctx);
+    /// XXX study a way to mantain a reference to the thread
+    res_thread.detach();
```
bassosimone: Some thoughts on the code in here:

1) I'd rather use `std::launch::async()` because it's higher level than `std::thread` and I think it's better to use the highest layer API that does the job for us

2) [however?] I imagine in the future we will have a queue for running "tasks" like this one, because it's not super robust to possible spawn N threads to do N concurrent DNS resolutions... I'm not saying we should be doing this right now but perhaps we should take this possibility into account

3) I'd rather use a lambda here instead of creating a separate function (but I guess this is a matter of personal taste and one could also argue against using a lambda... yet, probably when we have a task queue one simple way to implement it would be "call this lambda in a separate thread and callback in the main thread when done")

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+        if (servinfo != nullptr) {
+            freeaddrinfo(servinfo);
+        }
+    }
+};
+
+static inline void resolve_async(ResolverContext *ctx) {
+    Callback<Error, Var<Message>> callback = ctx->cb;
+    Var<Message> message_copy;
+    struct addrinfo *p;
```
bassosimone: (You can omit `struct` in C++)

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+    if (error) {
+        // check the error variable and return the correct error
+        ctx->logger->warn(gai_strerror(error));
+        ctx->reactor->call_soon(
+            [callback]() { callback(ResolverError(), nullptr); });
+        delete ctx;
+        return;
+    }
+
+    if (ctx->servinfo == nullptr) {
```
bassosimone: Two comments here:

1) it seems to me that `servinfo` is only needed in the scope of this function. If so, I'd restructure the code such that we can have this as a variable living on the stack (so to have minimal scope for variables)

2) my understanding is that `servinfo` must not be null after `getaddrinfo()`, is this correct? If so, probably here we can just `assert()`

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+        ctx->reactor->call_soon(
+            [callback]() { callback(ResolverError(), nullptr); });
+        delete ctx;
+        return;
+    }
+
+    void *addr_ptr;
+    char address[128];
+    std::vector<Answer> answers;
+    for (p = ctx->servinfo; p != nullptr; p = p->ai_next) {
```
bassosimone: I'd say `for (addrinfo *p = ctx->servinfo; ...`

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+        } else {
+            ctx->reactor->call_soon(
+                [callback]() { callback(ResolverError(), nullptr); });
+            delete ctx;
+            return;
+        }
+        if (inet_ntop(p->ai_family, addr_ptr, address, sizeof(address)) ==
+            NULL) {
+            ctx->logger->warn("dns: unexpected inet_ntop failure");
+            throw std::runtime_error("Unexpected inet_ntop failure");
```
bassosimone: (BTW, my impression is that some bits of code in here are copied from our existing resolver. I guess by looking at this code we're also finding issues with existing code, which should probably be fixed as well, isn't it?)

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,152 @@
...
+        }
+        if (p->ai_family == AF_INET) {
+            answer.ipv4 = std::string(address);
+        } else if (p->ai_family == AF_INET6) {
+            answer.ipv6 = std::string(address);
+        }
+        answers.push_back(answer);
+    }
+    ctx->message->answers = answers;
+    message_copy = ctx->message;
```
bassosimone: Is it correct to say that the scope of message is perhaps only this function?

test/dns/resolver.cpp
```diff
@@ -0,0 +1,39 @@
...
+                    for (auto answer : message->answers) {
+                        if (answer.ipv6 == "2a00:1450:4001:815::200e") {
+                                found = true;
+                        }
+                    }
+                    REQUIRE(found);
+                    break_loop();
+                    }, {}, Reactor::global(), Logger::global());
+            });
+}
```
bassosimone: If we can mock the various call, I think we can also be satisfied with full coverage with IPv6 failing on travis (and we can perhaps add logic to deal with failures in the travis case?)

src/libmeasurement_kit/dns/resolver.hpp
AntonioLangiu: yes, you are right!

src/libmeasurement_kit/dns/resolver.hpp
AntonioLangiu: Yes the address conversion part is similar to the one we already have in libevent::query

src/libmeasurement_kit/dns/resolver.hpp
AntonioLangiu: Yes, I did it to avoid refering explicitly to the context or pass a context reference to the callback!

src/libmeasurement_kit/dns/resolver.hpp
AntonioLangiu: Yeah, I know! I already used memset in the `ResolverContext` constructor to set the hints to 0, but it was to be more explicit! I'm removing all this assignations.

src/libmeasurement_kit/dns/resolver.hpp
bassosimone: I was asking another question, but after reading one more time the diff, I've understood that my question was based on the wrong assumption, sorry :)

src/libmeasurement_kit/dns/resolver.hpp
bassosimone: Going to remove manually this comment since it was addressed, thanks!

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,146 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+
+#include <measurement_kit/common.hpp>
+#include <measurement_kit/dns.hpp>
+#include <measurement_kit/portable/sys/socket.h>
+#include <measurement_kit/portable/sys/types.h>
+
+#include <assert.h>
```
bassosimone: Can you use `<cassert>` instead?

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,146 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+
+#include <measurement_kit/common.hpp>
```
bassosimone: No need to include `common.hpp` since it's already included by `dns.hpp`

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,146 @@
...
+#include <measurement_kit/dns.hpp>
+#include <measurement_kit/portable/sys/socket.h>
+#include <measurement_kit/portable/sys/types.h>
```
bassosimone: No need to include stuff in `portable` directly because they're included by `common.hpp`

src/libmeasurement_kit/dns/resolver.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include <assert.h>
+#include <future>
+
+#include <netdb.h>
```
bassosimone: You should add the `portable/netdb.h` file and make sure it's pulled by `common/aaa_base.hpp`

bassosimone:
Thanks, @AntonioLangiu, added some more comments.

test/dns/resolver
```diff
@@ -0,0 +1,228 @@
+#! /bin/bash
```
bassosimone: @AntonioLangiu you should not commit this file

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,146 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+
+#include <measurement_kit/dns.hpp>
+#include <measurement_kit/portable/sys/socket.h>
+#include <measurement_kit/portable/sys/types.h>
+#include <measurement_kit/portable/netdb.h>
```
bassosimone: These three files should not be here. They should be included by `measurement_kit/common/aaa_base.hpp`

bassosimone:
How annoying, it seems that we have IPv6 issues also after stopping using containers. Perhaps, we can discuss this further tomorrow and see if we can nail down the problem. Thanks!

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+    ~ResolverContext() {
+        if (servinfo != nullptr) {
+            freeaddrinfo(servinfo);
+        }
+    }
+};
+
+template<MK_MOCK(getaddrinfo), MK_MOCK(inet_ntop)>
+static inline void resolve_async(ResolverContext *context) {
```
bassosimone: You don't need to declare templates `static inline`

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+    std::unique_ptr<ResolverContext> ctx(context);
+    Callback<Error, Var<Message>> callback = ctx->cb;
+    Var<Message> message_copy;
```
bassosimone: I think you can slightly simplify by writing `message_copy = ctx->message` (perhaps in that case you can also call directly `message` the variable, since it's not a real copy but another reference to the same object?)

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+    int error = getaddrinfo(ctx->name.c_str(), nullptr, &(ctx->hints),
+                            &(ctx->servinfo));
```
bassosimone: It seems to me that `servinfo` only makes sense in the context of this function. So you should probably declare it here on the stack and you can perhaps use a `unique_ptr` to guarantee that it is always destroyed.

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+};
+
+template<MK_MOCK(getaddrinfo), MK_MOCK(inet_ntop)>
+static inline void resolve_async(ResolverContext *context) {
+    std::unique_ptr<ResolverContext> ctx(context);
+    Callback<Error, Var<Message>> callback = ctx->cb;
+    Var<Message> message_copy;
+
+    int error = getaddrinfo(ctx->name.c_str(), nullptr, &(ctx->hints),
```
bassosimone: It would suffice to write `&ctx->hints`

src/libmeasurement_kit/dns/system_resolver.hpp
bassosimone: It would suffice to write `&ctx->servinfo`

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+                            &(ctx->servinfo));
+    if (error) {
+        // check the error variable and return the correct error
+        ctx->logger->warn(gai_strerror(error));
```
bassosimone: I'd rewrite this as: `ctx->logger->warn("getaddrinfo failed: %s", gai_strerror(error))`

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+        ctx->reactor->call_soon(
+            [callback]() { callback(ResolverError(), nullptr); });
+        return;
+    }
+
+    assert(ctx->servinfo != nullptr);
+
+    void *addr_ptr;
+    char address[128];
+    std::vector<Answer> answers;
```
bassosimone: Can you write directly in `message->answers`?

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+        if (p->ai_family == AF_INET) {
+            answer.type = QueryTypeId::A;
+            addr_ptr = &((sockaddr_in *)p->ai_addr)->sin_addr;
+        } else if (p->ai_family == AF_INET6) {
+            answer.type = QueryTypeId::AAAA;
+            addr_ptr = &((sockaddr_in6 *)p->ai_addr)->sin6_addr;
+        } else {
+            ctx->reactor->call_soon(
+                [callback]() { callback(ResolverError(), nullptr); });
+            return;
```
bassosimone: My understanding of how `getaddrinfo()` works &mdash; correct me if I'm wrong &mdash; is that in output we can only have `AF_INET` or `AF_INET6`; then, we can probably `assert()` for family being either `PF_INET` or `PF_INET6`.

This would allow us to simplify this chain of ifs. We can probably write assignements to `answer.type` and to `addr_ptr` as short-circuit evaluation statements.

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+        if (inet_ntop(p->ai_family, addr_ptr, address, sizeof(address)) ==
+            NULL) {
```
bassosimone: Please, use `nullptr`rather than `NULL`.

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+            ctx->logger->warn("dns: unexpected inet_ntop failure");
+            throw std::runtime_error("Unexpected inet_ntop failure");
```
bassosimone: I don't think `inet_ntop` is going to fail if the buffer is large enough. Yet, for consistency, we may want to call the callback and return an error in case it fails, rather than throwing. Does it make sense?

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+            answer.ipv4 = std::string(address);
+        } else if (p->ai_family == AF_INET6) {
+            answer.ipv6 = std::string(address);
+        }
+        answers.push_back(answer);
+    }
+    ctx->message->answers = answers;
+    message_copy = ctx->message;
+    ctx->reactor->call_soon(
+        [callback, message_copy]() { callback(NoError(), message_copy); });
```
bassosimone: Note: if you pass `[=]` only the variables mentioned in the closure are actually copied.

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+template<MK_MOCK(getaddrinfo), MK_MOCK(inet_ntop)>
+inline void system_resolver(QueryClass dns_class, QueryType dns_type,
```
bassosimone: You don't need to specify `inline` here

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+                            std::string name, Callback<Error, Var<Message>> cb,
+                            Settings settings, Var<Reactor> reactor,
+                            Var<Logger> logger) {
+    std::unique_ptr<ResolverContext> ctx(new ResolverContext(dns_class, dns_type, name, cb,
+                                               settings, reactor, logger));
+    Query query;
+    ctx->hints.ai_flags = AI_ALL;
```
bassosimone: From `getaddrinfo(3)` on macOS:

```
    The AI_ALL bit without the AI_V4MAPPED bit is ignored.
```

Here instead I'd specify `AI_CANONNAME` and will also get and store it in the response (I recall this is a piece of information to which @hellais was interested and now I see we also have issue #742 for that)

test/dns/system_resolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+static const char *null_inet_ntop(int, const void *, char *, socklen_t) {
+    return nullptr;
+}
+
+static int null_getaddrinfo(const char *, const char *, const struct addrinfo *, struct addrinfo **) {
+    return 10;
+}
+
+TEST_CASE("the system resolver is able to resolve an ipv4 address") {
```
bassosimone: You should wrap this test case and all test cases where we use the network with

```C++
#ifdef ENABLE_INTEGRATION_TESTS
// test using the network ...
#endif
```

test/dns/system_resolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+    loop_with_initial_event([]() {
+            system_resolver("IN", "A", "www.neubot.org", [](Error e, Var<Message> message) {
+                    REQUIRE(!e);
+                    REQUIRE(message->answers.size() == 1);
+                    REQUIRE(message->answers[0].ipv4 == "130.192.16.172");
```
bassosimone: After you have implemented canonical name, remember to also add a test for it

test/dns/system_resolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+            system_resolver<null_getaddrinfo>("IN", "A", "www.neubot.org", [](Error e, Var<Message>) {
+                    REQUIRE(e);
+                    break_loop();
+                    }, {}, Reactor::global(), Logger::global());
+            });
+}
+
+TEST_CASE("the system resolver can handle a inet_ntop error") {
+    loop_with_initial_event([]() {
+            system_resolver<getaddrinfo, null_inet_ntop>("IN", "A", "invalid_site", [](Error e, Var<Message>) {
```
bassosimone: Beware of line length. Please run `clang-format` on the sources.

test/dns/system_resolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+                    REQUIRE(message->answers[0].ipv4 == "130.192.16.172");
+                    break_loop();
+                    }, {}, Reactor::global(), Logger::global());
+            });
+}
+
+TEST_CASE("the system resolver returns an error with an invalid_site") {
+    loop_with_initial_event([]() {
+            system_resolver("IN", "A", "invalid_site", [](Error e, Var<Message>) {
+                    REQUIRE(e);
```
bassosimone: And also do the same in all tests below, please

src/libmeasurement_kit/dns/system_resolver.hpp
AntonioLangiu: We do this also in the libevent dns resolver function, so we should change it also there.

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,141 @@
...
+    Callback<Error, Var<Message>> callback = ctx->cb;
+    Var<Message> message_copy;
+
+    int error = getaddrinfo(ctx->name.c_str(), nullptr, &(ctx->hints),
+                            &(ctx->servinfo));
+    if (error) {
+        // check the error variable and return the correct error
+        ctx->logger->warn(gai_strerror(error));
+        ctx->reactor->call_soon(
+            [callback]() { callback(ResolverError(), nullptr); });
```
AntonioLangiu: I think we should discuss how to manage dns errors. Should I create new errors for the system_resolver or use the equivalent errors already used for the libevent resolver? I was planning to do it in another PR where I will integrate the system resolver with the library, add all the required tests and refactor the dns tests in:

    test/dns/query to test the dns functionalities
    test/dns/system_resolver to test the system_resolver function
    test/dns/libevent_resolver to test the libevent resolver

src/libmeasurement_kit/dns/system_resolver.hpp
AntonioLangiu: yes!

src/libmeasurement_kit/dns/system_resolver.hpp
AntonioLangiu: yes!

src/libmeasurement_kit/dns/system_resolver.hpp
AntonioLangiu: thanks!

src/libmeasurement_kit/dns/system_resolver.hpp
bassosimone: > Should I create new errors for the system_resolver or use the equivalent errors already used for the libevent resolver?

Fair question. Are evdns and getaddrinfo errors semantically aligned? (I.e. can you find a 1:1 mapping from evdns errors to getaddrinfo errors?)

src/libmeasurement_kit/dns/system_resolver.hpp
bassosimone: > We do this also in the libevent dns resolver function, so we should change it also there.

Yes. I remember that the argument for throwing was "it should not happen". However, if something goes wrong and a mobile app crashes because we've thrown it's a pain to debug it. So, probably it would be better to return an error. On the other hands, things that really should not happen (e.g. you read up to 1,024 bytes and the result is 2,048 read bytes) should be assertions.

bassosimone:
I've moved this in release-0.4.0 again, because it seems me we're converging quickly enough to be able to achieve this goal and merge before we make next release. 👍

include/measurement_kit/dns/dns.hpp
```diff
@@ -26,6 +26,21 @@ MK_DEFINE_ERR(MK_ERR_DNS(13), UnsupportedClassError, "")
...
+MK_DEFINE_ERR(MK_ERR_DNS(20), InvalidHintsValue, "")
+MK_DEFINE_ERR(MK_ERR_DNS(21), NonRecoverableFailure, "")
+MK_DEFINE_ERR(MK_ERR_DNS(22), NotSupportedAIFamily, "")
+MK_DEFINE_ERR(MK_ERR_DNS(23), MemoryAllocationFailure, "")
+MK_DEFINE_ERR(MK_ERR_DNS(24), HostOrServiceNotProvidedOrNotKnown, "")
+MK_DEFINE_ERR(MK_ERR_DNS(25), ArgumentBufferOverflow, "")
+MK_DEFINE_ERR(MK_ERR_DNS(26), UnknownResolvedProtocol, "")
+MK_DEFINE_ERR(MK_ERR_DNS(27), NotSupportedServname, "")
+MK_DEFINE_ERR(MK_ERR_DNS(28), NotSupportedAISocktype, "")
+MK_DEFINE_ERR(MK_ERR_DNS(29), InetNtopFailure, "")
```
bassosimone: Could you add the suffix `Error` to all errors?

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,174 @@
...
+                    Callback<Error, Var<Message>> c, Settings s, Var<Reactor> r,
+                    Var<Logger> l) {
+        dns_class = dns_c;
+        dns_type = dns_t;
+        name = n;
+        cb = c;
+        settings = s;
+        reactor = r;
+        logger = l;
+        memset(&hints, 0, sizeof(hints));
```
bassosimone: I think you can obtain the same effect by saying `hints = {}` above

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,174 @@
...
+    Var<Message> message = ctx->message;
+    addrinfo *servinfo_p = nullptr;
+
+    int error =
+        getaddrinfo(ctx->name.c_str(), nullptr, &ctx->hints, &servinfo_p);
+    std::unique_ptr<addrinfo, std::function<void(addrinfo *)>> servinfo(
+        servinfo_p, [](addrinfo *servinfo) {
+            if (servinfo != nullptr)
+                freeaddrinfo(servinfo);
+        });
```
bassosimone: Also, could you add { } even though there is just one statement in the if branch?

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,174 @@
...
+        case EAI_SOCKTYPE:
+            resolver_error = NotSupportedAISocktype();
+            break;
+        default:
+            resolver_error = ResolverError();
+        }
+        ctx->reactor->call_soon([=]() { callback(resolver_error, nullptr); });
+        return;
+    }
+    assert(servinfo != nullptr);
```
bassosimone: Since `serveinfo` is a smart pointer it could also be written as `assert(!!servinfo)`

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,174 @@
...
+        assert(p->ai_family == AF_INET or p->ai_family == AF_INET6);
+        if (p->ai_family == AF_INET) {
+            answer.type = QueryTypeId::A;
+            addr_ptr = &((sockaddr_in *)p->ai_addr)->sin_addr;
+        } else if (p->ai_family == AF_INET6) {
+            answer.type = QueryTypeId::AAAA;
+            addr_ptr = &((sockaddr_in6 *)p->ai_addr)->sin6_addr;
+        }
+        if (p->ai_canonname != nullptr) {
+            answer.hostname = p->ai_canonname;
```
bassosimone: Question: is `hostname` reserved for this purpose in the structure, right?

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,174 @@
...
+}
+
+template <MK_MOCK(getaddrinfo), MK_MOCK(inet_ntop)>
+void system_resolver(QueryClass dns_class, QueryType dns_type, std::string name,
+                     Callback<Error, Var<Message>> cb, Settings settings,
+                     Var<Reactor> reactor, Var<Logger> logger) {
+    std::unique_ptr<ResolverContext> ctx(new ResolverContext(
+        dns_class, dns_type, name, cb, settings, reactor, logger));
+    Query query;
+    ctx->hints.ai_flags = AI_ALL | AI_V4MAPPED;
```
bassosimone: What does `AI_V4MAPPED` give us?

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,174 @@
...
+        return;
+    }
+
+    if (dns_type == QueryTypeId::A) {
+        ctx->hints.ai_family = AF_INET;
+    } else if (dns_type == QueryTypeId::AAAA) {
+        ctx->hints.ai_family = AF_INET6;
+    } else if (dns_type == QueryTypeId::CNAME) {
+        ctx->hints.ai_family = AF_UNSPEC;
+        ctx->hints.ai_flags |= AI_CANONNAME;
```
bassosimone: Question: my understanding is that we always want to know the canonical name. Perhaps we should always pass in `AI_CANONNAME`, or do we want the user to specify it explicitly?

bassosimone:
@AntonioLangiu thanks! I've committed a simple fix that should repair the build, merged again with master, and added a couple more comments. (I can take care of those comments, since they are mostly trivial, as part of merging the pull request, if you wish!)

src/libmeasurement_kit/dns/system_resolver.hpp
AntonioLangiu: yes 👍

src/libmeasurement_kit/dns/system_resolver.hpp
AntonioLangiu: In the first implementation we always got the canonical name, but then I say that QueryType has the CNAME query possibility and I changed then implementation to have a specific query for it! You can revert the commit and get the old behavior.

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,174 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+
+#include <measurement_kit/dns.hpp>
+
+#include <cassert>
+#include <future>
+
+#include <iostream>
```
AntonioLangiu: No

src/libmeasurement_kit/dns/system_resolver.hpp
AntonioLangiu: It should return (used with AI_ALL) the ipv6 address and the ipv4 to ipv6 mapped addresses when we ask for AAAA resolution.

src/libmeasurement_kit/dns/system_resolver.hpp
bassosimone: Ok, then, I'll remove it as part of merging

src/libmeasurement_kit/dns/system_resolver.hpp
bassosimone: Good!

bassosimone:
I am going to merge, even though circleci fails. I have additionally verified on my VoidLinux system using Valgrind that there are no memory errors inside of the system_resolver test case.

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -0,0 +1,184 @@
...
+    assert(sip != nullptr);
+    addrinfo_uptr servinfo(sip);
+
+    void *addr_ptr;
+    char abuf[128];
+    for (addrinfo *p = servinfo.get(); p != nullptr; p = p->ai_next) {
+        Answer answer;
+        answer.name = ctx->name;
+        answer.qclass = ctx->dns_class;
+        assert(p->ai_family == AF_INET or p->ai_family == AF_INET6);
```
darkk: Where does `or` come from? I'm a bit surprised to see `or` instead of usual `||`

bassosimone: It is part of the C++ standard: `or` is the alternative representation of `||`

bassosimone: See, for example, http://en.cppreference.com/w/cpp/language/operator_alternative

darkk: Thanks!
God bless trigraphs!

PR #920, opened 2016-11-16
Add c-ares based DNS resolver

bassosimone:
At the moment of opening this pull request, what we have here is still highly experimental work that I am publishing to make sure I'm not losing the branch and to show the initial prototype.

As such, this is flagged as "on hold"


bassosimone:
Obsoleted by #966

PR #925, opened 2016-11-16
Switch travis builds to VM

AntonioLangiu:
We were using the container based build system that doesn't support ipv6. 
To be able to test ipv6 we need to switch to the VM based build system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ after_success:
 
 matrix:
   include:
-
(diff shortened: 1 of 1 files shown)


bassosimone:
I suggest to commit this on top of #918 rather than making a separate pull request

AntonioLangiu:
Ok, I did it to split it logically! I will commit it on top of #918

Issue #741, opened 2016-07-31
http: make sure we have regress for parser errors


bassosimone:

bassosimone:
Trying to address this in #929 I've ended up refactoring significantly the parser. For this reason I am moving this issue from release-0.3.0 (stable) to release-0.4.0 (master).

bassosimone:
I think we should delay this, as it should go hand in hand with changes in `#1345

PR #929, opened 2016-11-17
Simplify http parser

bassosimone:
*Note*: superseded by #1041. What follows is the original rationale. At bottom there is a slightly modified TODO list that highlights what to do before closing this PR.

- - - 

The code for reading incoming HTTP responses was quite messy and not simple to follow. In particular, it mixed callbacks and exceptions, which makes it more complex to reason about its correctness.

This pull request modified the implementation to avoid both callbacks and exception. Now, when you have incoming data, you pass it to `feed()`. This function now returns an `Error` and specifically it can return one of the following errors:

1. *ParsingHeadersInProgressError*: meaning that the parser is still parsing response line and headers and more data is required to continue (this is similar to `EAGAIN`)

2. *PausedAfterParsingHeadersError*: meaning that the parser has stopped after encountering end of headers, such that you can inspect response metadata, and can be restarted using `parse()` (which returns the same set of errors as `feed()`)

3. *ParsingBodyInProgressError*: meaning that the parser is still parsing the body and more data is required to continue (this is similar to `EAGAIN`)

4. *NoError*: meaning that the parser has successfully parsed a complete message

5. Another, specific parser error (I have mapped all the errors relevant to us)

On EOF, you can call `eof()` that has similar return values.

Compared to the previous implementation, also:

1. the parser consumes all bytes that have been successfully parsed, allowing one to potentially know which was the first byte that failed the parser

2. the parse is paused after a full message has been received and it is currently not possible to resume it, meaning that we currently do not support pipelining and, more importantly, if the server maliciously sends us two back to back, unexpected responses, we return parse error rather than presenting only the second response to the user (this is basically an issue of the implementation in 0.3.x).

All of this started as an attempt to address #741, but I quickly realized I was not satisfied with the implementation and I wanted to do more refactoring to make it more easily testable. However, the amount of work done is too much to land in stable (even though I did not change the API). At the same time, even though much more refactoring was possible, I avoided it in the interest of making incremental changes.

- - - 

As stated below, this pull request has been superseded by #1041. Yet, before closing it, there are some activities that we must perform the extract the good parts:

- [ ] make sure we have no Valgrind related issues in #1041 
- [ ] write more regress tests for parser aiming at increasing coverage and spotting more corner cases
- [ ] add to #1041 note explaining that [stopping after headers would be complicated because of what seems to be a still open http-parser bug](https://github.com/nodejs/http-parser/issues/97)
- [ ] specifically make sure we stop parsing after the first response [NOT NEEDED ANYMORE]
- [ ] review HTTP requests we make and see whether we can make more of them
- [ ] see if we can make a long web-connectivity test using Valgrind


bassosimone:
Superseded by #1041 (which does more or less the same but much better). I'd like to keep to open because we can cherry pick from here some code to return more precise error codes for HTTP.

bassosimone:
To conclude this pull request. In addition to the refactoring of how we read, which was done better in #1041, this pull request contained two other interesting developments:

1. a complete enumeration of parser error code, now submitted as #1095 

2. code to prevent back to back requests to confuse the parser (added as a requirement to consider #1041 and ready for review again)

Given that everything interesting was either extracted or implemented better, we can now close.

PR #950, opened 2016-11-21
Make sure we always store and submit report-id

### LINKED ISSUE #905: ooni: report_id field is not set inside of measurement-kit ooni reports
Currently there is no `report_id` key (that is the value returned by the collector when a report is created) when writing reports to disk or when sending them to a collector.

bassosimone:
Closes #905. In detail:

1. Pass `Report &` to open() to avoid unnecessary copy.

   Since the `Report` owns the `Reporters` there is no need to
   pass the object and we can just pass a reference.

2. Add to `BaseReporter` a method to return report-id and
   teach the `Reporter` to retrieve it.

   The code is written assuming that just one base-reporter
   is able to provide us with a report-id. There is an
   assertion that makes sure we won't notice if this assumption
   is ever challenged by future changes.

3. Force setting the report-id (and warn) in collector client.

   This is meant as a fallback option. Code for that was already
   in place since ac145d1117505, but now the code has been
   moved into a lower level of abstraction such that we are
   sure that report-id is set in pretty much every code path
   which submits a report. Because this is a fallback, we
   emit a warning when we force-set the report-id.

4. Stop using attribute-like names for function arguments.

   As seen in 469d2fb4c3f3, it's not a smart pattern to give
   function arguments equal names to attributes.


src/libmeasurement_kit/report/report.cpp
```diff
@@ -44,7 +44,41 @@ void Report::open(Callback<Error> callback) {
...
+            }
+        }
+        /*
+         * XXX For now we trust the last reporter that provides us with a
+         * non-zero-length report-id. The reason for doing the full scan
+         * over the existing reporters and for asserting is to be ready for
+         * an hypothetical moment where more than one reporter will give
+         * back a good report-id (only ooni_reporter does that).
+         */
+        assert(count <= 1);
```
AntonioLangiu: I think it's excessive to use an assert here. I would prefer to return an error in this case, because this is for me not a case the library (and so the APP) should crash. What do you think about? Maybe we can add some unit test to be sure that in the future this case will never happens.

include/measurement_kit/report/report.hpp
```diff
@@ -41,7 +43,7 @@ class Report {
 
     void open(Callback<Error> callback);
 
-    void write_entry(Entry entry, Callback<Error> callback);
+    void write_entry(Entry entry, Callback<Error> callback, Var<Logger> logger);
```
AntonioLangiu: 👍  I like to pass explicitly the logger through the code!

src/libmeasurement_kit/report/report.cpp
bassosimone: I don't see such a change to easily slip through testing, but sure I'm adding an error for that.

bassosimone: Error added, see 3640a62. Thanks!

AntonioLangiu:
It looks good to me! Merging...

Issue #842, opened 2016-10-03
dns.hpp: get rid of query/type ids


bassosimone:
They occupy a lot of code and they also do not compile on Windows. I'd rather have an API where we pass in strings (as now, btw) and the strings are then validated by the code (as now, btw). By the way this should already be in #820.

bassosimone:
Closing because of #951

PR #951, opened 2016-11-21
dns: enums that should not conflict with windows

bassosimone:
In #842, I proposed of removing defines. After further discussions
others however I understood that it's better to keep them as they
give you certainty that a value is in a specific set.

So, since on Windows the build failed because there is a IN define
in some Windows system header, change the names used by defines.

While there, make code more robust and move implementation away
from the headers.


bassosimone:
Closes #842

Issue #693, opened 2016-06-28
Add bouncer API support in measurement-kit


hellais:
Currently measurement-kit uses a set of hardcoded values for collectors and test-helpers.

Ideally we should support communicating with the canonical ooni-probe bouncer and obtain the test-helpers and collector addresses from it.

bassosimone:
We can start implementing this by implementing an HTTP client for communicating with a specific bouncer for a specific test. For now, I'll just focus on the minimum amount of information we need to start off:

```C++
namespace mk {
namespace ooni {
namespace bouncer {

struct Reply {
    std::string https_collector;
    std::string https_helper;
};

void post_net_tests(
    std::string base_bouncer_url, // Example: "https://a.collector.ooni.io/bouncer/net-tests"
    std::string test_name,             // Example: "web_connectivity"
    std::string test_version,          // Example: "0.0.1"
    std::list<std::string> helpers,  // Example: ["web-connectivity"]
    Callback<Error, Reply> cb,
    Settings settings,
    Reactor reactor,
    Logger logger
);

}}}
```

Example JSON request to be sent for web_connectivity:

```JSON
{
    "net-tests": [
        {
            "input-hashes": [],
            "name": "web_connectivity",
            "test-helpers": [
                "web-connectivity"
            ],
            "version": "0.0.1"
        }
    ]
}
```

The method shall be `POST` and the request url path `/bouncer/net-tests`. Assuming `XO` contains the above JSON, with curl this would be:

```
curl -vd @XO https://a.collector.ooni.io/bouncer/net-tests
```

The JSON response, from which we should parse enough information to fill the structure above, looks like this:

```JSON
{
    "net-tests": [
        {
            "collector": "httpo://ihiderha53f36lsd.onion",
            "collector-alternate": [
                {
                    "address": "https://a.collector.ooni.io:4441",
                    "type": "https"
                },
                {
                    "address": "https://das0y2z2ribx3.cloudfront.net",
                    "front": "a0.awsstatic.com",
                    "type": "cloudfront"
                }
            ],
            "input-hashes": [],
            "name": "web_connectivity",
            "test-helpers": {
                "web-connectivity": "httpo://7jne2rpg5lsaqs6b.onion"
            },
            "test-helpers-alternate": {
                "web-connectivity": [
                    {
                        "address": "https://a.web-connectivity.th.ooni.io:4442",
                        "type": "https"
                    },
                    {
[...]

PR #967, opened 2016-11-24
Add OONI bouncer support

AntonioLangiu:
Add OONI bouncer support as exaplained in #693


src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+                            [=]() { cb(BouncerGenericError(), nullptr); });
+                        return;
+                    }
+                }
+
+                auto collector_alternate =
+                    response["net-tests"][0]["collector-alternate"];
+                bool collector_found = false;
+                for (auto collector : collector_alternate) {
+                    if (collector["type"] == "https") {
```
hellais: I would maybe put this as a settings. Something like "ooni/backend_type".

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+    request["net-tests"][0] = net_tests;
+
+    settings["http/url"] = base_bouncer_url;
+    settings["http/method"] = "POST";
+
+    http_request(
+        settings, {}, request.dump(),
+        [=](Error e, Var<http::Response> resp) {
+            if (e) {
+                reactor->call_soon([=]() { cb(e, nullptr); });
```
bassosimone: missing `return` here

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+namespace ooni {
+namespace bouncer {
+
+struct Reply {
+    std::string https_collector;
+    std::string https_helper;
+};
+
+template <MK_MOCK_NAMESPACE(http, request)>
+inline void
```
bassosimone: no need to inline; afaik templates are already inline bydef

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+                    logger->warn(
+                        "no https collector found in bouncer response");
+                    reactor->call_soon(
+                        [=]() { cb(MissingHttpsCollectorError(), nullptr); });
+                }
+                auto test_helpers_alternate =
+                    response["net-tests"][0]["test-helpers-alternate"]
+                            [test_name];
+                bool helper_found = false;
+                for (auto helper : test_helpers_alternate) {
```
hellais: Keep in mind that not all test helpers are https. Some test helpers (including the one used for http-invalid-request-line) is not https and you should just return that one.

In ooniprobe how we handle this is that we know which are the collectors for which there is the possibility of having alternate ones (currently only web-connectivity) and for the rest we just return the address.

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+                json response = json::parse(resp->body);
+                if (response.find("error") != response.end()) {
+                    if (response["error"] == "collector-not-found") {
+                        logger->warn(
+                            "collector not found for the requested test");
+                        reactor->call_soon([=]() {
+                            cb(BouncerCollectorNotFoundError(), nullptr);
+                        });
+                        return;
+                    } else if (response["error"] == "invalid-request") {
```
bassosimone: No need to `else` here since above you `return`

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+                        logger->warn("invalid request sent to the bouncer");
+                        reactor->call_soon([=]() {
+                            cb(BouncerInvalidRequestError(), nullptr);
+                        });
+                        return;
+                    } else {
```
bassosimone: No need to `else` here since above you `return`

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+                    }
+                }
+                if (!collector_found) {
+                    logger->warn(
+                        "no https collector found in bouncer response");
+                    reactor->call_soon(
+                        [=]() { cb(MissingHttpsCollectorError(), nullptr); });
+                }
+                auto test_helpers_alternate =
+                    response["net-tests"][0]["test-helpers-alternate"]
```
hellais: You probably want to also check to see if the array `response["net-tests"]` has a non zero length and fail gracefully in that case.

hellais: Edit: I see you handle that case inside of `catch (std::out_of_range &) {`, but actually here the meaning is that, no test helper was found.

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+                        reply->https_collector = collector["address"];
+                        collector_found = true;
+                        break;
+                    }
+                }
+                if (!collector_found) {
+                    logger->warn(
+                        "no https collector found in bouncer response");
+                    reactor->call_soon(
+                        [=]() { cb(MissingHttpsCollectorError(), nullptr); });
```
bassosimone: missing `return` here

test/ooni/bouncer.cpp
```diff
@@ -0,0 +1,85 @@
...
+    });
+}
+
+TEST_CASE("The bouncer can handle an invalid url") {
+    loop_with_initial_event([]() {
+        post_net_tests_impl(
+            "https://a.collector.ooni.io/bouncer/net-tests&antani",
+            "web-connectivity", "0.0.1", {"web-connectivity"},
+            [](Error e, Var<Reply>) {
+                // In this case the answer should be html and not json
```
hellais: I wouldn't depend on this current behaviour of the collector for client testing.

It could change in the future.

test/ooni/bouncer.cpp
```diff
@@ -0,0 +1,85 @@
...
+                     int nredirects = 0) {
+    Var<http::Response> response(new http::Response());
+    response->body = "{\"error\": \"invalid-request\"}";
+    cb(NoError(), response);
+}
+
+TEST_CASE("The bouncer can handle an invalid request") {
+    loop_with_initial_event([]() {
+        post_net_tests_impl<request_invalid>(
+            "https://a.collector.ooni.io/bouncer/net-tests", "web-connectivity",
```
hellais: I was initially confused by this unittest because I didn't notice the mocked out `request_invalid` and the request actually looked valid to me.

Maybe what you could be done is mark the request in some more obvious way (note the backend will reply with request invalid when it's not JSON).

AntonioLangiu: I'm not able to create an invalid json with the `post_net_tests_impl` function, so I need to mock the http_request function to get an invalid answer. I will add a comment to indicate this and make it more clear!

src/libmeasurement_kit/ooni/bouncer_impl.hpp
AntonioLangiu: Can you give me some examples of this cases please?

test/ooni/bouncer.cpp
hellais: Yes I would remove this test as it's testing something that may break not due to a bug in MK, but to a chance in the bouncer behaviour.

src/libmeasurement_kit/ooni/bouncer_impl.hpp
hellais: For example try requesting as a test_helper: `tcp-echo` (the test helper used in http-invalid-request-line).

What you will get back is not a .onion address and has no test-helpers-alternate.

Example:

```
curl --data '{"net-tests": [{"test-helpers": ["tcp-echo"], "name": "http_invalid_request_line", "input-hashes": [], "version": "1.0.0"}]}' https://bouncer.ooni.io/bouncer/net-tests
```

Response:
```
{"net-tests": [{"collector-alternate": [{"type": "https", "address": "https://a.collector.ooni.io:4441"}, {"front": "a0.awsstatic.com", "type": "cloudfront", "address": "https://das0y2z2ribx3.cloudfront.net"}], "version": "1.0.0", "name": "http_invalid_request_line", "test-helpers": {"tcp-echo": "213.138.109.232"}, "test-helpers-alternate": {}, "collector": "httpo://ihiderha53f36lsd.onion", "input-hashes": []}]}
```

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,116 @@
...
+    net_tests["version"] = test_version;
+    request["net-tests"][0] = net_tests;
+
+    settings["http/url"] = base_bouncer_url;
+    settings["http/method"] = "POST";
+    http_request(
+        settings, {}, request.dump(),
+        [=](Error e, Var<http::Response> resp) {
+            if (e) {
+                reactor->call_soon([=]() { cb(e, nullptr); });
```
bassosimone: Here I think we don't need `call_soon()` (or I don't understand why we're doing it)

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,116 @@
...
+                return;
+            }
+            Var<Reply> reply(new Reply);
+            try {
```
bassosimone: I don't like to have try... catch with such a wide scope. Could you refactor the code such that the `try...catch` only applies to the specific pieces of code that can throw? (I understand that it's common practice to code like that with OO code but my point is that *coupled with callbacks* increases the complexity of reviewing the code.)

AntonioLangiu: I'm working on this!

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -0,0 +1,167 @@
...
+            logger->warn("bouncer generic error");
+            return BouncerGenericError();
+        }
+        if (reply->response["net-tests"].empty()) {
+            logger->warn("generic bouncer error");
+            return BouncerTestHelperNotFoundError();
+        }
+    } catch (std::invalid_argument &) {
+        return JsonParseError();
+    } catch (std::out_of_range &) {
```
AntonioLangiu: This can't be thrown by the code inside try

src/libmeasurement_kit/ooni/bouncer_impl.cpp
```diff
@@ -0,0 +1,111 @@
...
+            logger->warn("generic bouncer error");
+            return BouncerTestHelperNotFoundError();
+        }
+    } catch (std::invalid_argument &) {
+        return JsonParseError();
+    } catch (std::out_of_range &) {
+        return JsonKeyError();
+    } catch (std::domain_error &) {
+        return JsonDomainError();
+    }
```
AntonioLangiu: @bassosimone I tried to split this code in different parts to make a smaller try_catch, but I don't like the idea to have some methods that can throw exceptions that are not catched correctly.

Because I was not able to understand exactly (watching the [documentation](https://nlohmann.github.io/json/)) what exceptions are thrown by the nlohmann::json methods, I would like to leave the code as it is, until we find a good way to use nlohmann::json without paying too much attention to exceptions.

bassosimone: Agreed!

bassosimone: @AntonioLangiu, in any case, I am fine with exceptions used like this. What makes me uncomfortable is mixing exceptions usage and calling callbacks (especially because also the callbacks may except).

bassosimone:
Thank you, merging!

PR #969, opened 2016-11-29
MK control port

bassosimone:
This pull request contains an initial implementation of OONI's control port. After a discussion with @hellais, we decided that it would be better to implement an HTTP and JSON based solution similar to the [solution implemented by Docker](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/) and having an API as close as possible with [the API of ooniprobe](https://github.com/TheTorProject/ooni-probe/blob/master/ooni/ui/web/server.py). Those changes will be performed on top of this pull request and the final result will then be squashed.


src/measurement_kit/cmdline/agent.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+/**
+ * Remove the next token from string.
+ * @param s String to read token from, updated in output.
+ * @return Next token or error.
+ */
+static ErrorOr<std::string> remove_next_token(std::string &s) {
+    size_t pos = s.find_first_of(" ");
+    if (pos == std::string::npos) {
+        return GenericError();
```
bassosimone: TODO: here we should probably return a better error

src/measurement_kit/cmdline/agent.cpp
```diff
@@ -0,0 +1,305 @@
...
+ * @returns NoError() on success, otherwise an error.
+ */
+static Error get_input_and_settings(const std::string &line, std::string &input,
+                                    Settings &options) {
+    nlohmann::json root;
+
+    try {
+        root = nlohmann::json::parse(line);
+    } catch (const std::exception &) {
+        return GenericError();
```
bassosimone: TODO: return a better error here

src/measurement_kit/cmdline/agent.cpp
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    if (root.find("input") != root.end()) {
+        try {
+            input = root["input"].get<std::string>();
+        } catch (const std::exception &) {
+            return GenericError();
```
bassosimone: TODO: return a better error here

src/measurement_kit/cmdline/agent.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    auto root_settings = root["settings"];
+    for (auto it = root_settings.begin(); it != root_settings.end(); ++it) {
+        std::string key = "";
+        std::string value = "";
+        try {
+            key = it.key();
+            value = it.value();
+        } catch (const std::exception &) {
+            return GenericError();
```
bassosimone: TODO: return a better error here

bassosimone:
We will prioritize python bindings, as discussed yesterday with @hellais

PR #972, opened 2016-12-01
Force the library to use the system resolver

AntonioLangiu:
This PR changes the default option of the library to use the `libevent resolver` to the `system_resolver`.

I will do a future "hot fix" PR to move the `test/dns/query.cpp` file into `test/dns/libevent_resolver.cpp` and add a generic `test/dns/query.cpp` that tests the query function!


AntonioLangiu:
[Wait for the review] I need to check why valgrind fails!

build/valgrind/openssl.supp
```diff
@@ -11,6 +11,6 @@
    fun:_ZN2mk8libevent6Poller4loopEv
    fun:_ZN2mk7Reactor23loop_with_initial_eventESt8functionIFvvEE
    fun:_ZN2mk23loop_with_initial_eventESt8functionIFvvEENS_3VarINS_7ReactorEEE
-   fun:_ZL31____C_A_T_C_H____T_E_S_T____418v
+   fun:_ZL31____C_A_T_C_H____T_E_S_T____*
```
AntonioLangiu: I decided to use this wildcard to match all tests having this valgrind problem.
The first implementation used another specific suppression inside this file, but I find this solution more easy and clear.

test/mlabns/mlabns.cpp
```diff
@@ -34,6 +34,7 @@ TEST_CASE("Query can pass the settings to the dns level") {
     settings["mlabns/metro"] = "trn";
     settings["mlabns/policy"] = "random";
     settings["dns/nameserver"] = "8.8.8.1";
+    settings["dns/engine"] = "libevent";
```
AntonioLangiu: This is a specific test that tries to pass a **dns/nameserver** to the lowest level from the higher (mlab-ns -> http -> net -> dns) to see it the settings are passed correctly. Since we are trying to change the **dns/nameserver** options, that is not supported with the system resolver, I decided to use libevent for this specific case!

AntonioLangiu:
@bassosimone it is ready to review

bassosimone:
@AntonioLangiu ack! I will review it either at end of this week or at beginning of next week!

build/valgrind/openssl.supp
bassosimone: Makes a lot of sense! 👍

test/mlabns/mlabns.cpp
bassosimone: Makes sense!

bassosimone:
@AntonioLangiu, it would make sense to see what would be the behavior of the library if you configure an invalid DNS. I expect that nothing would block for too much time, with `getaddrinfo()` eventually returning an error indicating timeout, but it would be great to test such an assumption.

AntonioLangiu:
Good idea, thanks! I'll check on my machine and report what happens!

AntonioLangiu:
We should also test this solution with an Android Device and iOS device.

bassosimone:
> We should also test this solution with an Android Device and iOS device.

Yep. I suggest you to do a private build of this branch and test this with some apps we develop. I am doing also something similar today testing `NdtMulti` (which uses the system resolver by default) on measurement-kit/ios-sample and measurement-kit/android-sample.

bassosimone:
I am going to merge this into master as soon as travis-ci completes.

PR #981, opened 2016-12-06
Add Apache Portable Runtime as dependency

bassosimone:
I did some *yolo* choices when developing this pull request to have APR compiling as quick as possible, but I feel like the following actions need to be done before being ready for review:

- [ ] evaluate the impact of the `./configure` tests that were disabled
- [ ] evaluate the impact of generating the characters table using the host's compiler
- [ ] make sure we pass `-Wall` and other warnings to compiler

### DIFF
--- a/build/android/make-toolchain
+++ b/build/android/make-toolchain
@@ -37,7 +37,8 @@ bash $MAKE_TOOLCHAIN \
   --toolchain=${ARCH}-4.9 \
   --install-dir=${INSTALL_DIR} \
   --use-llvm \
-  --stl=libc++
+  --stl=libc++ \
+  --force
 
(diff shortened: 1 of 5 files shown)


bassosimone:
With @hellais we recently concluded that we do not actually need this. We're going to have mainly Electron based apps, where path composing can be performed. Thinking more deeply, I am glad that we took this decision, since it allows to keep MK conceptually thinner.

As a byproduct of this discussion, I guess we should also deprecate and remove all the code for dealing with updating resources. We'll soon use orchestration to update the test lists and I think we probably want to use other mechanisms to update GeoIP files (I am not even sure we _need_ them, since we use Ubuntu to find out our IP address and that call also tells us the country code).

PR #992, opened 2016-12-08
Improve multi_ndt entry format

bassosimone:
- provide shortcuts for interesting bits of info in advanced

- better key name ("key" -> "fastest_test")

Bump test version.


bassosimone:
Oh, well, this can be merged as an hotfix

PR #994, opened 2016-12-09
Fix: libressl ios build

bassosimone:
So, when I submitted a previous version of the App in which 9b586f7 was used to define the prototype of `getentropy()`, I received an email from the Store stating the following:

> The app references non-public symbols in measurement-kit-test: _getentropy

In turn, 9b586f7 was added because the aforementioned prototype was missing and that, coupled with `-Werror` added by libressl for arm64 (strange eh?) lead the build to fail.

I admit my initial solution was quite ghetto. I basically attempted to please the compiler.

Anyway, after the error from the store, and digging further, I discovered that the issue was already reported libressl, tor and libevent (see  44587ee for pointers to such issues).

My understanding, [judging from tor's patch](https://gitweb.torproject.org/tor.git/commit/?id=https://gitweb.torproject.org/tor.git/commit/?id=16fcbd21c963a9a65bf55024680c8323c8b7175d), is that &mdash; depending on the operating system version one requests for &mdash; the function is hidden. So, to fix the issue, I have disabled entirely the check for `getentropy()`. This allowed me to publish on the store (we ares still talking about a TestFlight here, so only invited people can install the app).

Now, what would interesting to see before moving on is:

- [x] whether the issue is fixed in libressl-portable/portable@b5ebbf6 (i.e. `master`)... *no*.
- [x] whether by specifying a new version of iOS `getentropy()` becomes available... *apparently no?!*
- [x] whether we can prepare a more comprehensive fix for libressl (provided that there's need to do so, given that I am compiling not from master but from the `OPENBSD_6_0` branch, i.e. 2.4.4)... *done in libressl-portable/portable#248*
- [x] why `-Werror` is only specified for `arm64`... *antani*

# Availability of getentropy()

For the iPhoneOS platform we have:

```
$ pwd
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.1.sdk

$ grep -Rn getentropy *|awk -F: '{print $1}'|sort -u
usr/include/sys/syscall.h
usr/lib/libSystem.B.tbd
usr/lib/libSystem.tbd
usr/lib/libc.tbd
usr/lib/libdbm.tbd
usr/lib/libdl.tbd
usr/lib/libinfo.tbd
usr/lib/libm.tbd
usr/lib/libpoll.tbd
usr/lib/libproc.tbd
usr/lib/libpthread.tbd
usr/lib/librpcsvc.tbd

$ grep getentropy usr/include/sys/syscall.h 
#define	SYS_getentropy     500

$ grep -n getentropy usr/lib/libc.tbd 
622:                       '$ld$weak$os10.11$_getentropy', '$ld$weak$os10.11$_mkostemp', 
1631:                       _getegid, _getentropy, _getenv, _geteuid, _getfh, _getfsbyname,
```

For the MacOSX platform we have:

```
$ pwd
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk

$ grep -Rn getentropy usr|awk -F: '{print $1}'|sort -u
usr/include/python2.7/pyconfig.h
usr/include/sys/random.h
usr/include/sys/syscall.h
usr/lib/system/libsystem_c.tbd
usr/lib/system/libsystem_kernel.tbd
usr/share/man/man2/getentropy.2
usr/share/man/man4/random.4

$ grep -n getentropy usr/include/sys/random.h
37:int getentropy(void* buffer, size_t size);

$ grep -n getentropy usr/include/sys/syscall.h 
540:#define	SYS_getentropy     500

$ grep -n getentropy usr/lib/system/libsystem_c.tbd 
54:                       '$ld$weak$os10.11$_dirname_r', '$ld$weak$os10.11$_getentropy', 

$ grep -n getentropy usr/lib/system/libsystem_kernel.tbd 
177:                       _getdtablesize, _getegid, _getentropy, _geteuid, _getfh, _getfsstat,
```

So, in the macOS case there is a prototype whereas for iOS there is only a syscall. Also, the content of the `.tbd` files is not super clear but I know very little about them.


darkk:
As far as I understand

* darwin is the first(?) platform that have recently introduced it and had had no `getentropy()` API call before
* weak linking trickery SHOULD be done to use `getentropy` libc symbol while maintaining compatibility of the compiled binarty with older systems
* ignoring `getentropy` existence and using library-supplied replacement for a while MAY be done as well, but it's not perfect solution (assuming good *quality* of platform's `getentropy`)

So removing check for `getentropy` does not make things *worse*, but it does not benifit from shiny new API. There is a thing I would double-check -- does the built library binary export strong `getentropy` symbol?

I'm unsure if the case is really Apple's **bug**, it may be some misunderstanding as well. Anyway I have no Apple SW/HW to verify my assumptions. :)

Take it with a lot of skepticism as I've not touched iOS developer tools for six years.

darkk:
Another source of information: https://gitlab.com/gnutls/gnutls/issues/142

bassosimone:
@darkk thank you for your help. I have submitted a more comprehensive diff to libressl-portable and have also committed it in this pull request, such that we can test it. More structured reply after lunch.

bassosimone:
Simone wrote:
> More structured reply after lunch.

It seems this has been a long lunch, eh :-).

Leonid wrote:

>> As far as I understand
>>
>>  - darwin is the first(?) platform that have recently introduced it and had had no getentropy() API call before

I have checked. Getentropy [was added in OpenBSD 5.6](http://man.openbsd.org/OpenBSD-current/man2/getentropy.2), released on [October 1, 2014](https://www.openbsd.org/56.html), and according to `man getentropy` on my System: "The getentropy() function appeared in OSX 10.12" (i.e. Sierra, which was released on [September 20, 2016](https://en.wikipedia.org/wiki/MacOS_Sierra)).

>>  - weak linking trickery SHOULD be done to use getentropy libc symbol while maintaining compatibility of the compiled binarty with older systems

Good point. So, [using nm -gU](http://stackoverflow.com/a/4506140) on the library cross compiled for armv7 that's what I see:

```
$ nm -gU ./third_party/build/libressl/crypto/.libs/libcrypto.a |grep getentropy
./third_party/build/libressl/crypto/.libs/libcrypto.a(getentropy_osx.o):
00000000 T _getentropy
```

This means that getentropy's replacement is compiled as public symbol and included into the library. And getentropy is not defined as weak AFAICT. However, in the specific case of iOS I don't think weak linking giving preference to libc's symbol is what we want: if preference is given to the libc symbol, we are actually linking to a semi-hidden symbol that is not meant to be used (basically there is the symbol but it is not declared and the App store complains if your app needs that symbol). Speaking of OSX, instead, I believe that weak linking would instead be the optimal solution (compared to compiling getentropy replacement unconditionally if the user requests to compile a MacOS version lower than 10.11).

>>  - ignoring getentropy existence and using library-supplied replacement for a while MAY be done as well, but it's not perfect solution (assuming good quality of platform's getentropy)

Indeed.

>> So removing check for getentropy does not make things worse, but it does not benifit from shiny new API.

Agree. The patch I submitted to libressl does what tor does as far as macOS is concerned; if a version before 10.12 is required, the replacement is unconditionally built. If iOS is detected, the replacement is unconditionally compiled given that getentropy is not exported by iOS's libc.

>> There is a thing I would double-check -- does the built library binary export strong getentropy symbol?

Yes (see the above example).

>> I'm unsure if the case is really Apple's bug, it may be some misunderstanding as well.

I think here there are two issues: (according to tor's trac) Apple should flag symbols as "after specific version only" and do not allow `./configure` to find the `getentropy` symbol if earlier than 10.12 is requested; in iOS there is a `getentropy` symbol you can link with but no headers meaning that the symbol is not meant for user space consumption yet.

>> Anyway I have no Apple SW/HW to verify my assumptions. :)
>>
>> Take it with a lot of skepticism as I've not touched iOS developer tools for six years.

Thank you. Preparing this response I noticed my patch for libressl does not work correctly on macOS when earlier than 11.12 is requested because another bit of configure kicks in. As far as measurement-kit is concerned, I'd see we're good with merging my proposed patch (which "works" as in the app store accepts it and is similar in spirit to what tor does) and eventually the issue will be fixed upstream.

bassosimone:
Leoniv wrote:
> Another source of information: https://gitlab.com/gnutls/gnutls/issues/142

Ah, forgot to comment about this: it was a very interesting reading that helped me to understand what was actually going on. Basically, some symbols are marked as weak when they are not available on 10.11, so yes it's not totally clear whether it's a bug or a feature. It for sure requires people to do stunts.

darkk:
> Getentropy was added in OpenBSD 5.6

Yep, I meant *the first besides OpenBSD (as libressl is OpenBSD-native)*, so Mac is the very first platform with different `getentopy()` availability depending on version, so it's the very first platform that has to be dealt with by `libressl-portable` in terms of *binary* compatibility across platform versions.

I hope, this bag of words makes sense:-)

PR #995, opened 2016-12-09
Easy way to use json

AntonioLangiu:
This PR is a test to see if an implemention like this could be ok.

There are two functions that are able to block exceptions and return them as error_or.


AntonioLangiu:
Sorry, I just wanted to open this on my fork of measurement-kit and not on this repository. Closing!

PR #996, opened 2016-12-11
Implement mk::slurp() to read files at once

bassosimone:
Name inspired by Perl's slurp(). Implementation written in good old C
after checking what stackoverflow community have to suggest.

Unit test includes files with UTF-8 encoding and a gzipped file.


src/libmeasurement_kit/common/utils_impl.hpp
```diff
@@ -0,0 +1,63 @@
...
+ErrorOr<std::vector<T>> slurpv_impl(std::string p) {
+    FILE *filep = std_fopen(p.c_str(), "rb");
+    if (filep == nullptr) {
+        return FileIoError();
+    }
+    if (std_fseek_1(filep, 0, SEEK_END) != 0) {
+        std_fclose(filep);
+        return FileIoError();
+    }
+    // Note: ftello() might be better for reading very large files but
```
AntonioLangiu: [maybe bikeshedding] Considering the fact that we are working with small files, should we check somehow that the files are really small. Is my understanding right that if you try to read a file of 50mb with this function you will get a vector of 50mb in memory? If yes, should we limit this size of the file (checking the returned fseek position), considering that we are working on mobile devices??

test/common/utils.cpp
```diff
@@ -148,3 +148,158 @@ TEST_CASE("sha256_of() works as expected") {
...
+        REQUIRE(*maybe_res == expect);
+        REQUIRE(maybe_res->size() == 240);
+    }
+}
+
+TEST_CASE("slurp() works as expected") {
+    SECTION("In case of nonexistent file") {
+        auto maybe_res = mk::slurp("/nonexistent");
+        REQUIRE(!maybe_res);
+        REQUIRE(maybe_res.as_error().code == mk::FileIoError().code);
```
AntonioLangiu: why are you explicitly getting the code to compare them?

AntonioLangiu:
It looks good to me!

test/common/utils.cpp
bassosimone: Back in the day `Error` automatically converted to `int`. Then came `ErrorOr<T>` with automatic constructors for both `Error` and `T`. Then I started to wonder what would happen if `ErrorOr<int>` was passed a `Error`. Could the compiler, by mistake, convert the error to `int`?

This scared me and I reacted emotionally by removing automatic conversion to `int`. Probably it was not the optimal, rational reaction and it would have been better to add unit tests for that. I think we are still in time to amend my previous behavior and restore `Error` to its ancient glory.

Until we do that, we have the minor issue that `Catch` passes things you put into `REQUIRE` clauses to `std::cout` (or a similar stream). Unfortunately `Error` is only convertible to `bool`, therefore when `REQUIRE(err == GenericError())` is `false` you see something like:

> Assertion failed `1 == 1`

Which is counter-intuitive to say the least. That why I started adding `.code` explicitly. The same problem applies for the places in which I print `Error` using `%d`.

Now, considering that one can always convert to integer through a boolean (as is clearly exemplified in the `%d` case), I am not even sure my original fix was wise because conversion to `int` would probably still be possible through boolean. Furthermore, thinking more about the matter, I do not think a compiler would be allowed to implicitly convert `Error` to `int` when a constructor exists for both, given that `Error` is the type and `int` is available only after applying the cast operator.

Anyway, I guess we should open a issue about this and move on. Interesting how I did not see the matter until you commented and then, one second after reading the comment, the whole thing &mdash; including the wishful thinking, poor man fix &mdash; was apparent.

bassosimone:
@AntonioLangiu thank you for looking into it. I will now proceed with merging this!

src/libmeasurement_kit/common/utils_impl.hpp
bassosimone: I am not sure. In theory, I'd rather have the system tell me I'm using too much memory by having `malloc()` return `NULL`. Fixed constraints in general give a false sense of security and possibly limit resources unreasonably. Consider that by adding fixing constraints we're basically trying to do the job of the system with much less information. In general, that's a no-no.

But, on the other hand, we currently use `new` in many places and `new` can `throw` and we don't care about `std::bad_alloc()`. Furthermore, AFAIK under Linux `malloc()` never returns `NULL` because `Linux` has an optimistic memory allocation policy. That's basically my argument above except that it's applied to the `malloc()` libc function, which is less well versed that the kernel memory manager to judge and guess whether we're short of memory pages. The disadvantage of Linux policy however is that you basically `segfault` when you ask for memory and the system does not have a memory page available for you. That is, you cannot handle it gracefully. To be honest, I really don't like this Linux behavior and I prefer the behavior of BSD systems. (Note: what I am saying here is what I recall from some years ago, perhaps this changed.)

Bottom line is: we don't know in advance whether the system is going to tell us that we're running out of memory and in many cases we don't handle `bad_alloc()` for this reason. I still do not see a reason to add a fixed limit to the function, tho. That would still sound like arbitrary. I'd rather add a parameter allowing the caller to specify the maximum acceptable size of the file according to his judgement. This way it would not sound arbitrary to the user of the function.

Does it make sense?

test/common/utils.cpp
bassosimone: What do you think?

PR #1012, opened 2016-12-16
Implement update-resources task

bassosimone:


bassosimone:
We decided to use other means to update resources. So, let's attic this.

PR #1015, opened 2016-12-16
Use nlohmann::json without exceptions

AntonioLangiu:
This PR adds a function to use nlohmann::JSON without the need to catch the exceptions directly. This is very useful in asynchronous functions to avoid too many `try { } catch` in the code and make it easier to check all the possible exit point of a function.


bassosimone:
Given how json specific is all of this, how about ext.hpp -> json.hpp and ext/json.hpp -> json/json.hpp and common/noexceptions.hpp -> json/noexceptions.hpp?

AntonioLangiu:
So you are proposing to add a **json** folder and move all the files regarding json there!?!

bassosimone:
> So you are proposing to add a json folder and move all the files regarding json there!?!

Yes. What do you think?

doc/api/json.md
```diff
@@ -1,13 +1,13 @@
...
+json &mdash; module containing code used to handle json
 
 # LIBRARY
 MeasurementKit (libmeasurement_kit, -lmeasurement_kit).
 
 # SYNOPSIS
 ```C++
 // DO NOT INCLUDE THIS HEADER FILE DIRECTLY
-/*#include <measurement_kit/ext.hpp>*/
+/*#include <measurement_kit/json.hpp>*/
```
bassosimone: I guess it now makes sense to include the header

include/measurement_kit/json/noexceptions.hpp
```diff
@@ -0,0 +1,29 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+#ifndef MEASUREMENT_KIT_COMMON_NOEXCEPTIONS_HPP
+#define MEASUREMENT_KIT_COMMON_NOEXCEPTIONS_HPP
+
+#include <measurement_kit/common/error_or.hpp>
+
+namespace mk {
```
bassosimone: This probably should live in the `mk::json` namespace. Also, it would be nice to make `nlohmann::json` available as `mk::nlohmann::json` through a wrapper header (i.e. `<measurement_kit/json/json.hpp>` that includes the real file stored as `<measurement_kit/json/nlohmann/json.hpp>`. Also, if you have not already done that, I would make sure that the code for updating the `nlohmann/json.hpp` header is update to download the file in the correct location.

include/measurement_kit/json/noexceptions.hpp
```diff
@@ -0,0 +1,29 @@
...
+    } catch (std::invalid_argument &) {
+        return JsonParseError();
+    } catch (std::out_of_range &) {
+        return JsonKeyError();
+    } catch (std::domain_error &) {
+        return JsonDomainError();
+    } catch (Error &err) {
+        return err;
+    } catch (std::exception &ex) {
+        return UncaughtExceptionError(); // define a specific one
```
bassosimone: I guess we can now get rid of the comment, right?

AntonioLangiu:
To finish this PR I need to:

- [ ] Fix the documentation
- [ ] Move json_noexcept into the mk::json namespace, renaming it `noexcept`, so we will be able to use it like: `json::noexcept([]() {});`
- [ ] Add some tests

bassosimone:
A simpler implementation landed in the `v0.7.x` branch as #1253

PR #1014, opened 2016-12-16
Refine and finish bouncer implementation

bassosimone:

PR #1016, opened 2016-12-16
Start using bouncer

bassosimone:
Based on #1014.

TODO:

- [x] make sure we use the production bouncer unless we're in the regress tests (from #1061)


AntonioLangiu:
I merged master (fixing some conflicts) into this branch and now the diff is more clear. The list of commits contains also the commits from #1014, I could fix it with a rebase but I didn't want to force push on a branch I didn't wrote.

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -238,74 +238,131 @@ std::string Runnable::generate_output_filepath() {
     return filename.str();
 }
 
+void Runnable::contact_bouncer(Callback<Error> cb) {
+    if (!needs_bouncer) {
+        cb(NoError());
+        return;
+    }
+    auto bouncer = options.get(
+        "bouncer_base_url", std::string{"https://a.collector.ooni.io/bouncer"});
```
AntonioLangiu: I would define this string in [include/measurement-kit/ooni/collector_client.hpp](https://github.com/measurement-kit/measurement-kit/blob/master/include/measurement_kit/ooni/collector_client.hpp#L21) instead of hardcoding it here!

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -238,74 +238,131 @@ std::string Runnable::generate_output_filepath() {
...
+    logger->info("Contacting bouncer: %s", bouncer.c_str());
+    ooni::bouncer::post_net_tests(
```
AntonioLangiu: [OT] I know that this is not related to this PR but I don't like too much the name of this function.

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -238,74 +238,131 @@ std::string Runnable::generate_output_filepath() {
...
+                auto maybe_backend = reply->get_test_helper_alternate(
+                    th, "https"
+                );
+                if (!!maybe_backend) {
+                    logger->info("Bouncer discovered backend: %s",
+                         maybe_backend->c_str());
+                    if (options.find("backend") == options.end()) {
+                        logger->info("Using discovered backend");
+                        options["backend"] = *maybe_backend;
+                        // FIXME: This breaks multi-backend tests!!!
```
AntonioLangiu: [QUESTION] it would be easily solved if we would use a json instead of settings? What can we do in this case?

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -238,74 +238,131 @@ std::string Runnable::generate_output_filepath() {
...
+        bouncer, test_name, test_version, required_test_helpers,
+        [=](Error error, Var<BouncerReply> reply) {
+            if (error) {
+                cb(error);
+                return;
+            }
+            assert(!!reply);
+            auto maybe_collector = reply->get_collector_alternate("https");
+            if (!maybe_collector) {
+                cb(maybe_collector.as_error());
```
AntonioLangiu: Should we use a **default collector** in case we are not able to find it with the bouncer?

src/libmeasurement_kit/nettests/runnable.cpp
AntonioLangiu: Maybe a possible (not requested :) solution would be to make **std::list<std::string> required_test_helpers** a map instead of a list and then fill the values if we find them.

AntonioLangiu:
I reviewed and tested this implementation.

**I noticed that the collector resolved by the bouncer is not a testing collector. We should find a solution to avoid pushing data to the production collector.**

bassosimone:
> The list of commits contains also the commits from #1014, I could fix it with a rebase but I didn't want to force push on a branch I didn't wrote.

I wouldn't bother: we are most probably going to squash at the end. (Also, I don't like force pushing or rebasing because both can hide mistakes.)

src/libmeasurement_kit/nettests/runnable.cpp
bassosimone: True. Like it is done for the collector, we can use the testing server. It is `bouncer.test.ooni.io` on port 443 using https.

src/libmeasurement_kit/nettests/runnable.cpp
bassosimone: How would you call it?

src/libmeasurement_kit/nettests/runnable.cpp
bassosimone: Not sure. Maybe @hellais has an opinion on this. How does ooniprobe work?

src/libmeasurement_kit/nettests/runnable.cpp
bassosimone: I'd keep the list a list. But perhaps we can add a map and fill the map here. Does it make sense?

src/libmeasurement_kit/nettests/runnable.cpp
hellais: If we don't find the collector ooniprobe will fail.

bassosimone: Okay, good to know, thanks!

bassosimone:
@hellais This is now ready for review!

bassosimone:
There is also [this failing test to fix](https://travis-ci.org/measurement-kit/measurement-kit/jobs/233447611#L2119):

```
test/nettests/runnable.cpp:229
test/nettests/runnable.cpp:230: FAILED:
with expansion:
  1 == 9
```

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -284,62 +284,119 @@ std::string Runnable::generate_output_filepath() {
...
+                return;
+            }
+            logger->info("Bouncer discovered collector: %s",
+                         maybe_collector->c_str());
+            if (options.find("collector_base_url") == options.end()) {
```
AntonioLangiu: According with the assertion **"If we don't find the collector ooniprobe will fail"** this code should be refactored avoiding this check and directly setting the `options["collector_base_url"]` with `*maybe_collector`

AntonioLangiu:
I reviewed this PR and added some little fixes. I think it's ready to merge!

bassosimone:
This is now good to go. Only problem is that tests are failing [because we have an issue with the bouncer](https://paste.debian.net/936334/), as discussed on `#ooni-internal` with @anadahz (thanks for help!)

bassosimone:
Let's wait a little more to see whether the redeployed bouncer helps.

bassosimone:
Alright. This is now good-to-squash. I'll wait for travis to complete to see whether I managed to increase a bit the code coverage playing tricks with various backend ooni infrastructure.

PR #1017, opened 2016-12-18
Refactor and improve build scripts

bassosimone:


build/dependency
```diff
@@ -25,23 +25,54 @@ Examples:
...
+
 pkg_configure() {
     if [ -x ./configure ]; then
         echo "- ./configure --prefix=$pkg_prefix $pkg_configure_flags"
-        ./configure --prefix=$pkg_prefix $pkg_configure_flags
+        ./configure --prefix=/ $pkg_configure_flags
     elif [ -f ./CMakeLists.txt ]; then
         echo "- cmake -D CMAKE_INSTALL_PREFIX=$pkg_prefix $pkg_cmake_flags ."
-        cmake -D CMAKE_INSTALL_PREFIX=$pkg_prefix $pkg_cmake_flags .
+        cmake -D CMAKE_INSTALL_PREFIX=/ $pkg_cmake_flags .
```
bassosimone: Note: using `/` rather than `pkg_prefix` because we take advantage of `DESTDIR` to be able to create tarballs that we can store online for speeding up the build process (especially for mobile).

build/dependency
```diff
@@ -78,49 +132,40 @@ if [ -z "$pkg_prefix" ]; then
 fi
 export pkg_prefix
 cd $pkg_rootdir
-if [ -x build/spec/$1 ]; then
-    exec build/spec/$1  # If the spec is executable, pass control to it
-fi
```
bassosimone: Now all the software we build follows the spec. No need for this shortcut anymore.

build/dependency
```diff
@@ -78,49 +132,40 @@ if [ -z "$pkg_prefix" ]; then
...
-    git clone $pkg_clone_flags $pkg_repository third_party/cache/$pkg_name
-    cd third_party/cache/$pkg_name
-    if [ ! -z "$pkg_branch" ]; then
-        git checkout $pkg_branch
-        if [ "`git rev-parse HEAD`" != "$pkg_commit" ]; then
+if [ "$pkg_repository" != "" ]; then
+    rm -rf ./build/third_party/$pkg_name
+    git clone $pkg_clone_flags --single-branch --depth 2                       \
+        --branch $pkg_branch_or_tag $pkg_repository                            \
+          ./build/third_party/$pkg_name
```
bassosimone: This is a quick clone procedure adapted from CocoaPods. We fetch two commits rather than one such that we can make sure the tip of the branch is what we expect when we follow a branch rather than a tag.

Note that the variable indicating the branch has been renamed to be more precise.

build/dependency
```diff
@@ -78,49 +132,40 @@ if [ -z "$pkg_prefix" ]; then
...
+    cd ./build/third_party/$pkg_name
+    if [ ! -z "$pkg_tip" ]; then
+        if [ "`git rev-parse HEAD`" != "$pkg_tip" ]; then
```
bassosimone: The variable pkg_commit has been renamed as pkg_tip to better explain its purpose.

build/ios/archive-deps
```diff
@@ -1,16 +0,0 @@
...
-# remove directories that we don't need for cross compiling MK
-find tmp -type f -name \*.la -exec rm {} \;
-find tmp -type f -name \*.pc -exec rm {} \;
-find tmp -type d -depth -name share -exec rm -rf {} \;
-find tmp -type d -depth -name bin -exec rm -rf {} \;
-find tmp -type d -depth -name pkgconfig -exec rm -rf {} \;
-
-now=`date +%Y%m%dT%H%M%SZ`
-tar -czf ios-dependencies-$now.tgz tmp
-gpg2 -u 7733D95B -b --armor ios-dependencies-$now.tgz
```
bassosimone: Not needed anymore. Now we package each dependency separately.

build/ios/build-frameworks
```diff
@@ -4,6 +4,22 @@ set -e
...
+            libs="$libs tmp/iphonesimulator/$arch/lib/$input_lib.a"
+        fi
+    done
+    for arch in arm64 armv7s armv7; do
+        if [ -f tmp/iphoneos/$arch/lib/$input_lib.a ]; then
+            libs="$libs tmp/iphoneos/$arch/lib/$input_lib.a"
+        fi
+    done
+    echo $libs
+}
```
bassosimone: Factored out common code FTW.

bassosimone: Especially: allow some architectures to be missing.

build/ios/cross
```diff
@@ -1,35 +1,31 @@
 #!/bin/sh
 set -e
 
-if [ $# -lt 2 ]; then
-    echo "$0 iphonesimulator i386|x86_64 [command [args...]]" 1>&2
-    echo "$0 iphoneos armv7|armv7s|arm64 [command [args...]]" 1>&2
+if [ $# -lt 1 ]; then
+    echo "$0 i386|x86_64|armv7|armv7s|arm64 [command [args...]]" 1>&2
```
bassosimone: Auto-guess the platform from the arch. Again simplicity FTW.

build/ios/dependency
```diff
@@ -10,15 +10,9 @@ ROOTDIR=$(cd $(dirname $0)/../.. && pwd -P)
...
 IOSDIR=$BUILDDIR/ios
 
-echo ""
-echo "Cross-compiling $1 for all platforms and archs"
-
-$IOSDIR/cross iphonesimulator i386 $BUILDDIR/dependency $1
-$IOSDIR/cross iphonesimulator x86_64 $BUILDDIR/dependency $1
-$IOSDIR/cross iphoneos armv7 $BUILDDIR/dependency $1
-$IOSDIR/cross iphoneos armv7s $BUILDDIR/dependency $1
-$IOSDIR/cross iphoneos arm64 $BUILDDIR/dependency $1
```
bassosimone: Factored code using a loop

build/ios/dependency
```diff
@@ -10,15 +10,9 @@ ROOTDIR=$(cd $(dirname $0)/../.. && pwd -P)
...
-
-# Install spec so we know what version was used to build the archive
-install -d $IOSDIR/tmp/spec
-install -m644 $BUILDDIR/spec/$1 $IOSDIR/tmp/spec/$1
```
bassosimone: Now each `./build/dependency` run creates a tarball

build/ios/library
```diff
@@ -2,18 +2,5 @@
...
-    set -e # just in case
-    cd $ROOTDIR
-    DEPS_URL=https://github.com/measurement-kit/measurement-kit-deps/releases/download/2016-12-10/ios-dependencies-20161210T224506Z.tgz
-    DEPS_FILE=$(basename $DEPS_URL)
-    curl --progress-bar -LO $DEPS_URL
-    curl --progress-bar -LO $DEPS_URL.asc
-    gpg2 --verify $DEPS_FILE.asc
-    tar -xzf $DEPS_FILE
-    rm $DEPS_FILE $DEPS_FILE.asc
-)
```
bassosimone: Now dependencies are downloaded as part of `./build/dependency`

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -54,11 +54,21 @@ diff --git a/configure.ac b/configure.ac
...
+@@ -289,8 +288,8 @@ AC_C_INLINE
+ AC_HEADER_TIME
  
  dnl Checks for library functions.
- AC_CHECK_FUNCS([gettimeofday vasprintf fcntl clock_gettime strtok_r strsep])
+-AC_CHECK_FUNCS([gettimeofday vasprintf fcntl clock_gettime strtok_r strsep])
 -AC_CHECK_FUNCS([getnameinfo strlcpy inet_ntop inet_pton signal sigaction strtoll inet_aton pipe eventfd sendfile mmap splice arc4random arc4random_buf issetugid geteuid getegid getprotobynumber setenv unsetenv putenv sysctl])
++AC_CHECK_FUNCS([gettimeofday vasprintf fcntl strtok_r strsep])
 +AC_CHECK_FUNCS([getnameinfo strlcpy inet_ntop inet_pton signal sigaction strtoll inet_aton pipe eventfd sendfile mmap splice arc4random arc4random_buf arc4random_addrandom issetugid geteuid getegid getprotobynumber setenv unsetenv putenv sysctl])
  AC_CHECK_FUNCS([umask])
```
bassosimone: In this patch we get rid of `clock_gettime()` which is mainly used by libevent to implement precise timing but which is also something that's is only available on iOS >= 10.0. A better patch could be implemented like I already did for getentropy() for libressl, but for the purpose of blessing v0.4.0-beta this one would do.

bassosimone: xref: #1027

build/patch/pre-autogen/libressl/common/002.patch
```diff
@@ -0,0 +1,24 @@
...
+-	else
+-		git clone $LIBRESSL_GIT/openbsd
+-	fi
+-fi
+-(cd openbsd
+- git fetch
+- git checkout $openbsd_branch
+- git pull --rebase)
++git clone --single-branch --depth 2 --branch $openbsd_branch                   \
++  https://github.com/libressl-portable/openbsd.git
```
bassosimone: Speed up cloning openbsd sources.

build/release
```diff
@@ -6,8 +6,8 @@ if [ $# -ne 2 ]; then
 fi
 old_version=$1
 new_version=$2
-for file in configure.ac doc/api/common/version.md doc/index.md \
-            include/measurement_kit/common/version.h \
+for file in configure.ac doc/api/common/version.md doc/index.md                \
+            include/measurement_kit/common/version.h build/spec/mk             \
```
bassosimone: Also update the version number in `build/spec/mk` (not used yet, but `#correctness`)

bassosimone:
Okay, I guess this is ready for review now. As part of the next release cycle it would perhaps make sense to update also the Android build system to match the iOS build system enhancements implemented here.

build/dependency
```diff
@@ -25,23 +25,54 @@ Examples:
...
 
+pkg_depend() {
+    for dep in `echo $pkg_deps`; do
+        if [ "$pkg_cross" != "" -a "$pkg_arch" != "" ]; then
+            dep="$pkg_cross-$pkg_arch-$dep"
+        fi
+        if [ ! -f $pkg_rootdir/$dep.tgz ]; then
+            baseurl=https://github.com/measurement-kit/dependencies
+            wget -O $pkg_rootdir/$dep.tgz                                      \
+                $baseurl/releases/download/2016-12-19/$dep.tgz
```
AntonioLangiu: Is there some way of getting always the latest release instead of having to change it?

bassosimone: Possibly. However we may want to have different dependencies in different branches.

AntonioLangiu:
I reviewed and tested (normal build and cross build for iOS) this PR and It looks good to me. As discussed with @bassosimone, I would like to see a future PR with changes to make the build script more linear and avoid to many jumps from one script to the other, expecially when cross-compiling.

I will finally test it for Android and, if it's ok, merge this PR and open a new issue to indicate some changes I would like to see in the build system.

Thanks @bassosimone!

bassosimone:
Added 205491f to repair the Android build as spotted by @AntonioLangiu.

build/dependency
bassosimone: Maybe we can create a release called "stable" and a release called "testing", so the URL would be something like `$baseurl/releases/download/stable/$dep.tgz"?

bassosimone:
Given the above fix and reflecting about the size of this diff, I am starting to wonder whether it would be more proper to only commit to stable the fix for libevent and port this set of patches to master, where we can merge them now that we are far from a stable release, so that they can be tested more. I feel like merging into stable would violate our gitflow-like workflow too much.

@AntonioLangiu, do you think that my reasoning makes sense?

bassosimone:
> After some testing on Android I noticed that it breaks the compilation and so I think we need to fix it before we can merge this PR. We should divide this PR, moving into a new one the parts we think are ok.

I'm extracting the smallest possible patch for fixing the issue at hand from here. Meanwhile, I have rebased this into master and will update tracking PRs shortly. Some more bits have slipped through in this pull request because master and stable are not in sync but this will disappear soon (I will make a new alpha release of MK within end of day and that will be merged back into master).

AntonioLangiu:
Thanks 👍

build/android/archive-deps
```diff
@@ -13,4 +13,4 @@ find jni -type d -depth -name pkgconfig -exec rm -rf {} \;
 
 now=`date +%Y%m%dT%H%M%SZ`
 tar -czf android-dependencies-$now.tgz jni
-gpg2 -u 7733D95B -b --armor android-dependencies-$now.tgz
+$ROOTDIR/build/sign android-dependencies-$now.tgz
```
bassosimone: Replace custom signing code with specific script

build/dependency
```diff
@@ -32,6 +32,13 @@ apply_patches() {
     fi
 }
 
+pkg_init() {
+    if [ -f Makefile ]; then
+        echo "- make distclean"
+        make distclean
+    fi
+}
```
bassosimone: Add the `init` step, used by the `build/spec/mk` spec to cleanup the working dir

build/dependency
```diff
@@ -86,14 +93,17 @@ pkg_make_check() {
...
+    fi
     vspec=$pkg_branch_or_tag
     if [ "$pkg_tip" != "" ]; then
         vspec="$pkg_tip"
     fi
     based=$pkg_rootdir/third_party/tmp/$pkg_name-$vspec
     adir=$based/$pkg_cross/$pkg_cross_arch
-    make $pkg_make_install_flags install DESTDIR=$adir
+    echo "- make $pkg_make_install_flags $pkg_make_install_rule"
+    make $pkg_make_install_flags $pkg_make_install_rule DESTDIR=$adir
```
bassosimone: Allow a custom `install` rule to exist

build/ios/.gitignore
```diff
@@ -1,4 +1,3 @@
 .DS_Store
 *.tgz
 *.tgz.asc
-tmp
```
bassosimone: Always keep the directory there and ignore anything inside it.

build/ios/archive-deps
```diff
@@ -13,4 +13,4 @@ find tmp -type d -depth -name pkgconfig -exec rm -rf {} \;
 
 now=`date +%Y%m%dT%H%M%SZ`
 tar -czf ios-dependencies-$now.tgz tmp
-gpg2 -u 7733D95B -b --armor ios-dependencies-$now.tgz
+$ROOTDIR/build/sign ios-dependencies-$now.tgz
```
bassosimone: Use specific script for signing

build/release
bassosimone: Now there is a version number inside of the spec

build/spec/embedded-mk
```diff
@@ -1,21 +0,0 @@
-#!/bin/sh
-set -e
-
-# TODO: see if we can merge this and ./build/spec/mk
```
bassosimone: Todo addressed and spec merged with `./build/spec/mk`

bassosimone:
I have significantly simplified and refactored this PR, removing more complex code and only leaving small enhancements with respect to master. I am thus going to merge this as an hotfix.

Issue #988, opened 2016-12-07
dns: add getaddrinfo() like interface


bassosimone:
To start off, we could move `resolve_hostname()` from `net/connect.cpp` to the DNS layer and allow the code of other modules to call it publicly. As an improvement, it could probably be smart to have a specific implementation of this for the `system` engine where we can call directly `getaddrinfo` to obtain this information in one single step. Also the current implementation could be improved to run the queries in parallel since now we have abstractions to do that. All of this could be subject of one or more PRs.

bassosimone:
I am going to close this issue and create two child tickets out of it: #1038 and #1039.

PR #1022, opened 2016-12-20
Move dns resolution function from net to dns

AntonioLangiu:
As described in the issue #988

### DIFF
--- a/src/libmeasurement_kit/net/connect.hpp
+++ b/src/libmeasurement_kit/net/connect.hpp
@@ -31,13 +31,6 @@ void connect_first_of(Var<ConnectResult> result, int port,
                       Var<Logger> logger = Logger::global(), size_t index = 0,
                       Var<std::vector<Error>> errors = nullptr);
 
-typedef std::function<void(ResolveHostnameResult)> ResolveHostnameCb;
-
-void resolve_hostname(std::string hostname, ResolveHostnameCb cb,
-                      Settings settings = {},
-                      Var<Reactor> reactor = Reactor::global(),
-                      Var<Logger> logger = Logger::global());
-
 void connect_logic(std::string hostname, int port,
                    Callback<Error, Var<ConnectResult>> cb,
                    Settings settings = {},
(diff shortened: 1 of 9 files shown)


AntonioLangiu:
@bassosimone I moved different parts of code from net/connect to dns/query. I'm going to modify them according to the issue #988, do you prefer if I do it here or if I open another sequential PR based on this?

src/libmeasurement_kit/net/connect.hpp
```diff
@@ -31,13 +31,6 @@ void connect_first_of(Var<ConnectResult> result, int port,
                       Var<Logger> logger = Logger::global(), size_t index = 0,
                       Var<std::vector<Error>> errors = nullptr);
 
-typedef std::function<void(ResolveHostnameResult)> ResolveHostnameCb;
```
AntonioLangiu: This is the only part of code I removed when moving code. I removed it in favour of `Callback<ResolveHostnameResult> cb`

bassosimone:
At a first glance it seems good to me and I'd like to include it in 0.5.0-alpha. While at it, can you update the documentation and eventually fix other documentation inconsistencies and make sure the moved piece of code has coverage 100% (possibly it would be great to achieve coverage 100% *without* integration tests).

bassosimone:
@AntonioLangiu, can you document the functions and structures made public by this PR?

AntonioLangiu:
> update the documentation and eventually fix other documentation inconsistencies

I updated the documentation but I will take care of it after the next PR, that is going to refactor this code using **effectively the system resolver** and fixing some issues.

> make sure the moved piece of code has coverage 100%

It already has [100% coverage](https://coveralls.io/builds/9360981/source?filename=src%2Flibmeasurement_kit%2Fdns%2Fquery.cpp#L30). The case where the hostname is an ip (v4 or v6) address is tested with unit test. The case where the hostname is a domain is tested with integration tests, but it's a very used function and it's tested also by all the other procedures using it (consider that the resolution part is executed 699 times).

PR #1029, opened 2016-12-21
nettests: randomize input

bassosimone:
Closes #1023.

PR #1030, opened 2016-12-21
Support for country-specific test lists

### LINKED ISSUE #1026: runnable: support country specific test list

bassosimone:
Based on #1029 

Closes #1026 

- [x] add regress tests: will be done in another, upcoming PR
- [x] see if we can quickly address #1033: not needed (discussed w/ @hellais)


doc/api/nettests/base_test.md
```diff
@@ -79,15 +80,25 @@ The `set_verbosity` and the `increase_verbosity` methods allow to,
...
-set, respectively, the input file path (i.e. the path of the file
-containing the inputs that the test should consume) and the output
-file path (i.e. the file where the test should write the results
-of the test). By default there is no input file path and the output
-file path is the empty string. In that case, the test will generate
-an output file path appropriate for the test itself. If you want
-to disable writing output to disk, you should pass the test the
-`no_file_report` setting instead, by setting its value to false.
+The `add_input_filepath` adds one file path to the input paths list, which
+by default is empty. Tests that require input would fail if no input file
```
hellais: Typo:

s/would fail/will fail/

doc/api/nettests/base_test.md
```diff
@@ -79,15 +80,25 @@ The `set_verbosity` and the `increase_verbosity` methods allow to,
...
+path is specified, while tests that do not require input will ignore input
+if it is provided by the caller. If an input file path cannot be openned
+or read, measurement-kit will emit a warning message, but the test will not
+fail. If an input file path contains the string "${cc}" this is replaced
+with the current country code in lower case (that is, if you are in Italy
+&mdash; country code `IT` &mdash; the string "${cc}" is replaced with
+`it`).
```
hellais: I would rephrase this sentence as:

"If an input file path contains the string "${cc}" this is replaced with the two letter country code of the network the user is running the test from in lower case"

hellais: `<begin bikeshed>`
Maybe we can call `${cc}` `${probe_cc}` for consistency with ooniprobe?
`</end bikeshed>`

include/measurement_kit/ooni/error.hpp
```diff
@@ -9,23 +9,22 @@
...
+MK_DEFINE_ERR(MK_ERR_OONI(6), MissingRequiredUrlError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(7), GeoipDatabaseOpenError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(8), GeoipCountryCodeLookupError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(9), GeoipCountryNameLookupError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(10), GeoipCityLookupError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(11), GeoipAsnLookupError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(12), CannotGetResourcesVersionError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(13), CannotGetResourcesManifestError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(14), CannotGetResourceError, "")
+MK_DEFINE_ERR(MK_ERR_OONI(15), ResourceIntegrityError, "")
```
hellais: Why did the error codes change numbers?

I think you did it because you removed the "CannotOpenInputFileError", but isn't it better to perhaps keep the error codes consistent and just leave a "hole" where that error once was?

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -262,28 +262,42 @@ void Runnable::begin(Callback<Error> cb) {
...
-                        return;
+                    for (auto input_filepath: input_filepaths) {
+                        input_filepath = std::regex_replace(
+                            input_filepath,
+                            std::regex{R"(\$\{cc\})"},
+                            probe_cc_lowercase
+                        );
+                        std::ifstream input_generator{input_filepath};
+                        if (!input_generator.good()) {
+                            logger->warn("cannot open input file");
```
hellais: I understand why you did this here (so that when the country specific list does not exist we don't throw an error), but maybe we should check at the end of this loop if there is at least once good file and if not raise an exception because it means we are actually not going to be able to do much with this test.

bassosimone: I did that in 2bcf811, thanks

hellais:
I tested it locally and everything is green. You can proceed with merging 👍

bassosimone:
Thank you, I will merge!

PR #1035, opened 2016-12-27
ci: only use travis-ci and docker

bassosimone:
This pull request:

1. removes all build systems except travis (which means that the build status check will always fail because circle-ci file is missing, so check travis status only to decide whether to merge)

2. removes all complexity from `.travis.yml` and instead uses docker on travis-ci to build

3. uses yakkety (i.e. 16.10) because there is [a known bug of the Valgrind shipped with 16.04](https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/1574437)

4. simplifies the docker build system

5. removes vagrant files and scripts because now docker also runs on macos

6. allows to conditionally enable certain builds depending on the branch (I thought it was possible to specify this in the build matrix, like it's possible on other CI systems, however I couldn't enable it directly)

After this change the build is 2x slower (from around 10 minutes to around 20 minutes). However, no more wrestling with the super old 12.04 is needed to compute coverage and run regress tests through valgrind. I consider this an improvement, given all the time spent in the past to understand how to make 12.04 work for us. Also, this has the advantage that the same test scripts could also be run locally. Thus, I think we should merge this.

Things to check before moving to "ready for review":

- [x] that coverage is correctly measured for the current branch: see 386dd85 and [coveralls:9434129](https://coveralls.io/builds/9434129)
- [x] that the code for building conditionally works: see e5fa827 and [travis:186942867](https://travis-ci.org/measurement-kit/measurement-kit/builds/186942867)
- [x] that the valgrind build fails if we introduce memory errors: see bac4fd7 and [travis:186937198](https://travis-ci.org/measurement-kit/measurement-kit/jobs/186937198)


bassosimone:
After merging with master, it is clear that this pull request only contains subtle differences, e.g., the fact that here we do not use native travis but docker on travis to run valgrind. So, we can close.

bassosimone:
Forgot to say: of course this is because I've broken this pull request in pieces and I have submitted them as simple pull requests and merged them throughout the day.

PR #1036, opened 2016-12-27
Increase code coverage a bit

bassosimone:


bassosimone:
In e009c1a I have merged #1035 so to measure coverage. Apparently coveralls has sopped working for travis-ci on 12.04 and I honestly don't want to waste my time to fix it. Let's use docker for that.

PR #1037, opened 2016-12-27
By default use the system-resolver

bassosimone:
It should have been done in #972 but something went wrong during
the pull request process and we only created the condtions for that
to work correctly (I suspect some rebase went wrong).


AntonioLangiu:
I confirm what @bassosimone wrote. This diff should have been in #972, but for some reason is not (maybe I did something wrong with the rebase). I tested it locally and it passes all the tests. I will wait for travis to finish and then I will merge it!

PR #1041, opened 2016-12-27
Make response parser code less ugly

bassosimone:
Supersedes #929 

Related to #741 

To complete:
- [ ] make sure that we stop the parser at the end of the first request to prevent it from parsing further requests (otherwise, given two back to back requests, the code would return only the second one)
- [ ] add some regress tests
- [ ] make sure that issues fixed by #1174 and #1175 are still fixed here


bassosimone:
Commit a689fc2 forward ported #1121 to this branch. I think that due to merging #1121 to stable and that going into master after some time is going to lead to some conflict in here.

bassosimone:
Superseded by #1345

PR #1059, opened 2017-01-11
Add regress tests for input processor

bassosimone:


test/nettests/utils.cpp
```diff
@@ -0,0 +1,197 @@
...
+                                       "creativecommons.org",
+                                       "cyber.law.harvard.edu",
+                                       "duckduckgo.com",
+                                       "netflix.com",
+                                       "nmap.org",
+                                       "www.emule.com"};
+
+        std::deque<std::string> copy = expect;
+        auto count = 0;
+        for (; count < 8; ++count) {
```
hellais: I was about to say this was a small number, but actually P(x) = {same permutation x 8} = 1/10!^8 ~ 3,32579699816295e-53.
I guess that's quite unlikely.

hellais:
This diff looks good to me. Moar unittests are always good!

PR #1031, opened 2016-12-21
Feature/recognize utf8

bassosimone:
Closes #970.

Based on #1017.

PR #1062, opened 2017-01-13
Feature: recognize utf8 [rebased with more linear history]

bassosimone:
```
dunkerque$ git diff feature/recognize_utf8
dunkerque$
```

Supersedes #1031.


hellais:
I have begun reviewing this:

* [ ] Read the diff
* [ ] Build and run tests
* [ ] Do some testing with httpbin.org

hellais:
In testing this I noticed that measurement-kit does not strip IP addresses from the HTTP response body.

I don't think this should be part of this, but I filed a ticket about this in here: https://github.com/measurement-kit/measurement-kit/issues/1063

bassosimone:
Cc: @lorenzoPrimi

hellais:
I have done quite a bit of testing on this branch and this good to be merged for me.

PR #1076, opened 2017-01-15
Fix/case sensitive drive

hellais:
While investigating an issue I had with building ooniprobe-ios, I seem to have uncovered a bug related to case (in)sensitive drives.

For the record when running `pod update` (or `pod install`) I was getting the following error:

```
configure: WARNING: using cross tools not prefixed with host triplet
configure: WARNING: skipped the ca-bundle detection when cross-compiling
configure: WARNING: using cross tools not prefixed with host triplet
configure: WARNING: skipped the ca-bundle detection when cross-compiling
configure: WARNING: using cross tools not prefixed with host triplet
configure: WARNING: skipped the ca-bundle detection when cross-compiling
configure: WARNING: using cross tools not prefixed with host triplet
configure: WARNING: skipped the ca-bundle detection when cross-compiling
configure: WARNING: using cross tools not prefixed with host triplet
configure: WARNING: skipped the ca-bundle detection when cross-compiling
cp: tmp/iPhoneSimulator/i386/include/measurement_kit: No such file or directory
```

However from looking at a measurement-kit build the true path is actually `tmp/iphonesimulator/i386/` (all lowercase). I didn't encounter this issue in the past because my main drive was case insensitive and I suspect so are the drives of @bassosimone and @lorenzoPrimi .

I pushed this branch in the hope that I can build a good ooniprobe-ios. If this actually fixes it I will update this PR.

(note: this branch is actually based against stable)

### DIFF
--- a/build/ios/build-frameworks
+++ b/build/ios/build-frameworks
@@ -17,26 +17,26 @@ EXTLIBRARIES="libevent libevent_pthreads libevent_openssl libcrypto libssl libGe
 
     # Copying x86 headers. This has no implications because measurement-kit
     # uses no machine-dependent headers.
-    cp -Rp tmp/iPhoneSimulator/i386/include/measurement_kit \
+    cp -Rp tmp/iphonesimulator/i386/include/measurement_kit \
         Frameworks/measurement_kit.framework/Headers
 
     # Lipo external libraries
     for lib in $EXTLIBRARIES; do 
         lipo -create -output Frameworks/$lib.framework/$lib \
-          tmp/iPhoneOS/arm64/lib/$lib.a \
-          tmp/iPhoneOS/armv7s/lib/$lib.a \
-          tmp/iPhoneOS/armv7/lib/$lib.a \
-          tmp/iPhoneSimulator/i386/lib/$lib.a \
-          tmp/iPhoneSimulator/x86_64/lib/$lib.a
+          tmp/iphoneos/arm64/lib/$lib.a \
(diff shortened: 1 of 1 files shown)


bassosimone:
> ...
> (note: this branch is actually based against stable)

Yeah, makes sense. I'll forward port and cherry-pick this also to master.

I am sorry for all of that. Many years ago I freaked out against macOS (at the time still MacOS, I was using a PPC) because I discovered this case insensitive feature the hard way (having created a file named `CONFIGURE` as a backup for `configure` &mdash; bad idea), and installed Arch Linux for PPC as a result (that was fun, the installer crashed and I had to edit it as a script &mdash; luckily it was not using some sort of `systemd-installd` &mdash; using the true standard editor). Then, I assume that probably my brain always things in terms of real case sensitive file systems and so I overlooked the problem.

> "Anything uncatchable by my net is ipso facto outside the scope of icthyological knowledge. In short, "what my net can't catch isn't fish." 
> [Arthur Eddington](https://en.wikiquote.org/wiki/Arthur_Eddington)

bassosimone:
For the records,

> configure: WARNING: using cross tools not prefixed with host triplet

this warning is innocuous and can be ignored. When cross compiling, `configure` is complaining that it does not find tools like `ld` prefixed by host triplets (i.e. it finds `ld` and not `arm-apple-darwin-ld`. This has to do with the way in which Xcode is packaged and installed.

PR #1090, opened 2017-01-20
var.hpp: fix as() to use dynamic cast

bassosimone:
When developing a diff to improve NDT, I mistakenly wrote:

```C++
ErrorOr<std::vector<double>> ctimes = net::get_connect_times(err);
```

instead of:

```C++
ErrorOr<double> ctime = net::get_connect_time(err);
```

after `connect()`. Now, both functions internally use Var<T>::as()
to cast from a generic `ErrorContext` to a specific context.

The net effect of calling `get_connect_times` rather than the more
humble `get_connect_time` is that `ErrorContext` is downcasted to
`ConnectManyResult` rather than to `ConnectResult`. Since the cast
is not possible, one would have expected the cast to fail and the
resulting Var<T> to hold a nullptr. Hence the `ErrorOr` should not
be a value but rather an error. Right?

Wrong! What I obtained instead was a segfault.

It turns out that using a static-pointer cast in `as()` was a bad
idea, and that what I really would have wanted to have is a dynamic
cast instead. The dynamic cast, in fact, fails (as I expected) if
it is not possible to perform the cast.

### DIFF
--- a/doc/api/common/var.md
+++ b/doc/api/common/var.md
@@ -55,6 +55,12 @@ The `as()` method casts the pointee to type `R` if possible. If conversion
 is not possible, the returned `Var<>` would point to a null pointer and hence
 attempting to dereference it would result in `std::runtime_error`.
 
+# CAVEAT
+
+The `as()` method does not seem to preserve any custom deleter that may
+have been passed to the original `Var<>`; this seems reasonable since the
+new `Var<>` will delete another type of object (if the cast worked).
+
 # HISTORY
 
 The `Var` template class appeared in MeasurementKit 0.1.0.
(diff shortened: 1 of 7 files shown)


bassosimone:
Looking again at the size of the diff, I guess this qualifies as hotfix:

1. there is just one *relevant* line in the diff:

```diff
@@ -30,7 +30,7 @@ template <typename T> class Var : public std::shared_ptr<T> {
     }
 
     template <typename R> Var<R> as() {
-        return std::static_pointer_cast<R>(*this);
+        return std::dynamic_pointer_cast<R>(*this);
     }
  
    protected:
```

2. all the remainder is just trivially adapting the code or regress tests

3. regress tests now do check that now, when cast is not possible, `as()` returns a `Var<T>` wrapping a `nullptr`

For these reasons, I am going to merge. (I have squashed locally and have also verified locally that the between before and after squashing is empty.)

PR #1096, opened 2017-01-22
Make sure http template produces good json

### LINKED ISSUE #1065: ooni: sanitize http headers and response line
Steps to reproduce:

- build measurement-kit v0.4.0-alpha
- ./measurement_kit web_connectivity -v XYZ

where XYZ contains:

```
http://www.zoomshare.com/
```

Then try to parse the resulting JSON. It would fail because there is a header that is not UTF-8.

How ooni-probe deals with the same issue:

- https://github.com/TheTorProject/ooni-probe/blob/master/ooni/templates/httpt.py#L46

bassosimone:
Closes #1065. Specifically:

1) change name from represent_http_body to represent_string since
   the function is now used more generally than for the body

2) use represent_string to also represent the response line and the
   value of headers that both could be non UTF-8

There is no need to represent the name of the headers because
the standard says that it's

>
>    header-field   = field-name ":" OWS field-value OWS
>    field-name     = token
>
>    token = 1*tchar
>
>    tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*"
>                / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
>                / DIGIT / ALPHA
>                ; any VCHAR, except delimiters
>
>    ALPHA =  %x41-5A / %x61-7A   ; A-Z / a-z
>    DIGIT =  "0" / "1" / "2" / "3" / "4" / "5" / "6" /
>                   "7" / "8" / "9"
>

So, we really don't have UTF-8 / non UTF-8 issues for names.

See:

- https://tools.ietf.org/html/rfc7230#section-3.2
- https://tools.ietf.org/html/rfc5234#appendix-B.1

### DIFF
--- a/doc/api/ooni/utils.md
+++ b/doc/api/ooni/utils.md
@@ -25,7 +25,7 @@ void resolver_lookup(
         Var<Logger> logger = Logger::global()
 );
 
-report::Entry represent_http_body(const std::string &body);
+report::Entry represent_string(const std::string &s);
 
 } // namespace ooni
 } // namespace mk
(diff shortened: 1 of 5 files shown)


hellais:
This looks good to me.

bassosimone:
Thanks!

PR #1097, opened 2017-01-23
Use default backends when it makes sense

bassosimone:
Basically always except in the regress tests. Refactor regress tests
a bit to make sure it is not an hell to keep them up to date.


hellais:
The diff looks good to me. I will do some local smoke testing, but it's good to merge for me.

bassosimone:
Thank you!

bassosimone:
@hellais do you think we can bless this?

hellais:
> @hellais do you think we can bless this?

Yes please proceed with merging this.

Issue #935, opened 2016-11-18
Proposals for improving errors


bassosimone:
In this issue I shall summarize all the aspects of the current error systems that I don't like and will sketch out how I propose to fix them. Then I will also discuss things that I like and I'd expand.

# Annoying issues

## Error codes are explicit and updated manually

This is really annoying and error prone. The easy fix for this issue is to use enumerations. More specifically we can probably use the `enum class` construct. For example:

```C++
namespace mk {
namespace dns {

enum class ErrorCode {
    FORMAT_ERROR,
    SERVER_FAILED_ERROR,
    NOT_EXIST_ERROR,
    /* ... */
};

/* ... */

MK_DEFINE_ERR(MK_ERR_DNS(FORMAT_ERROR), FormatError, "dns_lookup_error")
MK_DEFINE_ERR(MK_ERR_DNS(SERVER_FAILED_ERROR), ServerFailedError, "dns_lookup_error")
MK_DEFINE_ERR(MK_ERR_DNS(NOT_EXIST_ERROR), NotExistError, "dns_lookup_error")

/* ... */
}}
```

## The category is encoded along with the error

Currently we encode error code and error category (e.g. `dns` or `net` together). I think it would be better to encode the category separately. The function to check for error equality could then become:

```C++
bool Error::operator==(Error &other) {
    return this->category == other.category && this->code == other.code;
}
```

This is a good place to mention that the current implementation of `Error` derives from `std::exception` because I believe it's handy to throw/catch errors sometimes. Given what I'm describing here, however, it would be worth investigating whether it would make sense to use [C++11 error_condition](http://en.cppreference.com/w/cpp/error/error_condition). It would also be worth investigating whether we can use not only the error condition but also the error category here.

## We do not map low-level errors

In most cases, when our code fails, we do not report the lowest-level error that occurred (for example, we see a `recv()` failing but we report something like "recv failed" rather than "host unreachable"). We can improve this situation by using the [portable errors defined by the standard C++11 library](http://en.cppreference.com/w/cpp/error/errc) that include many interesting errors including for example "host unreachable", "connection reset by peer", etc.

## No clear distinction between description and OONI errors

OONI defines errors as strings. We construct errors with macros where we specify the error code and the corresponding OONI error. We should extend this model to include, in addition to OONI error, a more descriptive explanation of the error itself. In many cases, OONI errors map to `undefined_error $code`, which is fine for OONI purposes but not that clear for the user.

[...]

PR #1098, opened 2017-01-23
Make sure network errors have a meaning

bassosimone:
While investigating measurement-kit/ooniprobe-ios#61, I realized it
is now time to start improving the error codes.

This pull request starts implementing some ideas in #935 to make
at least errors emitted by the network layer much more clear.

Specifically, I've chosen to rely on the C++ library standard `errc`
enum (and to the corresponding names) for mapping errno.

To be honest, I dunno whether this is portable to Windows, but
the `std::errc` enum is described as `C++11` at:

    http://en.cppreference.com/w/cpp/error/errc

Moreover, libcxx <system_error> actually defines some of the errors
that we are interested to map in here:

    https://github.com/llvm-mirror/libcxx/blob/a7ae303de36dfa1e0a165c52cd073c51dc85e8cb/include/system_error#L262

So, this seems to me more solid that just peeking at `errno`.


bassosimone:
I think we can close this PR. The most interesting part of it, the one in which we peek from `std::errc` has now been merged. The remainder is something not so interesting.

Issue #919, opened 2016-11-14
regress: avoid using global reactor whenever possible


bassosimone:
We currently use the global reactor for many regress tests, but we should not.

The problem is that, if an exception is thrown (and Catch checks are exceptions), we basically exit from the reactor the expedite way and, since the reactor is actually a libevent loop, it's not possible to restart again the poller because the event loop is not exited cleanly and would see the subsequent call to start the loop as a reentrant invocation.

This is bad because, if a test fails, all subsequent tests within the same test executable are going to fail because of reentrant invocation.

I propose to fix this issue by always using a custom poller for each test case, unless it does not make sense to do so (i.e. when we're testing the global poller), such that failure in one test case do not affect other test cases.

bassosimone:
Fixed in #1346

PR #1099, opened 2017-01-24
test/net/connect.cpp: refactor to avoid global poller

bassosimone:
This way, a test failing does not cause the failure of all the
following tests and we can see easily what has failed.

Related to #919

### DIFF
--- a/build/valgrind/openssl.supp
+++ b/build/valgrind/openssl.supp
@@ -10,7 +10,7 @@
    fun:_ZN2mk8libevent11poller_loopIXadL_Z9event_newEEXadL_Z10event_freeEEXadL_Z9event_addEEXadL_Z19event_base_dispatchEEEEvNS_3VarI10event_baseEEPNS0_6PollerE
    fun:_ZN2mk8libevent6Poller4loopEv
    fun:_ZN2mk7Reactor23loop_with_initial_eventESt8functionIFvvEE
-   fun:_ZN2mk23loop_with_initial_eventESt8functionIFvvEENS_3VarINS_7ReactorEEE
-   fun:_ZL31____C_A_T_C_H____T_E_S_T____418v
+   fun:_ZL31____C_A_T_C_H____T_E_S_T____440v
    fun:_ZN5Catch8runTestsERKNS_3PtrINS_6ConfigEEE
+   fun:main
 }
--- a/test/net/connect.cpp
+++ b/test/net/connect.cpp
@@ -29,14 +29,15 @@ struct bufferevent;
 static int fail(const char *, sockaddr *, int *) { return -1; }
 
 TEST_CASE("connect_base deals with evutil_parse_sockaddr_port error") {
-    loop_with_initial_event([]() {
+    Var<Reactor> reactor = Reactor::make();
+    reactor->loop_with_initial_event([=]() {
         connect_base<fail>("130.192.16.172", 80,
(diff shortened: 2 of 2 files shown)


build/valgrind/openssl.supp
```diff
@@ -10,7 +10,7 @@
    fun:_ZN2mk8libevent11poller_loopIXadL_Z9event_newEEXadL_Z10event_freeEEXadL_Z9event_addEEXadL_Z19event_base_dispatchEEEEvNS_3VarI10event_baseEEPNS0_6PollerE
    fun:_ZN2mk8libevent6Poller4loopEv
    fun:_ZN2mk7Reactor23loop_with_initial_eventESt8functionIFvvEE
-   fun:_ZN2mk23loop_with_initial_eventESt8functionIFvvEENS_3VarINS_7ReactorEEE
-   fun:_ZL31____C_A_T_C_H____T_E_S_T____418v
+   fun:_ZL31____C_A_T_C_H____T_E_S_T____440v
    fun:_ZN5Catch8runTestsERKNS_3PtrINS_6ConfigEEE
+   fun:main
```
bassosimone: I needed to update the suppression file, because the location from which the error was triggered has now changed slightly and we need to inform Valgrind about that.

test/net/connect.cpp
```diff
@@ -179,187 +187,198 @@ TEST_CASE("connect_base works with ipv6") {
...
-        connect_logic("www.google.com", 81, [](Error e, Var<ConnectResult> r) {
+    Var<Reactor> reactor = Reactor::make();
+    reactor->loop_with_initial_event([=]() {
+        connect_logic("www.google.com", 81, [=](Error e, Var<ConnectResult> r) {
             REQUIRE(e);
             REQUIRE(r->connected_bev == nullptr);
-            break_loop();
-        });
+            reactor->stop();
+        }, {{"net/timeout", 1.0}}, reactor);
```
bassosimone: Here, I have added a smaller timeout than before to make the test quicker.

test/net/connect.cpp
```diff
@@ -179,187 +187,198 @@ TEST_CASE("connect_base works with ipv6") {
...
                              });
                          }
-                     });
+                     }, {}, reactor);
     });
 }
 
 TEST_CASE("net::connect() works with a custom reactor") {
     // Note: this is how Portolan uses measurement-kit
-    set_verbosity(MK_LOG_DEBUG);
```
bassosimone: Made the test less verbose. It was really too much verbose.

test/net/connect.cpp
```diff
@@ -368,93 +387,100 @@ TEST_CASE("net::connect() works with a custom reactor") {
...
-                 {"net/ca_bundle_path", "test/fixtures/certs.pem"}});
+                 {"net/ca_bundle_path", "test/fixtures/certs.pem"}},
+                reactor);
     });
 }
 
 TEST_CASE("net::connect() ssl fails when presented a certificate with the "
           "wrong hostname") {
-    loop_with_initial_event([]() {
-        set_verbosity(MK_LOG_DEBUG);
```
bassosimone: Re: verbosity, same as above

bassosimone:
This clearly qualifies as a hotfix, because it only slightly modifies the regress tests. I have tested in the context of #1098 that this diff improve the situation as far as trying to understand what error did broke the unittest is concerned. I will merge as soon as travis-ci gives me green lights.

PR #1102, opened 2017-01-24
connect: support also ipv6 link local with scope

bassosimone:
Run log on a box where `nc -vl fe80::1%lo0 8080`:

```
$ ./measurement_kit net_connect -vvp 8080 fe80::1%lo0
[D] resolve_hostname: fe80::1%lo0
[D] resolve_hostname: is valid ipv6
[D] connect_first_of begin
[D] connect_base [fe80::1%lo0]:8080
[D] connect() in progress...
[D] connect time: 0.000344
[D] connect_first_of success
```

### DIFF
--- a/build/valgrind/openssl.supp
+++ b/build/valgrind/openssl.supp
@@ -10,7 +10,7 @@
    fun:_ZN2mk8libevent11poller_loopIXadL_Z9event_newEEXadL_Z10event_freeEEXadL_Z9event_addEEXadL_Z19event_base_dispatchEEEEvNS_3VarI10event_baseEEPNS0_6PollerE
    fun:_ZN2mk8libevent6Poller4loopEv
    fun:_ZN2mk7Reactor23loop_with_initial_eventESt8functionIFvvEE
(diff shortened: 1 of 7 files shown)


bassosimone:
I've moved this back into the ready for review queue, [because the Valgrind test on travis-ci is failing](https://travis-ci.org/measurement-kit/measurement-kit/builds/195083861) .

Interestingly, I've seen no issues with circle-ci, which also runs Valgrind checks. Now, I'm also running a third Valgrind locally. I wonder why this happens only on travis.

Anyway, this looks like something worth investigating *after* v0.4.0-alpha.2 is out.

bassosimone:
*Update*: the build is now fixed. I rewrote the code differently. It cannot be flagged as ready for review yet because there are little tweaks that I need to apply first:

- [ ] correct a couple of typos in comments
- [ ] make sure indent is correct
- [ ] fix regress to actually have a test case with numeric port

After that, this branch can be blessed as "ready".

bassosimone:
Thank you!

PR #1104, opened 2017-01-25
Merge stable

bassosimone:


src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -260,68 +263,17 @@ void Runnable::begin(Callback<Error> cb) {
...
-                        // http://en.cppreference.com/w/cpp/algorithm/shuffle:
-                        std::random_device rd;
-                        std::mt19937 g(rd());
-                        std::shuffle(inputs.begin(), inputs.end(), g);
-                    }
-                } else {
-                    // Empty string to call main() just once
-                    inputs.push_back("");
-                    num_entries = 1;
-                }
```
bassosimone: *TODO*: make sure this part did not contained any changes that we lost.

test/nettests/utils.hpp
```diff
@@ -0,0 +1,48 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
```
bassosimone: The name of this file *may* be a bit misleading; it's not the header for `test/nettests/utils.cpp`

bassosimone:
Okay. I've read the diff once more and I am convinced that no artefacts have been included during the merge. Since I don't like keeping huge diffs floating and now I am fresh with recent changes that are only going to fade away in the future, I'll proceed with merging this.

PR #1106, opened 2017-01-27
templates::http_request(): scrub IP address

### LINKED ISSUE #1063: Users IP address should be stripped from the headers and HTTP response
Similarly to what we do in ooniprobe, when running measurement-kit we should also strip the users IP address (if include-ip is set to false) from the HTTP response body and HTTP headers.

bassosimone:
Based on #1107.

Closes #1063.

### DIFF
--- a/src/libmeasurement_kit/nettests/runnable.cpp
+++ b/src/libmeasurement_kit/nettests/runnable.cpp
@@ -114,7 +114,7 @@ void Runnable::geoip_lookup(Callback<> cb) {
 
     // This is to ensure that when calling multiple times geoip_lookup we
     // always reset the probe_ip, probe_asn and probe_cc values.
-    probe_ip = "127.0.0.1";
+    real_probe_ip = probe_ip = "127.0.0.1";
     probe_asn = "AS0";
     probe_cc = "ZZ";
 
@@ -138,6 +138,8 @@ void Runnable::geoip_lookup(Callback<> cb) {
 
     // No need to perform further lookups if we don't need to save anything
     if (not save_ip and not save_asn and not save_cc) {
+        logger->warn("Not knowing user_ip means we cannot attempt to scrub it "
+                     "from the report");
         cb();
         return;
     }
@@ -146,6 +148,8 @@ void Runnable::geoip_lookup(Callback<> cb) {
         [=](Error err, std::string ip) {
             if (err) {
                 logger->warn("ip_lookup() failed: error code: %d", err.code);
+                logger->warn("Not knowing user_ip means we cannot attempt "
+                             "to scrub it from the report");
                 cb();
                 return;
             }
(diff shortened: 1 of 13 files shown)


src/libmeasurement_kit/ooni/templates.cpp
```diff
@@ -76,12 +79,26 @@ void http_request(Var<Entry> entry, Settings settings, http::Headers headers,
         settings["http/method"] = "GET";
     }
 
+    auto redact = [=](std::string s) {
+        if (!settings.get("save_real_probe_ip", false)) {
+            s = mk::ooni::scrub(s, probe_ip);
```
hellais: Initially I was tipped off by the fact that you were calling this `probe_ip` in here and thought you were passing in `probe_ip` instead of `real_probe_ip`. Properly inspecting the diff I realised that is not the case, but I would advise changing this name to `real_probe_ip` to avoid confusion.

hellais:
Summarising what was discussed on IRC:

We have come to the conclusion that the correct place to do scrubbing (and also encoding) is at the level of the nettest:Runnable by means of defining the fields that should be scrubbed and encoded inside of the metadata of the test and test template.

Since this requires a considerable amount of reworking we are going to postpone this to future versions and for the moment just do the following:

* [ ] Rename `probe_ip` to `real_probe_ip`
* [ ] Pass `real_probe_ip` down the chain via `Settings` so that we don't impact the public API

For the above work I am going to file a ticket.

src/libmeasurement_kit/ooni/templates.cpp
```diff
@@ -76,12 +78,31 @@ void http_request(Var<Entry> entry, Settings settings, http::Headers headers,
         settings["http/method"] = "GET";
     }
 
+    /*
+     * XXX probe ip passed down the stack to allow us to scrub it from the
+     * entry; see issue #1110 for plans to make this better.
+     */
+    std::string probe_ip = settings.get("real_probe_ip_", std::string{});
```
bassosimone: @hellais we initially discussed about renaming this `real_probe_ip` and that was sensible because the variable was passed down the stack. Since now is local and anyway the key passed down the stack is clear, I did not felt the need to also rename the variable (thus also allowing to stay within column 80).

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -154,6 +158,13 @@ void Runnable::geoip_lookup(Callback<> cb) {
                 logger->debug("saving user's real ip on user's request");
                 probe_ip = ip;
             }
+            /*
+             * XXX Passing down the stack the real probe IP to allow
+             * specific tests to scrub entries.
+             *
+             * See also measurement-kit/measurement-kit#1110.
+             */
+            options["real_probe_ip_"] = ip;
```
hellais: I assume the extra `_` at the end is used to indicate this is a "private" key.

bassosimone: Yes, indeed!

bassosimone:
Excellent, thank you!

PR #1115, opened 2017-02-01
The header key comparison MUST be done with the lowercase version

hellais:
Doing it with the normalised case version (Foo-Bar) will lead to the
intersection with "common headers" always being non null and hence we
will always believe that the headers being looked at are "uncommon" and
lead to MANY false negatives.


bassosimone:
After the merge, nothing is left in this pull request (I cherry picked the relevant commits into another integration branch that has been merged to master), for this reason I'm closing this PR.

PR #1118, opened 2017-02-02
Fix uninitialized values according to Valgrind

bassosimone:
See https://travis-ci.org/measurement-kit/measurement-kit/jobs/197586498#L2132


bassosimone:
Somehow I failed to add an explanation for this: we have already seen in the past that strlen() in GNU C library reads input word after word rather than byte after byte. Hence, it seems to be fair to classify this warning reported by Valgrind as a false positive ad add a suppression for it.

See: https://sourceware.org/git/?p=glibc.git;a=blob;f=string/strlen.c;h=b368ccf58fee2e686b9623b7895200afc23acf5e;hb=HEAD

PR #1124, opened 2017-02-07
http: fix schema less redirect

bassosimone:
Based on #1121


bassosimone:
I'm going to squash this into stable and then try to understand the circle-ci breakage (which probably, however, is best dealt with in master).

PR #1132, opened 2017-02-13
Allow to specify tests input directly

bassosimone:
This is all what is needed on the MK side to allow a user of
ooniprobe-{android,ios} to test for a specific URL.


hellais:
Does randomisation of inputs also apply to inputs added via `add_input`? Glancing quickly at the code I don't see why it wouldn't but just checking if that is indeed the case.

hellais:
I would be great if we also had some basic unittests for this functionality.

bassosimone:
> Does randomisation of inputs also apply to inputs added via add_input?

Actually no. This diff was meant as an hotfix for `stable`, then it was not merged and this allows us to do some refactoring needed to also implement randomisation for input added like this.

I can see an argument for not randomising the list of input provided via `add_input` (i.e. one has already selected and randomised specific URLs) but I guess in that case this can be implemented by using the setting that disables randomisation.

bassosimone:
Given that the target has changed and we have more room for refactoring, I believe this PR needs to return back to a `in progress` state. Doing it right now.

bassosimone:
Superseded by #1273. Closing.

PR #1134, opened 2017-02-13
Add a name to each error

bassosimone:
Make sure different errors has different names.


hellais:
By mapping values that were previously assigned to something we are breaking some checks in web_connectivity, namely:

* https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/web_connectivity.cpp#L150

* https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/web_connectivity.cpp#L313

Also mapping no_error leads this to break:

* https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/web_connectivity.cpp#L294

hellais:
I would suggest that in order to merge this we do the following:

1. For every error that is already mapped we leave it untouched

2. We either revert no_error to be `""` or we change web_connectivity to check for `NoError()`

bassosimone:
@hellais Thanks, I agree! I've implemented what you suggested and tweaked the diff such that it's most simple to go through it and see what has been changed.

bassosimone:
I have rebased and locally checked the rebase did not introduce artifacts.

PR #1141, opened 2017-02-14
meek fronting nettest

joelanders:
Drops a nice-looking report:
```
{
  "measurement_start_time": "2017-02-14 17:57:16",
  "test_start_time": "2017-02-14 17:57:15",
  "data_format_version": "0.2.0",
  "software_name": "measurement_kit",
  "report_id": "20170214T180045Z_AS0_8MuSUtM0H97TUS5xj1PCJzTV6TPgmu9sVi3cEJKIO0pFOfiMod",
  "input": "a0.awsstatic.com:d2zfqthxsdq309.cloudfront.net",
  "test_runtime": 0.189069032669067,
  "software_version": "0.5.0-dev",
  "test_keys": {
    "requests": [
      {
        "method": "GET",
        "response": {
          "body": "I\u2019m just a happy little web server.\n",
          "response_line": "HTTP/1.1 200 OK",
          "code": 200,
          "headers": {
            "Via": "1.1 d6fa2e1de8f392301c10fd5bb7b263c3.cloudfront.net (CloudFront)",
            "X-Amz-Cf-Id": "nS80GhpowWzU-aO49tTYUFBU4P8oysxKgd8chMpl4tyfOQPXXmphnA==",
            "X-Cache": "Hit from cloudfront",
            "Content-Type": "text/plain; charset=utf-8",
            "Date": "Mon, 13 Feb 2017 23:41:42 GMT",
            "Age": "65825",
            "Content-Length": "38",
            "Connection": "keep-alive"
          }
        },
        "failure": null,
        "request": {
          "body": "",
          "method": "GET",
          "url": "https://a0.awsstatic.com",
          "headers": {
            "Host": "d2zfqthxsdq309.cloudfront.net"
          }
        }
      }
    ],
    "socksproxy": null,
    "client_resolver": "127.0.0.1",
    "agent": "agent"
  },
  "test_version": "0.0.1",
  "test_name": "meek_fronting",
  "probe_asn": "AS0",
  "probe_cc": "ZZ",
  "probe_ip": "127.0.0.1"
}
```


src/libmeasurement_kit/ooni/meek_fronting.cpp
```diff
@@ -0,0 +1,67 @@
...
+    std::string outer_host = "https://" + outer_inner.front();
+    std::string inner_host = "https://" + outer_inner.back();
+
+    ErrorOr<http::Url> outer_url = mk::http::parse_url_noexcept(outer_host);
+    ErrorOr<http::Url> inner_url = mk::http::parse_url_noexcept(inner_host);
+
+    if (!outer_url || !inner_url) {
+        logger->debug("Invalid url: '%s' or '%s'", outer_host.c_str(),
+                      inner_host.c_str());
+        callback(entry);
```
bassosimone: Here I think it would be nicer to do:

```C++
reactor->call_soon([=]() { callback(entry); });
```

This will *schedule* calling `callback(entry)` during the next I/O loop.

I find it nicer because the caller has the guarantee that the callback will always execute after the function to which he/she passed the callback to has returned.

I also think that probably in this case you would like to set the `failure` of the entry.

bassosimone:
Great! :-)

src/libmeasurement_kit/nettests/meek_fronting.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <measurement_kit/nettests.hpp>
+#include <measurement_kit/ooni.hpp>
+
+namespace mk {
+namespace nettests {
+
+MeekFrontingTest::MeekFrontingTest() : BaseTest() {
+    runnable.reset(new MeekFrontingRunnable);
+    runnable->test_name = "meek_fronting";
```
hellais: The correct test_name is: `meek_fronted_requests`. TBH I never really liked this name that much, but it's what we have 🤕

src/libmeasurement_kit/nettests/meek_fronting.cpp
```diff
@@ -0,0 +1,24 @@
...
+    runnable->test_version = "0.0.1";
+    runnable->needs_input = true;
```
hellais: I believe this should be false.

hellais:
It's not writing the report in the correct format.

Two things need to happen:

1. The input needs to be a list with `['FRONT', 'FRONTED_DOMAIN']`

2. It should write the `test_keys->success` value to be true or false depending on whether or not it succeeds to connect to it and the body matches (https://github.com/TheTorProject/ooni-probe/blob/master/ooni/nettests/blocking/meek_fronted_requests.py#L69)

joelanders:
@hellais 
1. You mean split the `host1:host2` (from the file input) into `["host1", "host2"]` before passing it to the test implementation in `src/libmeasurement_kit/ooni/meek_fronting.cpp`? I think `process_input_filepaths_impl()` expects each line of an input file to be an input.

2. (acknowledged)

src/libmeasurement_kit/nettests/meek_fronting.cpp
bassosimone: Uhm, to my understanding, the meek test has input, no?

bassosimone:
@joelanders @hellais: regarding 1., it seems to me that the comment is more related to the `input` key inside the final report than the way in which it is passed around, correct? I can also confirm that `process_input_filepaths_impl()` deals with lines.

joelanders:
@bassosimone @hellais got it, that makes sense. It's also a bit hard to change, because `entry["input"]` is set (in `src/libmeasurement_kit/nettests/runnable.cpp`) in a callback that runs after everything in `mk::ooni::meek_fronting` (src/libmeasurement_kit/ooni/meek_fronting.cpp). If we change the behavior in `Runnable`, we might have to change every subclass...

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -79,7 +79,11 @@ void Runnable::run_next_measurement(size_t thread_id, Callback<Error> cb,
         report::Entry entry;
         entry["test_keys"] = *test_keys;
         entry["test_keys"]["client_resolver"] = resolver_ip;
-        entry["input"] = next_input;
+        if (!(*test_keys)["input"].is_null()) {
+            entry["input"] = (*test_keys)["input"];
+        } else {
+            entry["input"] = next_input;
+        }
```
bassosimone: As an alternative solution, what about this: there is a method in the base class that takes in input the `next_input` and returns in output a `report::Entry`. Such method has as default implementation:

```C++
report::Entry BaseClass::method_name(const std::string &next_input) {
    return next_input; /* Works because a string can be used to construct a valid JSON */
}
```

Then, only for this test, we override said method to construct the input representation we want.

Finally, in the code here we do:

```C++
    entry["input"] = method_name(next_input);
```

?

src/libmeasurement_kit/ooni/constants.hpp
```diff
@@ -56,6 +56,8 @@ const http::Headers COMMON_CLIENT_HEADERS = {
   }
 };
 
+const std::string MEEK_SERVER_RESPONSE = "I\u2019m just a happy little web server.\n";
```
bassosimone: Is `\u2019` a unicode character? If so, how wide are characters in this string? Are they UTF-16?

(I would have expected the server to return UTF-8, so I'm asking &mdash; and I guess that in UTF-8 that apostrophe is composed by two 8 bit characters, no?)

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    if (!options["outer_host"].empty() &&
+        !options["inner_host"].empty()) {
+        outer_host = options["outer_host"];
+        inner_host = options["inner_host"];
+    } else {
+        std::list<std::string> outer_inner = split(input, ":");
+        if (outer_inner.size() != 2) {
+            logger->debug("Couldn't split input: %s", input.c_str());
```
bassosimone: In this case you should set `entry["failure"] = SomeError()` (perhaps ValueError())

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,93 @@
...
+            return;
+        }
+        // XXX: We should make sure that we remove leading and trailing whitespaces
+        outer_host = outer_inner.front();
+        inner_host = outer_inner.back();
+    }
+
+    // url parsing methods require a schema
+    outer_host.insert(0, "https://");
+    inner_host.insert(0, "https://");
```
bassosimone: If what you need to parse is an endpoint (i.e. `<ADDRESS_OR_HOST>:<PORT>`, [we have code for doing that](https://github.com/measurement-kit/measurement-kit/blob/master/include/measurement_kit/net/utils.hpp#L12).

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    ErrorOr<http::Url> outer_url = mk::http::parse_url_noexcept(outer_host);
+    ErrorOr<http::Url> inner_url = mk::http::parse_url_noexcept(inner_host);
+
+    if (!outer_url || !inner_url) {
+        logger->debug("Invalid url: '%s' or '%s'", outer_host.c_str(),
+                      inner_host.c_str());
+        callback(entry);
```
bassosimone: Also here you should set the `failure` field of the entry.

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,93 @@
...
+    // url parsing methods require a schema
+    outer_host.insert(0, "https://");
+    inner_host.insert(0, "https://");
+
+    ErrorOr<http::Url> outer_url = mk::http::parse_url_noexcept(outer_host);
+    ErrorOr<http::Url> inner_url = mk::http::parse_url_noexcept(inner_host);
+
+    if (!outer_url || !inner_url) {
+        logger->debug("Invalid url: '%s' or '%s'", outer_host.c_str(),
+                      inner_host.c_str());
```
bassosimone: In case of errors like this, it's perhaps best to use `logger->warning`

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,93 @@
...
+        callback(entry);
+        return;
+    }
+
+    (*entry)["input"] = {outer_url->address, inner_url->address};
+
+    options["http/url"] = "https://" + outer_url->address;
+    options["net/ssl"] = true;
```
bassosimone: I *think* you don't need to specify `net/ssl` if the URL is `https`

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    templates::http_request(entry, options, headers, body,
+                            [=](Error err, Var<http::Response> response) {
+                                if (err) {
+                                    logger->debug(
+                                        "meek_fronted_requests: http-request error: %s",
+                                        err.explain().c_str());
+
+                                    (*entry)["meek_fronted_requests"] =
+                                        err.as_ooni_error();
```
bassosimone: Shouldn't this set the `failure` key?

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,93 @@
...
+                                }
+
+                                if (!response) {
+                                    logger->warn("null response");
+                                    callback(entry);
```
bassosimone: Uhm, here you end up calling `callback` twice. Perhaps it's best, given the way you're structuring the code, to refactor the branch to consider the `!!response` case and only leave the final invocation of callback at bottom?

bassosimone: Also: I'm not sure it's written anywhere, but: if there is no error, you can assume you have a valid response. In some cases, on error, the response is currently `nullptr`. I was thinking, however, to move to a situation where we guarantee that we never return `nullptr` (even if wrapped by `Var<>`), such that, if that happens, is a library bug causing an exception triggered by `Var<>` when you attempt to access the pointer, and one does not need to code to also consider the case in which it's null. Does this make sense?

src/measurement_kit/cmdline/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include <measurement_kit/http.hpp>
+
+#include <iostream>
+
+namespace mk {
+namespace cmdline {
+namespace meek_fronted_requests {
+
+static const char *kv_usage =
+    "usage: measurement_kit meek_fronted_requests [-v] input_file \n";
```
bassosimone: Note: now that #1156 has been merged, it would be optimal to refactor this file to be consistent with the new command line coding practices.

src/measurement_kit/cmdline/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,83 @@
...
+        while (argc > 0) {
+            test.add_input_filepath(argv[0]);
+            argc -= 1, argv += 1;
+        }
+    }
+
+    test
+        .set_options("expected_body", expected_body)
+        .set_options("outer_host", outer_host)
+        .set_options("inner_host", inner_host)
```
bassosimone: Uhm, I think it would perhaps be better, for now, to pass in input just a file containing the list of tests to be run as other tests do. WIP to pass direct input to tests is in #1132 but it's not complete yet (maybe we can work together to have it finished?). Anyway, even if we have functionality to pass direct input to test, it would seem more natural to me to specify input in the same format in which it would be specified in a file (that, according to my understanding, is `endpoint:endpoint`), unless OONI does that differently.

joelanders: Yeah, it'll read lines from an input file as inputs *unless* you specify some of these options. I wanted a way to run the test without an input file, but maybe we don't need that.

src/measurement_kit/cmdline/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,83 @@
...
+    if ((inner_host.empty() && !outer_host.empty()) ||
+        (outer_host.empty() && !inner_host.empty())) {
+        std::cout << "If you specify one of {outer,inner}_host, "
+                     "you must specify both.\n";
+        std::cout << kv_usage;
+        exit(1);
+    }
+
+    if (inner_host.empty() && outer_host.empty()) {
+        test.runnable->needs_input = true;
```
bassosimone: Ah, now I better understand the context. So, basically, look at #1132: the idea is to add there a function called `add_input()` with which you pass input to be processed in addition to input available via file.

With such API in place, it's easier because one is not forced to fiddle with `needs_input`.

src/measurement_kit/cmdline/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,83 @@
...
+        if (argc < 1) {
+            std::cout << kv_usage;
+            exit(1);
+        }
+        while (argc > 0) {
+            test.add_input_filepath(argv[0]);
+            argc -= 1, argv += 1;
+        }
+    }
```
bassosimone: ^^^ This comment is probably not totally accurate, because, at first, I failed to understand exactly why these checks where performed here, sorry.

src/libmeasurement_kit/ooni/constants.hpp
joelanders:
```python
>>> bytes("I\u2019m",encoding="utf8")
b'I\xe2\x80\x99m'
```
Is it better to write this string as `'I\xe2\x80\x99m`..., then?

src/libmeasurement_kit/nettests/runnable.cpp
joelanders: The Test (subclass) has a reference to the Runnable (subclass), but not the other way around. (include/measurement_kit/nettests/base_test.hpp)

 The code in question is in the Runnable base class. It sounds like you want to add an overridable method to the Test classes, but I'm not sure where this Runnable code would call that from (we don't seem to have a reference to a Test in this code).

Am I missing something?

src/libmeasurement_kit/ooni/constants.hpp
bassosimone: I don't know. I guess, to reply the first thing I need to do is to understand what unicode literals do to a C string, because honestly I don't know that :-).

src/measurement_kit/cmdline/meek_fronted_requests.cpp
bassosimone: It is certainly useful to do that. My point was mostly that this feature was already being developed in a different way (via `set_input()`) so perhaps it was best not to include the feature in this pull request and finish the other pull request so all tests have it.

src/libmeasurement_kit/nettests/runnable.cpp
bassosimone: > Am I missing something?

No, what you say is correct. I guess my original comment was incorrect in that I said `BaseClass`, which was a bit generic/misleading, and it would have been more correct to say that I was advocating for implementing this feature in `Runnable.`

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -77,9 +77,15 @@ void Runnable::run_next_measurement(size_t thread_id, Callback<Error> cb,
     logger->debug("net_test: running with input %s", next_input.c_str());
     main(next_input, options, [=](Var<report::Entry> test_keys) {
         report::Entry entry;
+        // simple way to override the default entry["input"] behavior
+        if (!(*test_keys)["input"].is_null()) {
+            entry["input"] = (*test_keys)["input"];
+            test_keys->erase("input");
+        } else {
+            entry["input"] = next_input;
+        }
```
bassosimone: Can you refactor the code such that the key inside of `*test_keys` is called `input_` rather than input to indicate that we're passing up information to be consumed at this layer? [This is something we already do in another case](https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/nettests/runnable.cpp#L164), so I'd rather see that for consistency.

Also, can you add `XXX` in front of the comment, to indicate that we can probably think at better solutions (but maybe not now, since the need just arose, and we'd better wait a little bit to figure out the best way, once more use cases emerge)?

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,88 @@
...
+    if (options["expected_body"].empty()) {
+        expected_body = constants::MEEK_SERVER_RESPONSE;
+    } else {
+        expected_body = options["expected_body"];
+    }
+
+    std::list<std::string> outer_inner = split(input, ":");
+    if (outer_inner.size() != 2) {
+        std::string fail_msg = "Couldn't split input: " + input;
+        logger->warn(fail_msg.c_str());
```
bassosimone: This should read as

```C++
logger->warn("Couldn't split input: %s", input.c_str());
```

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,88 @@
...
+        (*entry)["failure"] = fail_msg;
```
bassosimone: I think we should use existing errors for failures. Here it would probably suffice to use:

```C++
(*entry)["failure"] = ValueError().as_ooni_error();
```

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,88 @@
...
+        callback(entry);
+        return;
+    }
+    // XXX: We should make sure that we remove leading and trailing whitespaces
```
bassosimone: Perhaps we can create an issue for that?

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,88 @@
...
+    outer_host.insert(0, "https://");
+    inner_host.insert(0, "https://");
+
+    ErrorOr<http::Url> outer_url = mk::http::parse_url_noexcept(outer_host);
+    ErrorOr<http::Url> inner_url = mk::http::parse_url_noexcept(inner_host);
+
+    if (!outer_url || !inner_url) {
+        std::string fail_msg = "Invalid url: " + outer_host + " or "
+                                               + inner_host;
+        logger->warn(fail_msg.c_str());
```
bassosimone: Same as above, here there should be a proper format string.

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,88 @@
...
+        (*entry)["failure"] = fail_msg;
```
bassosimone: Regarding the error returned here, you can either use `ValueError` as above, or, as another option, you can perhaps extract the error from `inner_url` or `outer_url` using e.g. `inner_url.as_error()`.

src/measurement_kit/cmd/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,53 @@
...
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
+
+#include "../cmdline.hpp"
+
+namespace meek_fronted_requests {
+
+#define USAGE                                                  \
+    "usage: measurement_kit [options] meek_fronted_requests\n" \
+    "                       [-B expected_body] input_file\n"
```
bassosimone: For consistency with existing ooni-probe command line, there should be a `-f input_file` option rather than the possibility of receiving the input file after the options as an argument.

src/measurement_kit/cmd/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+int main(std::list<Callback<BaseTest &>> &initializers, int argc, char **argv) {
+    Settings settings;
+    std::string name = argv[0];
```
bassosimone: These two variables do not seem to be used

src/measurement_kit/cmd/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,53 @@
...
+    uint32_t verbosity = 0;
+    std::string expected_body;
```
bassosimone: Also this variable does not seem to be used

src/measurement_kit/cmd/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,53 @@
...
+            break;
+        case 'n':
+            test.set_options("no_collector", true);
+            break;
+        case 'v':
+            ++verbosity;
+            break;
+        default:
+            fprintf(stderr, "%s\n", USAGE);
+            exit(1);
```
bassosimone: You are parsing more options than required from the usage. Options `-n` and `-v` are already parsed by [the common code stage that invokes this function](https://github.com/measurement-kit/measurement-kit/blob/master/src/measurement_kit/main.cpp#L43), so you don't need to do that here.

src/measurement_kit/cmd/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    if (argc < 1) {
+        fprintf(stderr, "%s\n", USAGE);
+        exit(1);
+        /* NOTREACHED */
+    }
+    while (argc > 0) {
+        test.add_input_filepath(argv[0]);
+        argc -= 1, argv += 1;
+    }
```
bassosimone: As a more general comment to this main, I suggest to make it more similar to [the main of web_connectivity](https://github.com/measurement-kit/measurement-kit/blob/master/src/measurement_kit/cmd/web_connectivity.cpp), which currently is probably, in terms of functionality, a superset of what we need to do here.

test/fixtures/meek_fronted_requests.txt
```diff
@@ -0,0 +1 @@
+a0.awsstatic.com:d2zfqthxsdq309.cloudfront.net
```
bassosimone: If possible, I'd like to have two-three entries here (in general, things supposed to run more than once are more thoroughly tested if run more than once; when running once, again in general, I always have the doubt because you never know whether the state is properly reset after the first run)

test/nettests/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,27 @@
...
+using namespace mk;
+
+TEST_CASE("Synchronous meek-fronted-requests test") {
+    test::nettests::make_test<MeekFrontedRequestsTest>("meek_fronted_requests.txt")
+        .run();
+}
+
+TEST_CASE("Asynchronous meek-fronted-requests test") {
+    auto t = test::nettests::make_test<MeekFrontedRequestsTest>("meek_fronted_requests.txt");
+    t.runnable->needs_input = true;
```
bassosimone: I don't think overriding `needs_input` is needed here

bassosimone:
Other things that you should be doing as part of this PR include:

- [ ] mention the test in README.md
- [ ] add yourself to AUTHORS
- [ ] create basic docs in `doc/api/nettests/`
- [ ] update `docs/api/nettests.md` to refer to new doc

joelanders:
Ok, think I got all those.
But `doc/api/nettests.md` doesn't have an index.

bassosimone:
> But doc/api/nettests.md doesn't have an index.

Can you add one listing the files we already have inside `doc/api/nettests`?

Also, it seems the build fails,  but for a stupid reason related to coveralls. I'll try to fix it such that merging master into this pull request will allow us to move forward.

test/fixtures/meek_fronted_requests.txt
bassosimone: @joelanders do you know of another one-two lines we can put here?

src/libmeasurement_kit/ooni/meek_fronted_requests.cpp
```diff
@@ -0,0 +1,86 @@
...
+        callback(entry);
+        return;
+    }
+    // XXX: We should make sure that we remove leading and trailing whitespaces
+    outer_host = outer_inner.front();
+    inner_host = outer_inner.back();
+
+    // url parsing methods require a schema
+    outer_host.insert(0, "https://");
+    inner_host.insert(0, "https://");
```
bassosimone: Neat way to prepend to a string!

test/fixtures/meek_fronted_requests.txt
bassosimone: Please, disregard the above comment.

joelanders:
Ok, added the index. Also, I don't know any other meek-testing backends :/

PR #1142, opened 2017-02-14
mock.hpp: reduce implicit MK_MOCK_NAMESPACE magic

bassosimone:
@joelanders, do you think that this diff increases readability/understandability?

### DIFF
--- a/doc/api/common/mock.md
+++ b/doc/api/common/mock.md
@@ -10,11 +10,7 @@ MeasurementKit (libmeasurement_kit, -lmeasurement_kit).
 #include <measurement_kit/common.hpp>
 
 #define MK_MOCK(name_) decltype(name_) name_ = name_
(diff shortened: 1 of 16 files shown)


bassosimone:
Ah, I figured it out (`stupid /me`). Rather than playing tricks with collating tokens, it suffices to implement `MK_MOCK_AS(function, renamed_function)` so that, when one sees something like `net_connect`, it quickly notices that it's mentioned in the macro (reading something like `MK_MOCK_AS(net::connect, net_connect)` and so there is no magic, only syntactic sugar in that one does not need to write `decltype(net::connect) net_connect = net::connect`.

I'll tweak the diff when I have time, and force push on top of this PR.

bassosimone:
@AntonioLangiu, can you take care of this?

PR #1146, opened 2017-02-16
HTTP header fields manipulation test

joelanders:
This needs some style work, but also there's some HTTP parser error I don't understand.


hellais:
> also there's some HTTP parser error I don't understand

What is this error you are seeing exactly?

hellais:
@bassosimone why is travis not running?

src/libmeasurement_kit/ooni/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,90 @@
...
+static std::string randomly_capitalize(std::string input) {
+    std::random_device rd;
+    std::mt19937 g(rd());
+    for(auto &c : input) {
+        if(g()%2 == 0) {
+            c = toupper(c);
+        }
+    }
+    return input;
+}
```
hellais: I think these two functions should be moved into utils.

src/libmeasurement_kit/ooni/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+void http_header_field_manipulation(std::string input, Settings options,
+                   Callback<Var<report::Entry>> callback,
+                   Var<Reactor> reactor, Var<Logger> logger) {
+    Var<Entry> entry(new Entry);
+
+    options["http/url"] = "http://127.0.0.1:2000";
```
hellais: This needs to be obtained from `options["backend"]` and probably needs to have a deafult value set to something inside of constants (similarly to web_connectivity: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/web_connectivity.cpp#L378)

src/libmeasurement_kit/ooni/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,90 @@
...
+                            };
+
+    templates::http_request(entry, options, headers, body,
+                            [=](Error err, Var<http::Response> response) {
+                                if (err) {
+                                    logger->debug(
+                                        "http_header_field_manipulation: http-request error: %s",
+                                        err.explain().c_str());
+
+                                    (*entry)["http_header_field_manipulation_error"] =
```
hellais: This is not a valid key according to the test spec.

You are also missing some more keys that should be in here according to the spec, such as `tampering`.

src/libmeasurement_kit/ooni/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,90 @@
...
+    "Mozilla/5.0 (iPhone; U; CPU iPhone OS 3 1 2 like Mac OS X; en-us) AppleWebKit/528.18 (KHTML, like Gecko) Mobile/7D11",
+    "Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; de; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 6.1; de; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; de; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.1.7) Gecko/20091221 Firefox/3.5.7",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; de; rv:1.9.1.7) Gecko/20091221 Firefox/3.5.7 (.NET CLR 3.5.30729))",
+};
```
hellais: This should probably go inside of `constants.hpp`

src/measurement_kit/cmdline/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,54 @@
...
+            break;
+        default:
+            std::cout << kv_usage;
+            exit(1);
+        }
+    }
+    argc -= mkp_optind;
+    argv += mkp_optind;
+    while (argc > 0) {
+        test.add_input_filepath(argv[0]);
```
hellais: Why are you passing it an input_filepath?

test/nettests/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <measurement_kit/nettests.hpp>
+
+#include <chrono>
+#include <iostream>
+#include <thread>
+
+using namespace mk;
+
+TEST_CASE("Synchronous http-header-field-manipulation test") {
+    nettests::HeaderFieldManipulationTest{}
```
hellais: Isn't it called `HTTPHeaderFieldManipulationTest`?

test/nettests/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,43 @@
...
+//        .set_options("geoip_country_path", "GeoIP.dat")
+//        .set_options("geoip_asn_path", "GeoIPASNum.dat")
+//        .set_options("nameserver", "8.8.8.8")
+//        .set_input_filepath("test/fixtures/urls.txt")
+        .run();
+}
+
+TEST_CASE("Asynchronous http-header-field-manipulation test") {
+    bool done = false;
+    nettests::HeaderFieldManipulationTest{}
```
hellais: Same as above.

bassosimone:
> @bassosimone why is travis not running?

It's disable for pull requests opened from forks. We can either enable it, or give @joelanders permission to push branches, or &mdash; manually and painfully &mdash; push a branch with a different name and same commits identifiers so to trigger travis.

hellais:
> It's disable for pull requests opened from forks

Is there a particular reason to do so? Why would we not want CI testing on PRs from forks?

bassosimone:
> Is there a particular reason to do so? Why would we not want CI testing on PRs from forks?

When I modified the settings it looked like a good idea, but now it seems to me it's not really scalable and perhaps it would be better to enable them again.

src/libmeasurement_kit/http/request.cpp
```diff
@@ -65,13 +65,17 @@ void Request::serialize(net::Buffer &buff, Var<Logger> logger) {
...
+    if (headers.find("host") == headers.end()) {
+        buff << "Host: " << url.address;
+        if ((url.schema == "http" and url.port != 80) or
+            (url.schema == "https" and url.port != 443)) {
+            buff << ":";
+            buff << std::to_string(url.port);
+        }
+        buff << "\r\n";
     }
-    buff << "\r\n";
```
bassosimone: @joelanders, I think you should not move the final "\r\n" inside the if

bassosimone: ah, no, disregard: I was confused when reading the diff but reading the whole file clarified to me that this part of the diff is indeed correct :-)

bassosimone:
For the records, @hellais :
>> also there's some HTTP parser error I don't understand
>What is this error you are seeing exactly?

I discussed this offline with @joelanders. It turns out that the helper is an echo server, so what returns back is a valid HTTP request. However, we currently only configure our HTTP parser to parse HTTP responses. Therefore, the parser failed because of a invalid token when trying to parse `HTTP/1.1` (or something alike) and found instead a method. For this reason, we should find another strategy to deal with this test. I see that ooni-probe uses an HTTP template for this, but I fail to understand how this is possible (I expected the HTTP template to be able to deal with responses and not requests). For this reason I am now wondering whether what I remember (we noticed that the backend is an echo backend) was correct or I am misled.

hellais:
> I discussed this offline with @joelanders. It turns out that the helper is an echo server, so what returns back is a valid HTTP request. 

The helper for this test, is not a echo server, but the [HTTP return JSON headers helper](https://github.com/TheTorProject/ooni-spec/blob/master/test-helpers/th-002-http-return-json-headers.md) as specified in [the spec](https://github.com/TheTorProject/ooni-spec/blob/master/test-specs/ts-006-header-field-manipulation.md#test-preconditions)

bassosimone:
> Edit: looking at the code I see that you are using http://38.107.216.10:80 that is in fact the correct test-helper, so I am not sure why you are seeing it behave as an echo server 😕

We'll insist in trying to understand what was going on. To be honest, I should have written about all of this earlier, when I did my memory of what we noticed was not so current.

joelanders:
Bit of explanation on the "tampering" report key:
https://github.com/TheTorProject/ooni-probe/issues/735

(Also, the parsing error mentioned directly above was fixed. (my understanding of the backend helper was wrong). And we *are* using the HTTP template.)

src/libmeasurement_kit/common/utils.cpp
```diff
@@ -150,4 +151,23 @@ ErrorOr<std::string> slurp(std::string path) {
     return s;
 }
 
+std::string random_choice(std::vector<std::string> inputs) {
+    std::random_device rd;
+    std::mt19937 g(rd());
+    std::shuffle(inputs.begin(), inputs.end(), g);
+    return inputs[0];
+}
```
bassosimone: Can you write regress tests for this function?

src/libmeasurement_kit/common/utils.cpp
```diff
@@ -150,4 +151,23 @@ ErrorOr<std::string> slurp(std::string path) {
...
+std::string randomly_capitalize(std::string input) {
+    std::random_device rd;
+    std::mt19937 g(rd());
+    for(auto &c : input) {
+        if(g()%2 == 0) {
+            c = toupper(c);
+        }
+    }
+    return input;
+}
```
bassosimone: Can you write regress tests for this function?

src/libmeasurement_kit/ooni/constants.hpp
```diff
@@ -56,6 +56,19 @@ const http::Headers COMMON_CLIENT_HEADERS = {
...
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.1.7) Gecko/20091221 Firefox/3.5.7",
+    "Mozilla/5.0 (iPhone; U; CPU iPhone OS 3 1 2 like Mac OS X; en-us) AppleWebKit/528.18 (KHTML, like Gecko) Mobile/7D11",
+    "Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; de; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 6.1; de; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; de; rv:1.9.2) Gecko/20100115 Firefox/3.6",
+    "Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.1.7) Gecko/20091221 Firefox/3.5.7",
+    "Mozilla/5.0 (Windows; U; Windows NT 5.1; de; rv:1.9.1.7) Gecko/20091221 Firefox/3.5.7 (.NET CLR 3.5.30729))",
```
bassosimone: Those strikes me as very old user agents. I suggest we create an issue regarding updating them. My understanding is that they are same as ooni-probe ones. I am wondering whether it would make sense to create a resource file containing the list of user agents from which to extract one at random.

Cc: @hellais

src/measurement_kit/cmd/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,32 @@
...
+        case 'b':
+            backend = optarg;
+            break;
+        default:
+            fprintf(stderr, "%s\n", USAGE);
+            exit(1);
+        }
+    }
+    argc -= optind;
+    argv += optind;
```
bassosimone: After having updated `argc` and `argv`, for correctness here you should check whether `argc != 0` and exit with usage error in such case (this means the user passed in one or more strings after options).

test/ooni/utils.cpp
```diff
@@ -263,3 +263,83 @@ TEST_CASE("represent_string works") {
...
+    }
+    SECTION("For matching headers") {
+        response->body = "{\"headers_dict\": {\"foo\": \"bar\"}}";
+        ooni::compare_headers_response({ {"foo", "bar"} },
+                                       response,
+                                       entry,
+                                       Logger::global());
+        REQUIRE((*entry)["tampering"]["header_name_diff"] == std::set<std::string>{});
+        REQUIRE((*entry)["tampering"]["header_field_name"] == false);
+    }
```
bassosimone: !

src/libmeasurement_kit/ooni/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+                                    (*entry)["failure"] =
+                                        err.as_ooni_error();
+                                }
+
+                                if (!response) {
+                                    logger->warn("null response");
+                                } else {
+                                    compare_headers_response(headers, response,
+                                                             entry, logger);
```
bassosimone: Perhaps here it would make sense, for robustness, to wrap `compare_headers_response()` in a `try...except` where we basically catch for `(...)` and suppress exceptions but we set `failure` in such case.

Saying that because said function calls code that handles `nlohmann::json` and other stdlib stuff that *may* throw exceptions in case of strange/unexpected input, and I think it makes sense to stop exceptions propagating from there at this point and record an error in the entry.

What do you think?

test/ooni/utils.cpp
bassosimone: (The semantic of `!` in this context is "I like")

bassosimone:
Other things that you should be doing as part of this PR include:

- [ ] mention the test in README.md
- [ ] create basic docs in `doc/api/nettests/`
- [ ] update `docs/api/nettests.md` to refer to new doc

src/libmeasurement_kit/ooni/constants.hpp
joelanders: https://github.com/TheTorProject/ooni-spec/issues/83 <- updated to mention oldness.

bassosimone:
> I guess I can do the doc/api/nettests.md index after the meek branch is upstream of this one to avoid a silly merge conflict.

Sure, seems legit!

joelanders:
should be g2g!

test/nettests/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,29 @@
...
+TEST_CASE("Synchronous http-header-field-manipulation test") {
+    test::nettests::make_test<HttpHeaderFieldManipulationTest>()
+        .set_options("backend", "http://38.107.216.10:80")
+        .run();
+}
+
+TEST_CASE("Asynchronous http-header-field-manipulation test") {
+    test::nettests::run_async(
+        test::nettests::make_test<HttpHeaderFieldManipulationTest>()
+            .set_options("backend", "http://38.107.216.10:80")
```
hellais: It's better if we put a hostname here to be more future proof.

@anadahz what hostname should we use for http-header-field-manipulation test?

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -162,5 +164,58 @@ std::string scrub(std::string s, std::string real_probe_ip) {
...
+
+    std::set_difference(req_keys.begin(), req_keys.end(),
+                        resp_keys.begin(), resp_keys.end(),
+                        std::inserter(diff, diff.begin()));
+    std::set_difference(resp_keys.begin(), resp_keys.end(),
+                        req_keys.begin(), req_keys.end(),
+                        std::inserter(diff, diff.begin()));
+    (*entry)["tampering"]["header_name_diff"] = diff;
+    (*entry)["tampering"]["header_field_name"] = !diff.empty();
+}
```
hellais: Generally we have avoided to put test specific semantics inside of utils. I would suggest this function is refactored to have the semantics for the hhfm test be implemented in the test itself.

That is to say this function should either be put in the hhfm test or it should return some values, but not actually write the entry.

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -162,5 +164,58 @@ std::string scrub(std::string s, std::string real_probe_ip) {
...
+    }
+
+    nlohmann::json resp;
+    try {
+        resp = nlohmann::json::parse(response->body);
+    } catch (const std::invalid_argument &) {
+        logger->warn("response body not valid JSON");
+        (*entry)["tampering"]["total"] = true;
+        (*entry)["tampering"]["request_line_capitalization"] = true;
+        return;
```
hellais: Be careful to ensure that the keys `header_field_name` and `header_field_diff` are written also if you return early.

src/libmeasurement_kit/ooni/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,74 @@
...
+                                    }
+                                }
+
+                                callback(entry);
+                            },
+                            reactor, logger);
+}
+
+} // namespace ooni
+} // namespace mk
```
hellais: <bikeshed>
The formatting of this diff looks a bit weird. Was it formatted with `clang-format`?
</bikeshed>

src/libmeasurement_kit/ooni/http_header_field_manipulation.cpp
```diff
@@ -0,0 +1,74 @@
...
+                            };
+
+    templates::http_request(entry, options, headers, body,
+                            [=](Error err, Var<http::Response> response) {
+                                if (err) {
+                                    logger->debug(
+                                        "http_header_field_manipulation: http-request error: %s",
+                                        err.explain().c_str());
+
+                                    (*entry)["failure"] =
```
hellais: The failure key should always be present and initialised to `null`.

src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -162,5 +164,58 @@ std::string scrub(std::string s, std::string real_probe_ip) {
...
 }
 
+void compare_headers_response(http::Headers headers,
+                             Var<http::Response> response, Var<report::Entry> entry,
+                             Var<Logger> logger) {
+    if (response->body.empty()) {
+        logger->warn("empty response body");
+        (*entry)["tampering"]["total"] = true;
+        (*entry)["tampering"]["request_line_capitalization"] = true;
+        return;
```
hellais: Same comment as below.

test/nettests/http_header_field_manipulation.cpp
anadahz: > what hostname should we use for http-header-field-manipulation test?

`a.http.th.ooni.io`

src/libmeasurement_kit/ooni/utils.cpp
joelanders: I'm not too good at this C++ namespacing/includes stuff. If I put that function in the `src/libmeasurement_kit/ooni/http_header_field_manipulation.cpp` file, I have to do some more work to make it available to the unit tests. The unit tests include the network tests through `measurement_kit/nettests.hpp`, but we don't want to put this internal function on that interface...

src/libmeasurement_kit/ooni/utils.cpp
joelanders: Ok, I was copying the ooni-probe behavior, but I will change it.

src/libmeasurement_kit/ooni/utils.cpp
joelanders: @bassosimone ^ ideas?

bassosimone: You can create `src/libmeasurement_kit/ooni/http_header_field_manipulation.hpp` and put there the prototype for the function. Then you can move the implementation in `src/..../http_header_field_manipulation.cpp`. Finally, you can include `.../http_header_field_manipulation.hpp` in the unit tests where needed.

joelanders: Works a treat, thanks.

test/common/utils.cpp
```diff
@@ -346,3 +346,18 @@ TEST_CASE("mk::endswith() works as expected") {
...
+    REQUIRE(std::find(choices.begin(), choices.end(), choice) != choices.end());
+}
+
+TEST_CASE("randomly_capitalize isn't obviously wrong") {
+    auto lower_string = "abcdefghij";
+    auto upper_string = "ABCDEFGHIJ";
+    auto rc_lower_string = mk::randomly_capitalize(lower_string);
+    REQUIRE(rc_lower_string != lower_string);
+    auto rc_upper_string = mk::randomly_capitalize(upper_string);
+    REQUIRE(rc_upper_string != upper_string);
```
bassosimone: Note to self: this test will probably fail. I guess we should make randomly_capitalize also uncapitalize some letters at random (i.e. it should not assume input is lowercase).

test/ooni/utils.cpp
```diff
@@ -263,3 +263,83 @@ TEST_CASE("represent_string works") {
                  .dump()));
     }
 }
+
```
bassosimone: Note to self: this hunk should be moved into the above test file.

joelanders:
:octopus: :smile:

bassosimone:
^^^ interesting

joelanders:
ironically i'm still working on something to do with this PURGE pseudo-method that vodafone DE blocks.
```
l@tp ~/code/trafficserver [tags/7.1.6] $ curl -I -X PURGE example.com
HTTP/1.1 403 Forbidden
Server: WebProxy/1.0 Pre-Alpha
Date: Tue, 30 Jul 2019 14:27:28 GMT
Content-Length: 0
Connection: keep-alive
```

bassosimone:
I feel like I should do something about this

bassosimone:
Oh, hello!

```
$ curl -v -I -X PURGE example.com
* Rebuilt URL to: example.com/
*   Trying 93.184.216.34...
* TCP_NODELAY set
* Connected to example.com (93.184.216.34) port 80 (#0)
> PURGE / HTTP/1.1
> Host: example.com
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 403 Forbidden
HTTP/1.1 403 Forbidden
< Server: WebProxy/1.0 Pre-Alpha
Server: WebProxy/1.0 Pre-Alpha
< Date: Tue, 30 Jul 2019 14:30:34 GMT
Date: Tue, 30 Jul 2019 14:30:34 GMT
< Content-Length: 0
Content-Length: 0
< Connection: keep-alive
Connection: keep-alive

< 
* Connection #0 to host example.com left intact
```

This is a low-cost Vodafone company in Italy called Ho Mobile.

joelanders:
The server name doesn't inspire much confidence :)

PR #1157, opened 2017-03-02
replace short PGP key fingerprint with long one

joelanders:


bassosimone:
Thanks!

joelanders:
ah, whoops, I tested a bit with gpg (not gpg2)... x_x

PR #1162, opened 2017-03-05
Fix "unknown_failure 3" bug

bassosimone:
1. it turns out the new code to map to errno returned ValueError when
   errno was equal to zero, which happens in the connect() path in case
   of SSL error, however the code receiving the error code expected to
   receive NetworkError

2. we were using TLSv1 as method, under the expectation that it covers
   all above TLSv1, which is wrong; SSLv23 should be used instead, coupled
   with disabling SSLv2 and SSLv3 that are deprecated (actually SSLv2 is
   disabled by default but I wanted to be very explicit in new code)

### DIFF
--- a/build/valgrind/openssl.supp
+++ b/build/valgrind/openssl.supp
@@ -2,7 +2,6 @@
    <probably-false-positive-warning-on-travis>
    Memcheck:Addr4
    fun:tls_check_name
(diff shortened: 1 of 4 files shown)


bassosimone:
(merge details discussed with @hellais)

PR #1165, opened 2017-03-09
Release v0.4.2

bassosimone:


hellais:
Does this include the fix for error `3009` as well?

bassosimone:
> Does this include the fix for error 3009 as well?

Not yet, but I think it makes sense to add it.

PR #1170, opened 2017-03-11
ooni/template.cpp: don't assume !!response->request

### LINKED ISSUE #1169: http: response->request MAY be nullptr
# Acknowledging the existence of the bug

Recently, a couple of travis-ci builds failed. See for example [build #4766](https://travis-ci.org/measurement-kit/measurement-kit/builds/209648498). The failure cause is currently unknown but I'm sure here we're not dealing with a transient network error. In fact, it seems that the `test/nettests/web_connectivity` test fails due to:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  null pointer
```

I will now further investigate this issue and update the description as I have more info. Note that it is not clear to me, as of know, whether this error could impact `stable` or not.

bassosimone:
Closes #1169.

### DIFF
--- a/doc/api/http/request.md
+++ b/doc/api/http/request.md
@@ -181,11 +181,11 @@ http::request({
 
 # BUGS
 
-- It is not possible to search HTTP headers in a case insensitive fashion
+- The `http/ignore_body` setting is not implemented.
 
-- The `Response::response_line` field is always empty
-
-- The `http/ignore_body` setting is not implemented
+- The `Var<Response>` returned by the various callbacks MAY be pointing
+  to `nullptr` and, moreover, there MAY be cases where `Var<Response> response`
+  is pointing to a valid response but `response->request` is `nullptr`.
 
 # HISTORY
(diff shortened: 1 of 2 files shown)


doc/api/http/request.md
```diff
@@ -181,11 +181,11 @@ http::request({
...
 
-- It is not possible to search HTTP headers in a case insensitive fashion
+- The `http/ignore_body` setting is not implemented.
 
-- The `Response::response_line` field is always empty
-
-- The `http/ignore_body` setting is not implemented
+- The `Var<Response>` returned by the various callbacks MAY be pointing
+  to `nullptr` and, moreover, there MAY be cases where `Var<Response> response`
+  is pointing to a valid response but `response->request` is `nullptr`.
```
bassosimone: Ah, yes, while there, I also updated the documentation removing a couple of sentences that are not accurate anymore and making sure we document the behaviour observed in #1169.

PR #1177, opened 2017-03-14
Fix report/entry.hpp w/ new NDK, nlohmann::json

bassosimone:
This diff, which costed me more fatigue than it seems, stops using
the constructor of `nlohmann::json` for implementing `report::Entry`,
because of the following reasons:

1. the latest version of the NDK fails to build `report::Entry`
   with this interestingly not-so-clear error:

> In file included from .../include/measurement_kit/report/report.hpp:8:
> .../include/measurement_kit/report/entry.hpp:16:27: error: recursive
>     evaluation of default argument
>   using nlohmann::json::json;

2. upgrading to the latest version (2.1.1) of nlohmann::json breaks
   `report::Entry` build in several places

Instead, with this diff applied (and with further small tweaks for 2.1.1),
I have seen both NDK and with-v2.1.1 builds terminating, including at
least a build using Valgrind:

>   https://travis-ci.org/measurement-kit/measurement-kit/builds/210762213

It's still unclear whether we should crank minor after this diff is
applied to `stable` or whether we can consider this a non-API-changing
change considering that code using `Entry` does not require changes
once this diff has been applied. I'll think at this later.

### DIFF
--- a/include/measurement_kit/report/entry.hpp
+++ b/include/measurement_kit/report/entry.hpp
@@ -5,7 +5,7 @@
 #define MEASUREMENT_KIT_REPORT_ENTRY_HPP
 
 #include <measurement_kit/common.hpp>
-#include <measurement_kit/ext.hpp>
+#include <measurement_kit/ext/json.hpp> // vendored nlohmann::json
 
 namespace mk {
 namespace report {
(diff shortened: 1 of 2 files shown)


hellais:
The diff looks good to me. It seems like you are just "enriching" the constructors and methods to support more types and most is remaining unchanged.

I don't think this should be considered as an API change as the user should still be able to use it as they used to and nothing should change from that perspective.

For me this is good to merge.

bassosimone:
> I don't think this should be considered as an API change as the user should still be able to use it as they used to and nothing should change from that perspective.

Agreed!

> For me this is good to merge.

Thank you, will move forward with blessing 0.4.3 then!

🐳

PR #1185, opened 2017-03-28
Build and archive dependencies separately

bassosimone:
Probably supersedes #1017


build/android/cross
```diff
@@ -72,7 +72,7 @@ elif [ "$ARCH" = "armeabi-v7a" ]; then
     CXXFLAGS="$CFLAGS"
 elif [ "$ARCH" = "mips" ]; then
     # Note: mips is actually mipsel (i.e. little endian)
-    TOOLCHAIN_NAME=mipsel-linux-android
+    TOOLCHAIN_NAME=mips64el-linux-android
```
bassosimone: This is an upgrade in the toolchain by which we always need to use `mips64el`. I have tested this by running the resulting example app for Android on a `mips` emulator with API 16.

build/android/cross
```diff
@@ -125,10 +125,10 @@ export LDFLAGS="${LDFLAGS} -L${SYSROOT}/usr/lib${LIB_SUFFIX} -L${ANDROID_TOOLCHA
 export pkg_configure_flags="--host=${TOOLCHAIN_NAME} --disable-shared $CONFIG_EXTRA"
 
 # Where to install headers and libraries
-export pkg_prefix="${ROOTDIR}/jni"
+export pkg_prefix="${ROOTDIR}"
 
 # Tell ./build/dependency that we are cross compiling
-export pkg_cross="android"
+export pkg_cross="jni"
```
bassosimone: Names here changed to make sure we output in `${ROOTDIR}/jni`. Semantic of `pkg_cross` changed from generically meaning "we're cross compiling for platform" to mean exactly the type of platform ("jni" for Android and one of "iphoneos" and "iphonesimulator" for iOS).

build/fetch-dep
```diff
@@ -0,0 +1,45 @@
...
+fi
+set -e
+pkg_rootdir=$(cd $(dirname $(dirname $0)) && pwd -P)
+cd $pkg_rootdir
+if [ "$pkg_prefix" = "" ]; then
+    pkg_prefix="$pkg_rootdir/builtin"
+fi
+npm install download-github-release
+./node_modules/.bin/download-github-release $arguments                     \
+        measurement-kit dependencies
```
bassosimone: This is a fine `node` module that helps to download artefacts. Since the build process continues to download archives as before using `wget`, I'm not that concerned because this `node` utility is only used when one is developing to cross compile the next dependencies, so no more dependencies are needed in the use case where one wants just to build MK for Android or iOS.

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -1,31 +1,8 @@
...
-index 0000000..4c2880c
---- /dev/null
-+++ b/.travis.yml
-@@ -0,0 +1,6 @@
-+language: c
-+install: autoreconf -i
-+script: ./configure && make -j2 V=0 && make -j2 check V=0
-+compiler:
-+  - clang
-+  - gcc
```
bassosimone: This removes a hunk that was not actually needed.

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -36,68 +13,62 @@ index ad815d9..babd653 100644
...
  endif
- 
--SUBDIRS = . include sample test
-+SUBDIRS = . include
+@@ -143,8 +138,6 @@ CLEANFILES=
+ DISTCLEANFILES=
+ BUILT_SOURCES =
+ include include/include.am
+-include sample/include.am
+-include test/include.am
```
bassosimone: Update this hunk to libevent-2.1 style

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -36,68 +13,62 @@ index ad815d9..babd653 100644
...
+ 
+@@ -141,10 +137,6 @@ dnl Checks for libraries.
  AC_SEARCH_LIBS([inet_ntoa], [nsl])
  AC_SEARCH_LIBS([socket], [socket])
  AC_SEARCH_LIBS([inet_aton], [resolv])
--AC_SEARCH_LIBS([clock_gettime], [rt])
+-if test "x$enable_clock_gettime" = "xyes"; then
+-  AC_SEARCH_LIBS([clock_gettime], [rt])
+-  AC_CHECK_FUNCS([clock_gettime])
+-fi
```
bassosimone: Basically, we know we *must not* use clock_gettime() because it's not available on old systems.

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -36,68 +13,62 @@ index ad815d9..babd653 100644
...
+ 
+ dnl - check if the macro _WIN32 is defined on this compiler.
+@@ -342,6 +334,7 @@ AC_CHECK_FUNCS([ \
+   accept4 \
+   arc4random \
+   arc4random_buf \
++  arc4random_addrandom \
+   eventfd \
+   epoll_create1 \
+   fcntl \
```
bassosimone: Make sure we check for `arc4random_addrandom` because we need to know that later.

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -36,68 +13,62 @@ index ad815d9..babd653 100644
...
- 		void (*tptr)(void *,size_t) =
- 		    (void (*)(void*,size_t))arc4random_buf;
- 		if (tptr != NULL) {
--			return arc4random_buf(buf, n);
-+			arc4random_buf(buf, n);
-+			return;
- 		}
- 	}
- #endif
-@@ -163,6 +165,16 @@ ev_arc4random_buf(void *buf, size_t n)
```
bassosimone: This hunk here is not needed anymore because it has been integrated into libevent-2.1

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -123,4 +94,26 @@ index 284341c..5b8b0a6 100644
...
++#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined LIBRESSL_VERSION_NUMBER
+ 
+ static inline BIO_METHOD *BIO_meth_new(int type, const char *name)
+ {
+@@ -30,6 +30,6 @@ static inline BIO_METHOD *BIO_meth_new(int type, const char *name)
+ 
+ #define TLS_method SSLv23_method
+ 
+-#endif /* OPENSSL_VERSION_NUMBER < 0x10100000L */
++#endif /* OPENSSL_VERSION_NUMBER < 0x10100000L || defined LIBRESSL_VERSION_NUMBER */
```
bassosimone: Fix needed to compile with libressl

build/spec/embedded-mk
```diff
@@ -12,9 +12,10 @@ fi
 cd "${ROOTDIR}"
 test -f Makefile && make clean
 test -x configure || ./autogen.sh
-./configure $pkg_configure_flags --with-libevent=${pkg_prefix}                 \
-            --with-geoip=${pkg_prefix} --with-openssl=${pkg_prefix}            \
-            --disable-examples --prefix=${pkg_prefix}                          \
+dd=${pkg_prefix}/${pkg_cross}/${pkg_cross_arch}
+./configure $pkg_configure_flags --with-libevent=${dd}                         \
+            --with-geoip=${dd} --with-openssl=${dd}                            \
+            --disable-examples --prefix=${dd}                                  \
```
bassosimone: Update location where we search for stuff (the final path is equal, but it changed the meaning of variables and now `pkg_prefix` contains less information than before, so we need to add more variables).

build/spec/geoip
```diff
@@ -1,6 +1,6 @@
 pkg_name=geoip
 pkg_repository=https://github.com/maxmind/geoip-api-c.git
-pkg_branch_or_tag=v1.6.9
+pkg_branch_or_tag=v1.6.10
```
bassosimone: Update to latest geoip

build/spec/libevent
```diff
@@ -1,6 +1,6 @@
 pkg_name=libevent
 pkg_repository=https://github.com/libevent/libevent
-pkg_branch_or_tag=patches-2.0
-pkg_tip=7dedc077aa15d592c538121c7d5f2727f682d5a6
-export CPPFLAGS="$CPPFLAGS -I$pkg_prefix/include"
-export LDFLAGS="$LDFLAGS -L$pkg_prefix/lib"
+pkg_branch_or_tag=patches-2.1
+pkg_tip=e7ff4ef2b4fc950a765008c18e74281cdb5e7668
+export CPPFLAGS="$CPPFLAGS -I$pkg_prefix/$pkg_cross/$pkg_cross_arch/include"
+export LDFLAGS="$LDFLAGS -L$pkg_prefix/$pkg_cross/$pkg_cross_arch/lib"
```
bassosimone: Update to latest libevent and update also the flags

build/valgrind/openssl.supp
```diff
@@ -29,3 +29,19 @@
...
    fun:main
 }
+{
+   <same-as-above-but-for-another-version-of-libevent>
+   Memcheck:Addr4
+   fun:tls_check_name
+   fun:_ZNSt17_Function_handlerIFvN2mk5ErrorEP11buffereventEZNS0_3net11connect_sslES3_P6ssl_stSsSt8functionIS4_ENS0_3VarINS0_7ReactorEEENSA_INS0_6LoggerEEEEUlS1_S3_E_E9_M_invokeERKSt9_Any_dataOS1_OS3_
+   fun:mk_bufferevent_on_event
+   fun:bufferevent_run_deferred_callbacks_locked
+   fun:event_process_active_single_queue
```
bassosimone: This function above is not part of libevent-2.0, so we need two suppression files.

configure.ac
```diff
@@ -42,7 +42,7 @@ if test "`uname`" != "Darwin"; then
   CFLAGS="$CFLAGS -pthread"
   CXXFLAGS="$CXXFLAGS -pthread"
   LDFLAGS="$LDFLAGS -pthread"
-else
+elif test "$cross_compiling" != yes; then
```
bassosimone: Needed, otherwise we break cross compiling for iOS

bassosimone:
@AntonioLangiu could you review this PR? You previously have reviewed a more complex PR that was similar to this one, i.e. #1017.

doc/build/dependency.md
```diff
@@ -0,0 +1,94 @@
+# Building dependencies
```
AntonioLangiu: I think we can move (and update) the file ./build/android/README.md into this documentation folder or merge it with this file. We should update it because some lines are not more valid with the latest NDK. For example, with the suggested NDK_DIR the script doesn't find the build toolchain python script.

build/fetch-dep
```diff
@@ -0,0 +1,45 @@
...
+        --)
+            shift
+            break
+            ;;
+        *)
+            echo "usage: $0 [-p] [-s regex]" 1>&2
+            exit 1
+    esac
+done
+if [ $# -ne 0 ]; then
```
AntonioLangiu: In which case can you enter into this if condition?

build/patch/pre-autogen/libevent/common/000.patch
AntonioLangiu: May you can leave this last part (from line 52 to 55). It should fail if you remove the part above (from line 41 to 43). This gives you a smaller diff. #bikeshedding

build/fetch-dep
bassosimone: When you provide non-option arguments, e.g. `./build/fetch-dep foobar`

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -36,68 +13,62 @@ index ad815d9..babd653 100644
  pkgconfigdir=$(libdir)/pkgconfig
  LIBEVENT_PKGCONFIG=libevent.pc libevent_core.pc libevent_extra.pc
  
-@@ -122,12 +118,11 @@ endif
+@@ -130,7 +126,6 @@ endif
  
  if INSTALL_LIBEVENT
  lib_LTLIBRARIES = $(LIBEVENT_LIBS_LA)
 -pkgconfig_DATA = $(LIBEVENT_PKGCONFIG)
```
bassosimone: This is to avoid installing the `PKGCONFIG`, which we don't need.

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -36,68 +13,62 @@ index ad815d9..babd653 100644
...
 +++ b/configure.ac
-@@ -112,7 +112,6 @@ dnl Checks for libraries.
+@@ -121,10 +121,6 @@ AC_ARG_ENABLE([function-sections],
+ AC_ARG_ENABLE([verbose-debug],
+ 		AS_HELP_STRING([--enable-verbose-debug, verbose debug logging]),
+ 	[], [enable_verbose_debug=no])
+-AC_ARG_ENABLE([clock-gettime],
+-     AS_HELP_STRING(--disable-clock-gettime, do not use clock_gettime even if it is available),
+-  [], [enable_clock_gettime=yes])
+-
```
bassosimone: This removes the `--disable-clock-gettime` argument.

build/patch/pre-autogen/libevent/common/000.patch
bassosimone: True, but I'd like more the scenario where I can search for `clock_gettime` in the file and cannot find it because I have removed all mentions of it.

build/patch/pre-autogen/libevent/common/000.patch
bassosimone: This diff should be equal to before, except that it's updated for libevent-2.1

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -111,10 +82,10 @@ index 284341c..5b8b0a6 100644
 + */
 +#define _EVENT_HAVE_ARC4RANDOM_ADDRANDOM 1
```
bassosimone: It *may* be possible that we should update this part of the diff to work differently. It seems now libevent does not use anymore leading underscored for variables it defines. Also, I wonder why forcing the check here when we also perform the check in the configure. I guess I'll need to double check that.

build/patch/pre-autogen/libevent/common/000.patch
bassosimone: Basically the gist of it is that libressl uses an older API and here we enable backward compat for that.

build/dependency
```diff
@@ -87,15 +87,19 @@ pkg_make_check() {
 
 pkg_make_install() {
     echo "- make $pkg_make_install_flags install"
-    based=$pkg_rootdir/third_party/tmp/$pkg_name-$pkg_branch_or_tag
-    adir=$based/$pkg_cross_arch
+    vspec=$pkg_branch_or_tag
+    if [ "$pkg_tip" != "" ]; then
+        vspec="$pkg_tip"
+    fi
```
bassosimone: We use as version indicator the tip of the branch, if available. Elsewhere in this file there is a check by which, if the branch tip is different from the expected tip, the compile script fails. This is done to ensure that we are building with exactly the branch tip that we want and not some random commit in the branch.

build/dependency
```diff
@@ -87,15 +87,19 @@ pkg_make_check() {
...
+    adir=$based/$pkg_cross/$pkg_cross_arch
     make $pkg_make_install_flags install DESTDIR=$adir
     find $adir -depth -type d -name bin -exec rm -rf {} \;
     find $adir -depth -type d -name pkgconfig -exec rm -rf {} \;
     find $adir -depth -type d -name sbin -exec rm -rf {} \;
     find $adir -depth -type f -name \*.la -exec rm -rf {} \;
-    tarname=$pkg_rootdir/$pkg_name-$pkg_branch_or_tag
-    tarname=$tarname-`echo $pkg_cross_arch|sed 's/\//-/g'`
+    tarname=$pkg_rootdir/$pkg_name-$vspec
+    tarname=$tarname-$pkg_cross-`echo $pkg_cross_arch|sed 's/\//-/g'`
```
bassosimone: Note to self: the `sed` part is probably not needed anymore.

build/ios/.gitignore
```diff
@@ -1,4 +1,6 @@
 .DS_Store
 ios-dependencies.tgz
 ios-dependencies.tgz.asc
+*.tgz
+*.tgz.asc
```
bassosimone: Note to self: the rule above becomes redundant, then. Also, it should probably make sense to add a similar rule also for Android, which is currently missing it.

build/ios/cross
```diff
@@ -43,8 +43,8 @@ export LDFLAGS="-arch ${ARCH} $MINVERSION -isysroot $(xcrun -sdk ${PLATFORM} --s
 
 export pkg_configure_flags="$EXTRA_CONFIG"
 export pkg_prefix="$DESTDIR"
-export pkg_cross="ios"
-export pkg_cross_arch="${PLATFORM}/${ARCH}"
+export pkg_cross="${PLATFORM}"
+export pkg_cross_arch="${ARCH}"
```
bassosimone: Adjust platform and arch so that we end up installing libs where we would like them to be. See above comment regarding the fact that the semantic of `pkg_cross` is now changed.

build/dependency
```diff
@@ -87,15 +87,19 @@ pkg_make_check() {
...
+    adir=$based/$pkg_cross/$pkg_cross_arch
     make $pkg_make_install_flags install DESTDIR=$adir
     find $adir -depth -type d -name bin -exec rm -rf {} \;
     find $adir -depth -type d -name pkgconfig -exec rm -rf {} \;
     find $adir -depth -type d -name sbin -exec rm -rf {} \;
     find $adir -depth -type f -name \*.la -exec rm -rf {} \;
-    tarname=$pkg_rootdir/$pkg_name-$pkg_branch_or_tag
-    tarname=$tarname-`echo $pkg_cross_arch|sed 's/\//-/g'`
+    tarname=$pkg_rootdir/$pkg_name-$vspec
+    tarname=$tarname-$pkg_cross-$pkg_cross_arch
```
bassosimone: The sed part is not needed anymore because the information that earlier was part of `pkg_cross_arch` is not split in `pkg_cross` and `pkg_cross_arch` after the semantics of `pkg_cross` changed.

build/patch/pre-autogen/libevent/common/000.patch
bassosimone: So, for sure the correct name of this define would be `EVENT__HAVE_ARC4RANDOM_ADDRANDOM`.

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -108,19 +79,49 @@ index 284341c..5b8b0a6 100644
 + * have removed arc4random_addrandom() since 1.26. Since then, other libcs
 + * followed suit (e.g. Android's own libc). But libevent's arc4random.c
 + * copy still implement arc4random_addrandom().
++ *
++ * See also:
++ *
++ * - https://github.com/measurement-kit/libevent/commit/8b275d967d7ffd95d5cc12069aef35669126c6d9
++ * - https://bugzilla.mozilla.org/show_bug.cgi?id=931354
++ * - https://bug931354.bmoattachments.org/attachment.cgi?id=829728
```
bassosimone: @AntonioLangiu I've added to the patch more documentation that explains what is going on here, after digging back into the history and understanding why the patch was issued in the first place.

build/patch/pre-autogen/libevent/common/000.patch
```diff
@@ -108,19 +79,49 @@ index 284341c..5b8b0a6 100644
...
+ #endif /* } !EVENT__HAVE_ARC4RANDOM */
  
  void
-@@ -174,7 +186,8 @@ evutil_secure_rng_get_bytes(void *buf, size_t n)
+@@ -195,8 +213,10 @@ evutil_secure_rng_get_bytes(void *buf, size_t n)
  void
  evutil_secure_rng_add_bytes(const char *buf, size_t n)
  {
-+#if defined _EVENT_HAVE_ARC4RANDOM_ADDRANDOM
++#if defined EVENT__HAVE_ARC4RANDOM_ADDRANDOM
```
bassosimone: @AntonioLangiu Here, basically, the name has been changed to follow libevent-2.1 convention where underscore names are no longer used for macro variables.

doc/build/dependency.md
bassosimone: I updated it and added a reference to this file.

> For example, with the suggested NDK_DIR the script doesn't find the build toolchain python script.

Are you sure? Are you using the latest NDK?

README.md
```diff
@@ -114,8 +114,10 @@ make check
 
 ### How to build MeasurementKit on Android
 
-To compile MeasurementKit for Android, see the README.md file contained in
-the `build/android` directory of this repository.
+We have [a specific repository](https://github.com/measurement-kit/android-libs)
+for compiling MeasurementKit for Android. You may also want to read the
+[documentation explaining how to cross compile MK dependencies for Android](
+doc/build/android.md).
```
bassosimone: @AntonioLangiu here I've updated the README file to directly point users to the `android-libs` repository unless they have other, specific needs.

doc/build/android.md
```diff
@@ -1,7 +1,8 @@
-# Build measurement-kit for Android
```
bassosimone: File moved and updated as suggested by @AntonioLangiu

bassosimone:
@AntonioLangiu: I think I have addressed your comments. I now need to rebuild and retest everything, because I made possibly breaking changes to libevent.

I will ping you when I think it's time to do another pass of reviewing this diff, thanks!

doc/build/dependency.md
AntonioLangiu: Yes, I updated the version of the NDK and all the other build tools from Android Studio. The version of NDK I have currently installed is 14.1.3816874 and is the only one (I don't have other versions installed with brew). When I try to build a dependency using the suggestions provided in the documentation I get an error because it expects the NDK_DIR and not the NDK build script.

i.e. on my system

    $ ./build/android/dependency ~/Library/Android/sdk/ndk-bundle/ndk-build libevent 

should be 

    $ ./build/android/dependency ~/Library/Android/sdk/ndk-bundle/ libevent

Can you check if the previous command works in your system?

bassosimone: Ah, you're right, that is indeed an error!

bassosimone:
Okay, so it seems that *sometimes* in the travis environment there are [Valgrind false positives that happen in LibreSSL code](https://travis-ci.org/measurement-kit/measurement-kit/jobs/219580461#L2136). I was wondering whether I did something wrong and we were not linking with the right LibreSSL anymore, however, running `ldd` on the dynamic library yields:

```
	linux-vdso.so.1 =>  (0x00007ffc5b24e000)
	libGeoIP.so.1 => /home/travis/build/measurement-kit/measurement-kit/builtin/lib/libGeoIP.so.1 (0x00007febb2971000)
	libevent_openssl-2.1.so.6 => /home/travis/build/measurement-kit/measurement-kit/builtin/lib/libevent_openssl-2.1.so.6 (0x00007febb2769000)
	libevent_pthreads-2.1.so.6 => /home/travis/build/measurement-kit/measurement-kit/builtin/lib/libevent_pthreads-2.1.so.6 (0x00007febb2566000)
	libevent-2.1.so.6 => /home/travis/build/measurement-kit/measurement-kit/builtin/lib/libevent-2.1.so.6 (0x00007febb2311000)
	libssl.so.39 => /home/travis/build/measurement-kit/measurement-kit/builtin/lib/libssl.so.39 (0x00007febb20b5000)
	libcrypto.so.38 => /home/travis/build/measurement-kit/measurement-kit/builtin/lib/libcrypto.so.38 (0x00007febb1cdc000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007febb19ba000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007febb16bd000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007febb12fc000)
	/lib64/ld-linux-x86-64.so.2 (0x000055a2930ab000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007febb10e5000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007febb0ec7000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007febb0cbf000)
```

For this reason, it seems to me consistent the hypothesis that these false positives can be safely ignored and I am going to add suppressions for them as part of this pull request.

bassosimone:
Another data point: we see the same issue when enabling valgrind on docker, where we use openssl 1.0.0 rather than libressl. Now the question is why I am failing to suppress that.

bassosimone:
I nailed it: the culprit is the test for `bacardi.com` where all the false positives are triggered. After I removed the test in 52773f5, subsequent runs of travis never failed.

What I am going to do now:

- [ ] minimise the current diff removing non-needed parts introduced to understand failures
- [ ] keep the broken test commented out and open an issue for that
- [ ] merge this pull request

bassosimone:
^^^ lol ?!

PR #1191, opened 2017-04-13
Implement OONI telegram test

joelanders:
This hits up all of the endpoints. It doesn't yet calculate the total blocked/not-blocked status (ooni-probe says its blocked if all endpoints fail). Some other final report keys are probably missing. But I think the guts are mostly there.

(It's a real surprise C++ doesn't have some kind of future-chaining stuff in `std`! Who wants asynchronous code with callback pyramids?)

```
l@tp ~/code/measurement-kit [feature/telegram] $ ./measurement_kit -vn telegram
Your public IP address: x.x.x.x           
Your country: x
Your ISP identifier: x
5%: geoip lookup
10%: open report
starting telegram test
telegram: success HTTP connecting to http://web.telegram.org/
telegram: success HTTP connecting to https://web.telegram.org/
done testing telegram web
telegram: success TCP connecting to 149.154.167.51:80
telegram: success TCP connecting to 149.154.167.51:443
telegram: success TCP connecting to 149.154.167.91:80
telegram: success TCP connecting to 149.154.167.91:443
telegram: success TCP connecting to 149.154.175.50:80
telegram: success TCP connecting to 149.154.175.100:80
telegram: success TCP connecting to 149.154.175.50:443
telegram: success TCP connecting to 149.154.175.100:443
telegram: success TCP connecting to 149.154.171.5:80
telegram: success TCP connecting to 149.154.171.5:443
done testing endpoints as TCP
telegram: success HTTP connecting to http://149.154.167.51:443
telegram: success HTTP connecting to http://149.154.167.51:80
telegram: success HTTP connecting to http://149.154.167.91:443
telegram: success HTTP connecting to http://149.154.167.91:80
telegram: success HTTP connecting to http://149.154.175.100:80
telegram: success HTTP connecting to http://149.154.175.100:443
telegram: success HTTP connecting to http://149.154.175.50:80
telegram: success HTTP connecting to http://149.154.175.50:443
telegram: success HTTP connecting to http://149.154.171.5:80
telegram: success HTTP connecting to http://149.154.171.5:443
done testing endpoints as HTTP
95%: ending the test
100%: test complete

```


src/libmeasurement_kit/ooni/telegram.cpp
```diff
@@ -0,0 +1,194 @@
...
+                result["status"]["success"] = false;
+                result["status"]["failure"] = err.as_ooni_error();
+                close_txp = false;
+            } else {
+                logger->info("telegram: success TCP connecting to %s:%d",
+                             ip.c_str(), port);
+                result["status"]["success"] = true;
+            }
+            (*entry)["tcp_connect"].push_back(result);
+            *ips_tested += 1;
```
joelanders: Is this racy? See, for example, https://code.facebook.com/posts/1661982097368498 `multiFooAsync()`. I guess it's not if we're single-threaded and there are no exceptions or IO operations in this code?

bassosimone:
> (It's a real surprise C++ doesn't have some kind of future-chaining stuff in std! Who wants asynchronous code with callback pyramids?)

There is something but it's more related to threads than to async code. We may consider implementing or lifting promises from somewhere in the future. Speaking specifically of callback pyramids, I guess it's a matter of taste, but I like it when there are 1-2 level of callbacks, because it has this nice LISP-like property that you clearly see the execution tree. When more callbacks are there (like, for example, in the driver code for NDT) I tend to err on your side and wish there is some chaining, so to inline stuff.

src/libmeasurement_kit/ooni/telegram.cpp
bassosimone: >  I guess it's not if we're single-threaded

Yes, there is a single-threaded assumption in many places in the code. Even without multiple threads we may enter into race conditions domain, if we're using Windows with IOCP where basically whenever I/O completes a callback is called (pretty much like signals). The way in which we solve this is by configuring `libevent` to serialize callbacks for us.

In the future, I would like to pay more attention to this potential race conditions, perhaps by creating some abstraction that combines `atomic` and `Var`. Another thing to consider is that, in general, I find that for our use case it would be fine to have a single I/O thread and perhaps a thread pool to hand over heavy work to. Anyway, having the `atomic + Var` thing is perhaps a matter of correctness and I guess it should be done in any case (it will basically boil down to chain an indirect access through `Var` and an atomic access).

> and there are no exceptions

In regards to exceptions, the model that I'm trying to push is the following: in most cases basic sync operations return `ErrorOr<Value>` (if we switch to C++17 this could easily become a tuple), pretty much like `golang` does, because I don't like to *implicit* implied by mixing code that may throw exceptions and code with callbacks. With @AntonioLangiu we were also writing on a more generic template piece of code that basically wraps code that can throw and, depending on usage, either returns or forwards the error that occurred. This way, just by reading the code, one can be sure that exceptions are properly forwarded along the callbacks chain.

> or IO operations in this code?

Speaking of I/O operations, the way in which we're trying to structure them is this:

- if it's a network I/O operation, it has a callback that takes an error as parameter and it should never throw an exception, unless something very bad happens

- otherwise, we generally rely on the fact that C++ streams do not throw by default unless instructed to do so, and then we check the result bits after I/O and act accordingly

src/libmeasurement_kit/ooni/telegram.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+using namespace mk::report;
+
+static void tcp_many(const std::vector<std::string> ip_ports,
+                     Var<Entry> entry,
+                     Callback<Error> cb,
+                     Var<Reactor> reactor,
+                     Var<Logger> logger) {
+    // C++20 should have continuations in std; until then,
+    // we copy the ghetto pattern from web_connectivity
```
bassosimone: I think there is already an abstraction that can be used to do that in `common`. It is called `parallel` and runs a specific algorithm over a vector of input. I think that we should:

1) check whether we can refactor that code to work in this case (and also in web_connectivity's case)

2) check which should be the expected syntax of C++20 continuations and see whether we can modify the API of `parallel` such that, when we switch to C++20, the transition is as smooth as possible.

Regarding parallel, useful links are:

a) [documentation](https://github.com/measurement-kit/measurement-kit/blob/de0c4c4abef0ebeab81691015c7c92dec7696884/doc/api/common/parallel.md)

b) [example of usage](https://github.com/measurement-kit/measurement-kit/blob/1aa5bd817dfa309b8574e919526d8f7862060f53/src/libmeasurement_kit/nettests/runnable.cpp#L304)

bassosimone: BTW, we also have an abstraction called `Continuation` that basically is a suspended function that can be resumed at a later point in time. See its [an example of usage](https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/report/file_reporter.cpp#L40); perhaps it can also be useful in this case?

src/libmeasurement_kit/ooni/telegram.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+void telegram(std::string input, Settings options,
+              Callback<Var<report::Entry>> callback,
+              Var<Reactor> reactor, Var<Logger> logger) {
+    std::vector<std::string>
+        TELEGRAM_WEB_URLS = { "http://web.telegram.org/",
+                              "https://web.telegram.org/" };
+    // should probably just make these std::pair<std::string,int>,
+    // but I'm not sure if this will be better later when I get
+    // rid of the duplication between this and the http ones
```
bassosimone: You can probably write a function that maps the list of pairs to a list of URLs.

src/libmeasurement_kit/ooni/telegram.cpp
```diff
@@ -0,0 +1,194 @@
...
+        tcp_many(TELEGRAM_TCP_ENDPOINTS, entry, [=](Error err){
+            logger->info("done testing endpoints as TCP");
+            // then try endpoints as HTTP
+            http_many(TELEGRAM_HTTP_ENDPOINTS, entry, [=](Error err){
+                logger->info("done testing endpoints as HTTP");
+                // then finish!
+                callback(entry);
+            }, reactor, logger);
+        }, reactor, logger);
+    }, reactor, logger);
```
bassosimone: Speaking of chaining stuff, one thing that I have observed is that, since in most cases we need to pass anyway reactor and logger, it's better to refactor arguments such that the callback is always the last. I have observed that this makes the code more readable than having the callback in the middle and have started to apply this pattern to new code.

Speaking specifically of your general comment regarding chaining functions, this code has the issue that you are ignoring errors, which perhaps is not what we want? I guess we want to use an abstraction that executes a list of functions in sequence (or in parallel) and produces a single error that contains as sub-errors the results of each step; does this make sense to you?

src/measurement_kit/cmd/telegram.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace telegram {
+
+#define USAGE                                       \
+    "usage: measurement_kit [options] telegram\n"   \
+
+int main(std::list<Callback<BaseTest &>> &initializers, int argc, char **argv) {
+    mk::nettests::TelegramTest test;
+    int ch;
+    while ((ch = getopt(argc, argv, "B:f:")) != -1) {
```
bassosimone: Here you should clear the options string, if no options are recognized.

src/libmeasurement_kit/ooni/telegram.cpp
joelanders: re: exceptions, check out this `Expected` thing from this guy, whose name keeps coming up in what I'm reading recently: https://en.wikipedia.org/wiki/Andrei_Alexandrescu (it sounds like what you're describing)

bassosimone: Thanks, I will look into it!

src/libmeasurement_kit/ooni/telegram.cpp
joelanders: Don't know what you mean?

src/measurement_kit/cmd/telegram.cpp
joelanders: Whoops, done.

joelanders:
Ok, fixed the getopt() string, used the new testing style, threading the options dict through, and removed the input parameter (spelled wrong in the commit message, whoops).

PR #1192, opened 2017-04-18
Implement OONI's Facebook Messenger test

joelanders:
Doesn't fill in the report, but the outline is roughly there.
If we're shopping for abstractions, the one I want to buy is some syntax like:
```
Future<double> f = 
  fooFuture(input)
  .then([](Output o) {
    return o * M_PI;
  })
  .onError([](std::exception const& e) {
    cerr << "Oh bother, " << e.what()
      << ". Returning pi instead." << endl;
    return M_PI;
  });
```
as provided by Facebook's [folly](https://code.facebook.com/posts/1661982097368498). Is that off-brand? :) Maybe boost has something like it.
```
-rw-r--r--  1 root staff 132M Apr 17 16:53 libfolly.a
```

```
l@nl2 ~/code/measurement-kit [feature/facebook_messenger] $ ./measurement_kit -vn facebook_messenger                                                                                        
Your public IP address: 81.4.110.111
Your country: NL
Your ISP identifier: AS198203
5%: geoip lookup
10%: open report
starting facebook_messenger
(1) b-api.facebook.com ipv4: 31.13.92.37
(1) b-graph.facebook.com ipv4: 31.13.93.37
(1) edge-mqtt.facebook.com ipv4: 31.13.92.5
(1) external.xx.fbcdn.net ipv4: 31.13.92.14
(1) scontent.xx.fbcdn.net ipv4: 31.13.93.7
(1) star.c10r.facebook.com ipv4: 31.13.93.3
(1) stun.fbsbx.com ipv4: 69.171.239.36
done doing dns lookups
tcp success to 31.13.93.37:443
tcp success to 31.13.92.37:443
tcp success to 31.13.92.5:443
tcp success to 31.13.93.7:443
tcp success to 31.13.93.3:443
tcp success to 31.13.92.14:443
done doing tcp tests
95%: ending the test
100%: test complete
```


bassosimone:
> as provided by Facebook's folly. Is that off-brand? :)

No, we can see to implement something like that in MK or borrow the implementation from somewhere.

> Maybe boost has something like it.

I'd rather avoid boost and only rely on standard C++. Boost is huge and I don't want to depend on the whole of it (which is super huge), plus with the advancements of C++ standard, I'd rather bump our minimum C++ version requirements than integrate boost.

joelanders:
Yep, look at that libfolly.a ^ 132MB!

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,228 @@
...
+                cb(NoError(), entry, *fb_service_ips2, reactor, logger);
+                return;
+            }
+        };
+    };
+
+    for (auto const& service_and_hostname : fb_service_hostnames) {
+        std::string service = service_and_hostname.first;
+        std::string hostname = service_and_hostname.second;
+        Settings options; // XXX ought this to forward from above?
```
bassosimone: Yes, generally we want to bring around the same `settings` object.

include/measurement_kit/ooni/nettests.hpp
```diff
@@ -36,6 +36,11 @@ void http_header_field_manipulation(std::string input, Settings,
                                     Var<Reactor> = Reactor::global(),
                                     Var<Logger> = Logger::global());
 
+void facebook_messenger(std::string input, Settings,
```
bassosimone: If I'm correct, this test does not take any input. As such, you don't need to pass input here, @joelanders.

src/libmeasurement_kit/nettests/facebook_messenger.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+FacebookMessengerTest::FacebookMessengerTest() : BaseTest() {
+    runnable.reset(new FacebookMessengerRunnable);
+    runnable->test_name = "facebook_messenger";
+    runnable->test_version = "0.0.1";
+    runnable->needs_input = false;
+}
+
+void FacebookMessengerRunnable::main(std::string input, Settings options,
+                            Callback<Var<report::Entry>> cb) {
```
bassosimone: @joelanders the fact that here you receive an `input` that you don't use is a limitation of the implementation of `Runnable`, but in fact the input is ignored, so you should not pass it downstream.

include/measurement_kit/ooni/nettests.hpp
bassosimone: @joelanders can you remove the `input` parameter from here, since the test does not receive any input?

src/libmeasurement_kit/nettests/facebook_messenger.cpp
bassosimone: Basically here you should have `/*input*/` since we don't need to pass it downstream.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+                logger->info("fb_messenger: dns error for %s, %s",
+                    service.c_str(), hostname.c_str());
+            } else {
+                for (auto answer : message->answers) {
+                    if ((answer.ipv4 != "") || (answer.hostname != "")) {
+                        //XXX
+                        //std::string asn_p = options.get("geoip_asn_path", std::string{});
+                        std::string asn_p = "GeoIPASNum.dat";
+                        GeoipDatabase geoip(asn_p);
+                        ErrorOr<std::string> asn = geoip.resolve_asn(answer.ipv4);
```
bassosimone: Here you should:

1. fetch the geoip file name from options (you should pass them as argument to the function running the test and then pass them around)

2. use the thread-local geoip cache

You can see how this is done in the implementation of web connectivity.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+    (*entry)["facebook_dns_blocking"] = false;
+
+    int ips_count = 0;
+    for (auto const& service_and_ips : fb_service_ips) {
+        // skipping stun for now
+        if (service_and_ips.first == "stun") {
+            continue;
+        }
+        for (auto const& ip : service_and_ips.second) {
+            ips_count += 1;
```
bassosimone: Is it possible to refactor this loop to count otherwise? I've turned on `-Werror` and this does not compile because the variable named `ip` is actually unused (renaming it as `_` wouldn't work).

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+    Var<int> ips_tested(new int(0));
+    if (ips_count == *ips_tested) {
```
bassosimone: I do not understand the purpose of this code: you allocate `ips_tested` on the heap to be equal to zero and then you check whether `ips_count` is equal to it. Can't you just check whether it's equal to zero?

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+        // skipping stun for now
+        if (service == "stun") {
+            continue;
+        }
+        for (auto const& ip : service_and_ips.second) {
+            Settings tcp_options;
+            tcp_options["host"] = ip;
+            int port = 443; //XXX hardcoded
+            tcp_options["port"] = port;
+            tcp_options["net/timeout"] = 10.0; //XXX hardcoded
```
bassosimone: I think we can avoid hard-coding these values by passing the proper options.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+            templates::tcp_connect(tcp_options, tcp_cb(service, ip, port),
+                                   reactor, logger);
+        }
+    }
+    return;
+}
+
+void facebook_messenger(std::string input, Settings options,
```
bassosimone: This `options` variable is the one that you should pass around and fetch options from.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+    mk::fcompose(
+                 mk::fcompose_policy_async(),
+                 dns_many,
+                 tcp_many
+                )(NoError(),
+                  entry,
+                  FB_SERVICE_HOSTNAMES,
+                  reactor,
+                  logger,
+                  callback
```
bassosimone: If you want to pass around less stuff, you can just put everything into a `Var<Context>`.

test/nettests/facebook_messenger.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include "../nettests/utils.hpp"
+
+using namespace mk::nettests;
+using namespace mk;
+
+TEST_CASE("Synchronous facebook_messenger test") {
+    test::nettests::make_test<FacebookMessengerTest>("meek_fronted_requests.txt")
+        .run();
+}
```
bassosimone: After recent changes, you should rewrite this as:

```C++
test::nettests::make_test<FacebookMessengerTest>(test::nettests::run);
```

(Note that passing to this test the `meek_fronted_requests.txt` input file seems wrong to me)

test/nettests/facebook_messenger.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+TEST_CASE("Asynchronous meek-fronted-requests test") {
+    auto t = test::nettests::make_test<FacebookMessengerTest>("meek_fronted_requests.txt");
+    test::nettests::run_async(t);
+}
```
bassosimone: After recent changes, you should drop this. Now running sync or async tests stress the same code paths, so there is not need to check for both here in the regress tests.

joelanders:
Yep! on it.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
joelanders: done

joelanders: done

include/measurement_kit/ooni/nettests.hpp
joelanders: done

src/libmeasurement_kit/nettests/facebook_messenger.cpp
joelanders: done

src/libmeasurement_kit/ooni/facebook_messenger.cpp
joelanders: done

src/libmeasurement_kit/ooni/facebook_messenger.cpp
joelanders: done

src/libmeasurement_kit/ooni/facebook_messenger.cpp
joelanders: done

src/libmeasurement_kit/ooni/facebook_messenger.cpp
joelanders: I left this as it is because these seem kind of local to the tcp_connect() call.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
joelanders: done

src/libmeasurement_kit/ooni/facebook_messenger.cpp
joelanders: Left it as is.

test/nettests/facebook_messenger.cpp
joelanders: Done.

test/nettests/facebook_messenger.cpp
joelanders: done

joelanders:
Rebased, killing the commits your review points were attached to... but I think I addressed everything.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,221 @@
...
+        return;
+    }
+
+    // if we find any inconsistent DNS later, switch this to true
+    (*entry)["facebook_dns_blocking"] = false;
+
+    int names_count = fb_service_hostnames.size();
+    Var<int> names_tested(new int(0));
+
+    if (names_count == *names_tested) {
```
bassosimone: @joelanders I still don't get the logic of this `if`: you've just created `names_tested` to be zero. Is the objective to check whether `names_count` is zero?

joelanders: ah whoops, i changed one location and not the other

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,221 @@
...
+    };
+
+    for (auto const& service_and_hostname : fb_service_hostnames) {
+        std::string service = service_and_hostname.first;
+        std::string hostname = service_and_hostname.second;
+        templates::dns_query(entry, "A", "IN", hostname, "",
+                             dns_cb(service, hostname), options, reactor,
+                             logger);
+    }
+    return;
```
bassosimone: _style_: can you drop this `return` at the end, please?

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,221 @@
...
+            }
+            *names_tested += 1;
+            if (names_count == *names_tested) {
+                cb(NoError(), entry, *fb_service_ips, options, reactor, logger);
+                return;
+            }
+        };
+    };
+
+    for (auto const& service_and_hostname : fb_service_hostnames) {
```
bassosimone: @joelanders is `fb_service_hostnames` constant? If so, I guess we can initialize it only once and share it as a constant reference all around. I did something similar in #1272 with the constant `input_t`.

bassosimone: Especially, we can avoid passing it around and we can access it as a singleton from a `static` factory.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,221 @@
...
+                            }
+                        }
+                    }
+                }
+                // if any consistent IPs, consistent = true
+                (*entry)["facebook_" + service + "_dns_consistent"] =
+                    !(*fb_service_ips)[service].empty();
+            }
+            *names_tested += 1;
+            if (names_count == *names_tested) {
```
bassosimone: Here, can we `assert` that `names_count` is `<= *names_tested`. Also, can you declare `names_count` as `size_t` such that it cannot be negative? (Since you assign it using the `.size()` of a collection, `size_t` is the proper type anyway)

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,221 @@
...
+        cb(entry);
+        return;
+    }
+
+    // if we find any blocked TCP later, switch this to true
+    (*entry)["facebook_dns_blocking"] = false;
+
+    int ips_count = 0;
+    for (auto const& service_and_ips : fb_service_ips) {
+        // skipping stun for now
```
bassosimone: Why do we skip `stun`?

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,221 @@
...
+                (*entry)["facebook_" + service + "_blocking"] = true;
+                (*entry)["facebook_dns_blocking"] = true;
+                close_txp = false;
+            } else {
+                logger->info("tcp success to %s at %s:%d", service.c_str(),
+                    ip.c_str(), port);
+                (*entry)["facebook_" + service + "_blocking"] = false;
+            }
+            *ips_tested += 1;
+            if (ips_count == *ips_tested) {
```
bassosimone: Also here, can we `assert` that `ips_count <= *ips_tested` and make sure `ips_count` is `size_t`?

src/libmeasurement_kit/ooni/facebook_messenger.cpp
bassosimone: I'd initialize `tcp_options` using the current `options` and then override specifics as you do.

Since you are purposefully setting the port and the timeout, probably `XXX` is not a good comment, because it implies something untidy, but here it seems it's exactly what you want to do, no?

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,221 @@
...
+        FB_SERVICE_HOSTNAMES = {
+            {"stun", "stun.fbsbx.com"},
+            {"b_api", "b-api.facebook.com"},
+            {"b_graph", "b-graph.facebook.com"},
+            {"edge", "edge-mqtt.facebook.com"},
+            {"external_cdn", "external.xx.fbcdn.net"},
+            {"scontent_cdn", "scontent.xx.fbcdn.net"},
+            {"star", "star.c10r.facebook.com"}};
+
+    options["geoip_asn_path"] = "GeoIPASNum.dat";
```
bassosimone: You don't need this. With properly configured upstream options this file should already be set to point to the proper file.

src/measurement_kit/cmd/facebook_messenger.cpp
```diff
@@ -0,0 +1,33 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software. See AUTHORS and LICENSE for more
+// information on the copying conditions.
```
bassosimone: Note to self: update header as part of merging the PR.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
joelanders: Because ooni-probe does :) and I think because it uses UDP? *me reads* I guess it uses both, but we'd only be able to check the TCP bit?

bassosimone: Ah, okay: can you document that we're doing what ooni-probe does in the comment?

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "private/ooni/constants.hpp"
+#include <measurement_kit/ooni.hpp>
+
+namespace mk {
+namespace ooni {
+
+using namespace mk::report;
+
+static const std::map<std::string, std::string>
+    &FB_SERVICE_HOSTNAMES = {
```
bassosimone: I seem to recall that the most robust C++ way to initialize global objects is through factory functions. I will see whether I can find a link regarding this. Going on memory, not doing that led to strange behavior sometimes.

I also found it strange for the variable to be a reference type. But probably it is initialized as a reference to an anonymous object initialized by the initialiser list?

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+        {"b_api", "b-api.facebook.com"},
+        {"b_graph", "b-graph.facebook.com"},
+        {"edge", "edge-mqtt.facebook.com"},
+        {"external_cdn", "external.xx.fbcdn.net"},
+        {"scontent_cdn", "scontent.xx.fbcdn.net"},
+        {"star", "star.c10r.facebook.com"}};
+
+static void dns_many(Error error,
+                     Var<Entry> entry,
+                     std::map<std::string, std::string> fb_service_hostnames,
```
bassosimone: I think this can be dropped since now it's global, no?

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+                     Settings options,
+                     Var<Reactor> reactor,
+                     Var<Logger> logger,
+                     Callback<Error,
+                              Var<Entry>,
+                              std::map<std::string, std::vector<std::string>>,
```
bassosimone: Ditto

src/libmeasurement_kit/ooni/facebook_messenger.cpp
```diff
@@ -0,0 +1,220 @@
...
+        std::string hostname = service_and_hostname.second;
+        templates::dns_query(entry, "A", "IN", hostname, "",
+                             dns_cb(service, hostname), options, reactor,
+                             logger);
+    }
+}
+
+static void tcp_many(Error error,
+                     Var<Entry> entry,
+                     std::map<std::string, std::vector<std::string>> fb_service_ips,
```
bassosimone: Ditto

src/libmeasurement_kit/ooni/facebook_messenger.cpp
bassosimone: My mistake. This map needs to be passed around.

src/libmeasurement_kit/ooni/facebook_messenger.cpp
bassosimone: My mistake. We need to pass around this map.

bassosimone:
Merged as #1332

PR #1193, opened 2017-04-18
Implement OONI's WhatsApp test

joelanders:
Aiight, so if we generalize these `dns_many(), tcp_many(), http_many()` functions, these IM tests become quite small and simple. I think those functions would be a lot nicer with some kind of `collect(futures)` thing.

I'll get these done (with what we got) tomorrow. (error handling, report filling, etc.)


src/libmeasurement_kit/ooni/whatsapp.cpp
```diff
@@ -0,0 +1,211 @@
...
+            *names_tested += 1;
+            if (names_count == *names_tested) {
+                logger->info("dns_many() found %d ips", (int)ips->size());
+                cb(NoError(), *ips);
+            }
+        };
+    };
+
+    for (auto const& hostname : hostnames) {
+        Settings options; // XXX ought this to forward from above?
```
bassosimone: Yep, I guess this should be the options from above. AFAICT, you don't need to worry about clobbering options because in MK options are always passed by copy to functions precisely for this reason.

src/libmeasurement_kit/ooni/whatsapp.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+}
+
+void whatsapp(std::string input, Settings options,
+              Callback<Var<report::Entry>> callback,
+              Var<Reactor> reactor, Var<Logger> logger) {
+    std::vector<std::string>
+        WHATSAPP_WEB_URLS = { "https://web.whatsapp.com/",
+                              "http://web.whatsapp.com/" };
+    // XXX generate these with a range or something
```
bassosimone: Ideally, I guess that this can be some sort of input file. (Now that you have implemented all the IM tests, I see that they have lots in common and they mostly differ for their input. I also see that there are a number of places where we can benefit from adding more functional abstractions to the code, and perhaps we should do that so to simplify coding.)

bassosimone:
@joelanders can you resolve conflicts and fill in the entry such that it matches the entry generated by ooniprobe? I would like to merge this test into the next release of MK. Thank you!

joelanders:
@bassosimone sorry, I should have posted somewhere: I started work on moving these to the `mk::parallel()` function which removes some of the duplication. But the callback `parallel()` calls doesn't take any args other than an `Error` (so I can't pass the results of `dns_many()` to `tcp_many()`.

Do you want me to get this test working without `mk::parallel()` in the meantime?

bassosimone:
@joelanders could you fix the build?

joelanders:
Ok, it builds, but this is still WIP as we need to figure out how to determine DNS consistency in absence of that CIDR.txt.

joelanders:
Also need to remove that unneeded input parameter.

bassosimone:
The [link we were using](https://www.whatsapp.com/cidr.txt) now says:
> Dear partners, Please note that we have migrated the latest IP pools of WhatsApp to Facebook Mobile Partner Portal. Feel free to browse to the Settings page of the portal and download the latest WhatsApp IP pool: https://fb.me/mpp_support 
Further IP pool updates are also done through the portal and are no longer distributed via email or through WhatsApp web site.
If you have not yet registered on the Mobile Partner Portal or have difficulties accessing it - please request access through the following form and we'll be happy to assist: https://fb.me/mpp_access
For any technical requests please contact us through the Support section of the portal: https://fb.me/mpp_support 
WhatsApp team

hellais:
> @hellais how is ooni-probe coping with this issue?

This change happened since we implemented the test. We have currently not dealt with it. I guess that if they have moved onto using Facebook IP space then we can determine consistency similarly to the facebook messenger test (i.e. checking the facebook ASN).

@joelanders can you check to see if now the app is connecting to facebook IP addresses?

bassosimone:
Since the point of CIDR.txt is not clear, I'd defer this to v0.8.0-beta.

src/libmeasurement_kit/ooni/whatsapp.cpp
```diff
@@ -0,0 +1,575 @@
...
+    for (auto net : nets) {
+        auto res = ip_in_net(ip, net);
+        if (!!res && (res.as_value() == true)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+// XXX not even beginning to fill in entry
```
bassosimone: @joelanders is this comment still actual; if so, can you fill the entry?

src/libmeasurement_kit/ooni/whatsapp.cpp
```diff
@@ -0,0 +1,575 @@
...
+    int ips_count = ips.size() * 2;
+    Var<int> ips_tested(new int(0));
+    if (ips_count == *ips_tested) {
+        cb(NoError());
+        return;
+    }
+
+    auto tcp_cb = [=](std::string ip, int port) {
+        return [=](Error err, Var<net::Transport> txp) {
+            bool close_txp = true; // if connected, we must disconnect
```
bassosimone: @joelanders you can actually simplify the `close_txp` logic. I recently changed connect such that it should always return a transport, perhaps not connected but still valid. Also, there is actually no need to pass a callback to `close()`: you can pass `nullptr` instead. In `v0.9.0` I'll remove the requirement of passing a callback. (If we leak, we notice with Valgrind, so there is no need to have this leak detection mechanism baked into the way in which we close a Transport.)

bassosimone: (See how I refactored the code in telegram or facebook_messanger)

src/libmeasurement_kit/ooni/whatsapp.cpp
```diff
@@ -0,0 +1,575 @@
...
+        };
+    };
+
+    for (auto const& ip : ips) {
+        // XXX hardcoded
+        std::vector<int> ports {443, 5222};
+        for (auto const& port : ports) {
+            options["host"] = ip;
+            options["port"] = port;
+            options["net/timeout"] = 10.0; //XXX check if set upstream?
```
bassosimone: The timeout is set by default. I'd initialize the local options variable from the options of the test.

src/libmeasurement_kit/ooni/whatsapp.cpp
```diff
@@ -0,0 +1,575 @@
...
+
+    auto dns_cb = [=](std::string hostname) {
+        return [=](Error err, Var<dns::Message> message) {
+            if (!!err) {
+                logger->info("whatsapp: dns error for %s", hostname.c_str());
+            } else {
+                std::vector<std::string> this_ips;
+                for (auto answer : message->answers) {
+                    // XXX i don't get this answer format
+                    // copying from web_connectivity
```
bassosimone: It's basically that, depending on the response type, the answer is stored in different fields.

How would you like it to be instead?

src/libmeasurement_kit/ooni/whatsapp.cpp
bassosimone: You can do `Settings local_options = options;` to make a copy in a in-loop variable.

src/libmeasurement_kit/ooni/whatsapp.cpp
```diff
@@ -0,0 +1,575 @@
...
+//            dns_many(WHATSAPP_ENDPOINT_HOSTNAMES, entry,
+//                    [=](Error err, std::vector<std::string> ips) {
+//                logger->info("done dns-testing whatsapp endpoints");
+//                tcp_many(ips, entry, [=](Error err) {
+//                    logger->info("done tcp-testing whatsapp endpoints");
+//                    callback(entry);
+//                }, reactor, logger);
+//            }, reactor, logger);
+//        }, reactor, logger);
+//    }, reactor, logger);
```
bassosimone: This commented out code can probably be removed, right?

src/measurement_kit/cmd/whatsapp.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace whatsapp {
+
+#define USAGE                                       \
+    "usage: measurement_kit [options] whatsapp\n"   \
+
+int main(std::list<Callback<BaseTest &>> &initializers, int argc, char **argv) {
+    mk::nettests::WhatsappTest test;
+    int ch;
+    while ((ch = getopt(argc, argv, "B:f:")) != -1) {
```
bassosimone: Uhm, here you actually support `-B <arg>` and `-f <arg>`. You should probably empty the format string.

test/nettests/whatsapp.cpp
```diff
@@ -0,0 +1,98 @@
...
+    std::vector<uint8_t> ex = { 1, 0, 0, 0 };
+    std::vector<uint8_t> bs = mk::ooni::ip_to_bytes("1.0.0.0");
+    REQUIRE( bs == ex );
+
+    ex = { 255, 255, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+    bs = mk::ooni::ip_to_bytes("ffff::");
+    REQUIRE( bs == ex );
+}
+
+TEST_CASE("same_pre"){
```
bassosimone: You can probably _group_ all the `same_pre` tests into the same `TEST_CASE` and use `SECTION` for each specific sub-case so to have logically similar things all together. Also, can you provide more descriptive test cases descriptions?

test/nettests/whatsapp.cpp
```diff
@@ -0,0 +1,98 @@
...
+     //no_err = mk::ooni::net_match4("0.0.1.0", "0.0.1.0/24", inclusion);
+     //no_err = mk::ooni::net_match4("0.1.0.0", "0.1.0.0/24", inclusion);
+     //no_err = mk::ooni::net_match4("1.0.0.0", "1.0.0.0/24", inclusion);
+     //no_err = mk::ooni::net_match4("1.0.0.0", "1.0.0.0/24", inclusion);
+     //no_err = mk::ooni::net_match4("2000::/3", "2000::/3", inclusion);
+ //    REQUIRE( no_err == true );
+ //    REQUIRE( inclusion == true );
+     //bool b2 = mk::ooni::net_match4("10.10.1.90", "10.10.1.32/27", inclusion);
+ //    REQUIRE( no_err == true );
+ //    REQUIRE( inclusion == false );
```
bassosimone: Should we keep this code and uncomment?

test/nettests/whatsapp.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include "../nettests/utils.hpp"
+
+using namespace mk::nettests;
+using namespace mk;
+
+//TEST_CASE("Synchronous whatsapp test") {
+//    test::nettests::make_test<WhatsappTest>("meek_fronted_requests.txt")
+//        .run();
+//}
```
bassosimone: You should uncomment this. You don't need to provide any filepath argument in this case, AFAICT.

test/nettests/whatsapp.cpp
```diff
@@ -0,0 +1,98 @@
...
+//
+//TEST_CASE("Asynchronous meek-fronted-requests test") {
+//    auto t = test::nettests::make_test<WhatsappTest>("meek_fronted_requests.txt");
+//    test::nettests::run_async(t);
+//}
```
bassosimone: This test is not needed.

src/libmeasurement_kit/ooni/whatsapp.cpp
joelanders: acked

bassosimone:
@joelanders it seems [the build fails](https://travis-ci.org/measurement-kit/measurement-kit/jobs/277415649#L1780) for a stupid reason: recently I `s/Var/SharedPtr/g` after a discussion with @hellais and this branch needs to be converted to be able to compile. Can you do that?

src/libmeasurement_kit/ooni/whatsapp.cpp
joelanders: acked

src/libmeasurement_kit/ooni/whatsapp.cpp
joelanders: acked (and I can't reply to the other options comment for some reason)

src/libmeasurement_kit/ooni/whatsapp.cpp
joelanders: acked

src/measurement_kit/cmd/whatsapp.cpp
joelanders: acked

test/nettests/whatsapp.cpp
joelanders: old function, so deleted.

test/nettests/whatsapp.cpp
joelanders: acked

test/nettests/whatsapp.cpp
joelanders: acked

joelanders:
Ok I addressed all of your comments. If I try to do the "always close tcp transport" thing, I get some runtime exception. Maybe I'm doing something wrong, so I'll have another look tomorrow when I'm fresh. Last commit here turns it from working to not-working.

bassosimone:
> Ok I addressed all of your comments.
Thanks!

> If I try to do the "always close tcp transport" thing, I get some runtime exception. Maybe I'm doing something wrong, so I'll have another look tomorrow when I'm fresh. Last commit here turns it from working to not-working.
Ah, I've seen the diff. What I meant is: you can refactor the code such that you can _always_ close the transport, regardless of whether you got an error or not. Also, no need to pass any function to `close`, just pass `nullptr`, but put the code that was inside that function below `close()`.

See, e.g.: https://github.com/measurement-kit/measurement-kit/blob/792dff75af8896d5bf24356cc63b95f3c38f525c/src/libmeasurement_kit/ooni/telegram.cpp#L23

bassosimone:
Let's try to merge with master and see whether this fixes.

joelanders:
One more thing we should figure out: do we want to test all e{1-16}.whatsapp.net endpoints or maybe a subset of them?

joelanders:
Ok, it picks a random endpoint to test (default) or all of them (with -e option, like ooniprobe). Also fixed a bug and squashed some compiler warnings. I think this may be ready for a final approval?

bassosimone:
Yeah, it seems so, thanks! I am going to create an integration branch for this PR and to apply some small fixes (e.g. reindent and the like) in such branch. 🐳

PR #1218, opened 2017-05-08
Allow multiple DNS servers

bassosimone:
Supersedes #913.


bassosimone:
Anyway: we don't need this. For common cases, we use the system resolver. When we need to query a _specific_ DNS server, we only want to query it. Closing.

PR #1220, opened 2017-05-09
PoC: Better resources management

bassosimone:


include/measurement_kit/common/pool.hpp
```diff
@@ -0,0 +1,44 @@
...
+    Pool();
+    ~Pool();
+
+    template <typename T, typename... A> Var<T> alloc(A &&... a) {
+        Var<T> p = Var<T>::make(std::forward<A>(a)...);
+        active_.insert(p);
+        return p;
+    }
+
+    template <typename T> void free(Var<T> p) {
```
bassosimone: @hellais suggested to call this something like `free_later`

bassosimone:
So, I actually recently figured out that the real issue we have with resources (mitigated by all the scattered `call_soon()` at end of test) is basically that we pass `this` to lambda closures. We should not do that. If we don't do that, there is no need to have garbage-collector-like functionality. Closing the PR.

PR #1228, opened 2017-05-10
Use also SSLv3 and SSLv3 with web_connectivity

bassosimone:


hellais:
This looks good to me

bassosimone:
Thanks!

PR #1231, opened 2017-05-12
Releases v0.6.x

bassosimone:


bassosimone:
Ah, I see that regress tests are failing. The reason is quite stupid: I forgot to backport a small diff that changes regress tests that explicitly set the DNS nameserver to use the libevent engine.

This is not an obstacle for releasing and shipping v0.6.2. I will anyway make, later today, a v0.6.3 that incorporates this small diff, such that we can merge this branch in stable.

Issue #1233, opened 2017-05-15
dns system resolver is not asynchronous


bassosimone:
I believe this to be true based on my observation of reading the output of a long running web connectivity where all log output of three parallel tests blocked when a single DNS call was pending. After a cursory look at [std::launch](http://en.cppreference.com/w/cpp/thread/async) manual page, it seems we did not consider this statement implications:

> If the std::future obtained from std::async is not moved from or bound to a reference, the destructor of the std::future will block at the end of the full expression until the asynchronous operation completes, essentially making code such as the following synchronous:
>
> ```C++
>    std::async(std::launch::async, []{ f(); }); // temporary's dtor waits for f()
>    std::async(std::launch::async, []{ g(); }); // does not start until f() completes
>   ```

😠

bassosimone:
I believe we can close this issue, since it landed in #1234 (what a nice PR number, BTW). As such, this will be fixed in the next release of measurement-kit.

PR #1234, opened 2017-05-16
Refactor DNS for v0.7.x

bassosimone:
1. switch to C++14 to make use of moving within lambda capture

2. implement a worker set of threads

3. implement a true asynchronous `getaddrinfo()` (thus fixing #1233)

4. split former system resolver implementation in small functions

5. add some useful, simple templates


acinclude.m4
```diff
@@ -214,60 +214,22 @@ AC_DEFUN([MK_AM_REQUIRE_C99], [
   fi
 ])
 
-AC_DEFUN([MK_AM_REQUIRE_CXX11], [
+AC_DEFUN([MK_AM_REQUIRE_CXX14], [
```
bassosimone: Switching to C++14, now supported by compilers. [Debian testing is currently using g++ 6.3(https://packages.debian.org/search?keywords=gcc&searchon=names&suite=all&section=all)], where [C++14 is the default compiler mode when compiling c++](https://gcc.gnu.org/projects/cxx-status.html#cxx14).

src/libmeasurement_kit/dns/system_resolver.cpp
```diff
@@ -1,18 +0,0 @@
...
-
-void system_resolver(QueryClass dns_class, QueryType dns_type, std::string name,
-                     Callback<Error, Var<Message>> cb, Settings settings,
-                     Var<Reactor> reactor, Var<Logger> logger) {
-    system_resolver_impl(dns_class, dns_type, name, cb, settings, reactor,
-            logger);
-}
-
-} // namespace dns
-} // namespace mk
```
bassosimone: Code removed. @AntonioLangiu initially organized the code to have everything into a single private header. Later, I started splitting code in foo.hpp, foo_impl.hpp, and foo.cpp. The argument for doing that was that you catch compile errors early because you instantiate templates in the related filed. But, after reading again the code, I realized I like more the approach of @AntonioLangiu, because it's more compact. So, I'm basically moving again everything into a header.

src/libmeasurement_kit/dns/system_resolver.hpp
```diff
@@ -6,18 +6,59 @@
...
+    message->queries.push_back(query);
+
+    getaddrinfo_async<getaddrinfo, inet_ntop>(
+        name, hints, reactor, logger, Worker::global(),
+        [ message = std::move(message),
+          cb = std::move(cb) ](Error error, std::vector<Answer> answers) {
+            message->answers = std::move(answers);
+            cb(error, message);
+        });
+}
```
bassosimone: Code adapted from previous system resolver. I removed the class for keeping the state, since now we can move around things using lambda, which leads to more compact code. Also, modified the code such that the last call is actually a call to the async implementation of getaddrinfo.

src/libmeasurement_kit/dns/system_resolver_impl.hpp
```diff
@@ -1,194 +0,0 @@
...
-        case EAI_SOCKTYPE:
-            resolver_error = NotSupportedAISocktypeError();
-            break;
-        default:
-            resolver_error = ResolverError();
-            break;
-        }
-        ctx->reactor->call_soon([=]() { callback(resolver_error, nullptr); });
-        return;
-    }
```
bassosimone: This switch is now a function in `getaddrinfo_async.hpp` that just maps the error.

src/libmeasurement_kit/dns/system_resolver_impl.hpp
```diff
@@ -1,194 +0,0 @@
...
-                [=]() { callback(InetNtopFailureError(), nullptr); });
-            return;
-        }
-        if (p->ai_family == AF_INET) {
-            answer.ipv4 = std::string(abuf);
-        } else if (p->ai_family == AF_INET6) {
-            answer.ipv6 = std::string(abuf);
-        }
-        message->answers.push_back(answer);
-    }
```
bassosimone: This code to process result is now a specific function in `getaddrinfo_async.hpp` that just does that task.

src/libmeasurement_kit/dns/system_resolver_impl.hpp
```diff
@@ -1,194 +0,0 @@
...
-
-    query.type = dns_type;
-    query.qclass = dns_class;
-    query.name = name;
-
-    ctx->message->queries.push_back(query);
-
-    std::async(std::launch::async, resolve_async_impl<getaddrinfo, inet_ntop>,
-               ctx.release());
```
bassosimone: This is the piece of code move in `system_resolver.hpp` and modified as explained above.

bassosimone:
@hellais, this fixes the bug that system DNS was not async (see #1233). In working this feature, I learned how to use C++14 to move state around with lambdas, leading to simpler code, and also learned a few template tricks that simplify life when writing algorithms. Each patch should be standalone (possibly depending on previous patches). As many of them are simple, they can be cherry-picked and applied to master quite easily. After this is approved, I plan perhaps on adding some more documentation fixes, if needed, before merging this into master.

src/libmeasurement_kit/dns/getaddrinfo_async.hpp
```diff
@@ -0,0 +1,121 @@
...
+    XX(0, NoError)                                                             \
+    XX(EAI_AGAIN, TemporaryFailureError)                                       \
+    XX(EAI_BADFLAGS, InvalidFlagsValueError)                                   \
+    XX(EAI_FAIL, NonRecoverableFailureError)                                   \
+    XX(EAI_FAMILY, NotSupportedAIFamilyError)                                  \
+    XX(EAI_MEMORY, MemoryAllocationFailureError)                               \
+    XX(EAI_NONAME, HostOrServiceNotProvidedOrNotKnownError)                    \
+    XX(EAI_OVERFLOW, ArgumentBufferOverflowError)                              \
+    XX(EAI_SERVICE, NotSupportedServnameError)                                 \
+    XX(EAI_SOCKTYPE, NotSupportedAISocktypeError)
```
bassosimone: @hellais this is the specific place where we can seek to reduce the number of errors.

hellais: It seems like you are missing some of the possible error values for `getaddrinfo`, namely:

* `EAI_NODATA `: "No address associated with NAME."
* `EAI_ADDRFAMILY `: "Address family for NAME not supported"
* `EAI_SYSTEM`: "System error returned in `errno`"

hellais: I would suggest we leave these as separate errors (in terms of `Error` definition), but map the string encoding of the error to a set of known ooni error code.

The mapping I would suggest is as follows (for reference ooni error string [are defined here](https://github.com/TheTorProject/ooni-spec/blob/master/data-formats/df-000-base.md#error-strings))

* `EAI_AGAIN`->`dns_name_failure` (Note: This could create confusion in ooniprobe as twisted does not map this `EAGAIN`)
* `EAI_BADFLAGS `->`unknown_failure dns_bad_flags`
* `EAI_FAIL, NonRecoverableFailureError `->`dns_name_failure`
* `EAI_FAMILY, NotSupportedAIFamilyError `->`unknown_failure dns_family_failure`
* `EAI_MEMORY, MemoryAllocationFailureError `->`unknown_failure dns_memory_failure`
* `EAI_NONAME, HostOrServiceNotProvidedOrNotKnownError `->`dns_name_error`
* `EAI_OVERFLOW, ArgumentBufferOverflowError `->`unknown_failure dns_overflow_failure`
* `EAI_SERVICE, NotSupportedServnameError `->`unknown_failure dns_service_failure` (I think the mapping in MK `NotSupportedServnameError` is misleading. I would instead write it as `NotSupportedServiceError` as `Servname` seems to imply it has something to do with DNS server)
* `EAI_SOCKTYPE, NotSupportedAISocktypeError `->`unknown_failure dns_socktype_failure`

bassosimone:
I am going to merge this into the integration branch. After this merge, I am going to do more refactoring that involves moving around files. I would like to keep this diff separate from that diff.

PR #1235, opened 2017-05-17
Composition of arbitrary callables

bassosimone:
This proof of concept shows how we could implement stackless coroutines and continuation. We start by implementing continuation with templates. For example, the following compiles with `clang++ -Wall -Wextra -I include -std=c++14` and produces the expected result:

```C++
int main() {
    mk::fcompose(
        [](double d, int off) { return d * 2 + off; },
        [](double d) { return d + 1; },
        [](double d) { return std::make_tuple(3, std::to_string(d)); },
        [](int x, std::string s) {
            std::cout << x << " " << s << "\n";
        }
    )(3.17, 7);
}
```

This functionality can be used to chain callbacks called in response to events (which is part of what @joelanders would like to see implemented, if I understood correctly).

Then, by adding a `YieldError` and by making sure the code that calls callbacks is prepared to deal with such error, we can have stackless coroutines, as shown in the modified HTTP example.

Rather than having a `YieldError`, another possibility would be to pass a callable a callback to be called when it is time to pass the token to the next "stage" of the pipeline.


acinclude.m4
```diff
@@ -226,7 +226,7 @@ AC_DEFUN([MK_AM_REQUIRE_CXX11], [
      AC_MSG_RESULT([no])
      AC_MSG_ERROR([a C++11 compiler is required])
     ])
-  CXXFLAGS="$mk_saved_cxxflags -std=c++11"
+  CXXFLAGS="$mk_saved_cxxflags -std=c++14"
```
bassosimone: Quick and dirty way to force C++14 so to have some of its goodies

example/http/request.cpp
```diff
@@ -68,24 +107,26 @@ int main(int argc, char **argv) {
...
+            }
+            http::request_send(
+                t, settings, headers, body, [=](Error e, Var<http::Request> r) {
+                    if (e) {
+                        throw e;
+                    }
+                    mk::warn("Request sent... waiting for response...");
+                    t->on_data(fcompose(http_read_response_coro(),
+                                        fcompose(add_request(std::move(r)),
+                                                 print_response_and_exit())));
```
bassosimone: With more template-foo, `fcompose` will be able to:

1. chain more than two functors at once

2. deduce the argument types of the first functor and apply them, so to avoid the need to pass the first argument as a `std::function<>` explicitly, which is very annoying

So, this piece of code should actually be imagined as:

```C++
t->on_data(fcompose(
    http_read_response_coro(),
    [r = std::move(r)](http::Response x) {
        x.request = r;
        return x;
    },
    [=](http::Response resp) {
        // Here code of `print_response_and_exit`
    });
```

example/http/request.cpp
```diff
@@ -30,6 +32,43 @@ static bool set_header(http::Headers &headers, const std::string option) {
     return true;
 }
 
+static std::function<http::Response(net::Buffer)> http_read_response_coro() {
```
bassosimone: Note: of course this could made be more robust to stop after a single message has been read.

bassosimone:
With also the possibility of composing functions asynchronously, the example becomes:

Full example program:

```C++
int main() {
    mk::fcompose_async(
        [](double d, int off, mk::Callback<double> done) { done(d * 2 + off); },
        [](double d, mk::Callback<double> done) { done(d + 1); },
        [](double d, mk::Callback<int, std::string> done) {
            done(3, std::to_string(d));
        },
        [](int x, std::string s, mk::Callback<> done) {
            std::cout << "async: " << x << " " << s << "\n";
            done();
        })(3.14, 5, []() {
            mk::fcompose_sync(
                [](double d, int off) { return d * 2 + off; },
                [](double d) { return d + 1; },
                [](double d) { return std::make_tuple(3, std::to_string(d)); },
                [](int x, std::string s) {
                    std::cout << "sync: " << x << " " << s << "\n";
                })(3.14, 5);
        });
}
```

bassosimone:
This branch scope has now changed to only implement composition of arbitrary callables following different types of policies. It will be merged into `v0.7.x-slim` because it now seems mature.

bassosimone:
What is missing before merging is: some more documentation, more thorough tests.

bassosimone:
Here I was a bit swamped with handling of exceptions. For this reason, I extracted #1262 from this pull request, where I can focus on the easy part of finalizing code that is stable.

example/common/fcompose.cpp
```diff
@@ -7,16 +7,14 @@
...
-                          });
-    f(0, 4, 7);
+    auto func = mk::fcompose(mk::fcompose_policy_async_and_route_exceptions(
+                                   [](const std::exception &exc) {
+                                       std::clog
+                                             << "Error occurred: " << exc.what()
+                                             << "\n";
+                                   }),
+                             sum3, tostr, print_fail);
+    func(0, -4, -7, []() { /* NOTHING */ });
```
bassosimone: If we want to merge, we should fix this

bassosimone:
@joelanders, I have merged into `v0.7.x` integration branch the basic part of this work. Now, it is basically possible to compose functions in the sync fashion (f-composed-g that takes in input f arguments and returns the return value of g) and in the async fashion (f-composed-g that takes as arguments the arguments of f minus f's callback and, as last argument, a callback that receives as parameters the arguments that are received by g's callback).

What is left in this branch is the possibility to route exceptions automatically.

When composing functions the sync way, I do not actually see the need to treat exceptions in any specific way. It suffices to surround the invocation of the composed function with `try...catch` (or to use the new abstraction added in `sandbox.hpp`). When composing async functions, I still have some conceptual problems related to when and how it is optimal to route exceptions.

Ideally, what I wanted to achieve was that either the composed async function succeded, thus reaching the final registered callback, or there was an exception and thus the exception handler was called. I really don't like it when, for example, you reach the final callback, there is an exception there, and then this is routed through the exception handler. I actually think I was able to reach this objective by wrapping every code block, minus the final callback, such that exceptions are routed.

However, there is one more problem. If inside the code of one of the callbacks in the chain you do a `call_soon` to delay the execution of a piece of code, you then loose the protection that your code is wrapped against exceptions. Meaning that, if the delayed bit of code throws, the whole of measurement-kit goes down. Considering this, I am leaning towards not filtering exceptions automatically as part of the composing functions together functionality. In general, I think we should avoid exceptions in our code, since they don't naturally mix and match well with async code. Then, when we have code that can throw, we can instead just use the code in `sandbox.hpp` to filter and forward exceptions.

This strikes me as an approach where it is simpler to reason about the routing of exceptions. That is, you review the code, and you explicitly can say whether exceptions are routed or not. Conversely, adding exception filtering as part of composing seems to me something that requires a lot of reasoning when reviewing the code for correctness.

What do you think?

bassosimone:
I discussed the feature or routing exceptions with callback in here with @hellais and we were both a bit skeptical about this approach. This is enough to close this PR.

PR #1238, opened 2017-05-19
Implement Captive Portal detection

joelanders:
This is the hard part done -- as usual, I haven't filled the report in yet, or written the unit/integration tests, but it hits up all the remote resources.

It's not a straight port of the ooniprobe test: I looked up what Microsoft, Apple, and Google currently do in their products. (Google uses 4 different endpoints in the last 4 Android releases, so I included all of them). (Also, I'm proceeding on the assumption the 0x20 randomized-case part of the ooniprobe test doesn't actually have anything to do with captive portals, so I didn't include it here.)

We could make these inputs less of a compile-time fixture if we had them in an input file, but I think such a file would be pretty messy if we stuck to the line-delimited text structure. Would be really cool if we could specify the `http::Response -> boolean` lambdas in an input file, but I think it'd have to be like a DLL kinda thing, and that's not so cool...


joelanders:
Ok this should be nearing completion. As it's doing the tests a different way than ooni-probe's, we're going to have to change the [spec](https://github.com/TheTorProject/ooni-spec/blob/master/test-specs/ts-010-captive-portal.md) as well. Here's what I'm proposing the new report look like:
```
    "vendor_tests": [
      {
        "name": "Microsoft HTTP",
        "failure": null,
        "actual_body": "Microsoft NCSI",
        "actual_status": 200,
        "expected_body": "Microsoft NCSI",
        "url": "http://www.msftncsi.com/ncsi.txt",
        "expected_status": null,
        "result": true
      },
      {
        "name": "Apple HTTP 1",
        "failure": null,
        "actual_body": "<HTML><HEAD><TITLE>Success</TITLE></HEAD><BODY>Success</BODY></HTML>\n",
        "actual_status": 200,
        "expected_body": "<HTML><HEAD><TITLE>Success</TITLE></HEAD><BODY>Success</BODY></HTML>\n",
        "url": "http://captive.apple.com",
        "expected_status": null,
        "result": true
      },
      {
        "name": "Apple HTTP 2",
        "failure": null,
        "actual_body": "<HTML><HEAD><TITLE>Success</TITLE></HEAD><BODY>Success</BODY></HTML>\n",
        "actual_status": 200,
        "expected_body": "<HTML><HEAD><TITLE>Success</TITLE></HEAD><BODY>Success</BODY></HTML>\n",
        "url": "http://captive.apple.com/hotspot_detect.html",
        "expected_status": null,
        "result": true
      },
      {
        "name": "Android KitKat HTTP",
        "failure": null,
        "actual_body": "",
        "actual_status": 204,
        "expected_body": null,
        "url": "http://clients3.google.com/generate_204",
        "expected_status": "204",
        "result": true
      }
]
```
(etc.)

I think having the expected/actual keys might make the UI nicer and easier later (at the cost of some redundancy, as the "expected" keys are already in the code). We can tell the user "expected to get a 204, but got a 302."

joelanders:
https://github.com/TheTorProject/ooni-spec/pull/87
^ link to ooni-spec with proposed format change.

bassosimone:
Superseded by #1272.

PR #1244, opened 2017-05-24
Start requiring move semantic for callbacks

bassosimone:
This diff starts requiring move semantic for poller callbacks
as well as for the worker callbacks.

Moving in callbacks is a good pattern, because it enforces single
ownership of said callbacks, which helps with multiple threads.

While there, also:

- revamp the documentation of the reactor

- make sure in the reactor we use new names as real functions
  and compatibility names as aliases

- provide an implementation of Reactor::pollfd() where the
  callback is the last argument (useful when guessing the
  position of the callback with templates) but retain also
  compatibility with previous API

- propagate changes

- write basic unit testing for Worker addressing a case that
  has been raised yesterday by @AntonioLangiu


doc/api/common/reactor.md
```diff
@@ -64,85 +97,137 @@ on the heap whose lifecycle is manager using a `Var<>` smart pointer.
...
+
+The `call_soon`, `call_later`, `loop_with_initial_event`, `loop`, `loop_once`,
+`break_loop`, `run_with_initial_event`, `run`, and `stop` functions are
+syntactic sugar that call the respective method of the global reactor (i.e.
+the one obtained with `Reactor::make()`.
+
+# GUARANTEES
+
+1. it is safe to call `global` and `global_remote` concurrently from
+   multiple threads.
```
bassosimone: Currently not true. Will be fixed by a subsequent PR.

PR #1240, opened 2017-05-23
Release v0.7.x [superseded by #1250]

bassosimone:
Superseded by #1250. I am keeping this open because there are some patches that I'd like to cherry-pick into master after we have merged #1250.

PR #1250, opened 2017-05-28
Integration branch for v0.7.x

bassosimone:
This is similar to #1240, except that it contains just the commits we need.


test/dns/system_resolver.cpp
```diff
@@ -5,159 +5,119 @@
...
-        system_resolver_impl<getaddrinfo, null_inet_ntop>(
-            "IN", "A", "neubot.org",
-            [](Error e, Var<Message>) {
-                REQUIRE(e == InetNtopFailureError());
-                break_loop();
-            },
-            {}, Reactor::global(), Logger::global());
-    });
+    run_system_resolver<getaddrinfo, fail_inet_ntop>(
+        "IN", "A", "neubot.org",
```
hellais: Since we don't own this domain maybe we oughta use something else here. Would it be useful to setup some A records on `ooni.io` just for doing CI?

include/measurement_kit/nettests/runnable.hpp
```diff
@@ -47,13 +48,18 @@ class Runnable : public NonCopyable, public NonMovable {
...
+    std::list<std::string> test_helpers_bouncer_names();
+
   private:
     report::Report report;
     tm test_start_time;
     std::deque<std::string> inputs;
     double beginning = 0.0;
 
     void run_next_measurement(size_t, Callback<Error>, size_t, Var<size_t>);
+    void contact_bouncer(Callback<Error>);
```
hellais: <bikeshed>
maybe `query_bouncer` is a more accurate name?
</bikeshed>

src/libmeasurement_kit/ooni/bouncer_impl.hpp
```diff
@@ -64,16 +64,25 @@ void post_net_tests_impl(std::string base_bouncer_url, std::string test_name,
...
-    settings["http/url"] = base_bouncer_url + "/net-tests";
-    settings["http/method"] = "POST";
+    std::string bbu = base_bouncer_url + "/bouncer/net-tests";
+    std::string bm = "POST";
+    settings["http/url"] = bbu;
+    settings["http/method"] = bm;
+
+    logger->debug("%sing to '%s' data '%s'", bm.c_str(), bbu.c_str(),
+                  request.dump().c_str());
     http_request(settings, {{"Content-Type", "application/json"}},
```
hellais: Maybe this can benefit from the JSON helpers added in some later commits.

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -284,62 +284,124 @@ std::string Runnable::generate_output_filepath() {
...
-
-                // Run `parallelism` measurements in parallel
-                Var<size_t> current_entry(new size_t(0));
-                mk::parallel(
-                    mk::fmap<size_t, Continuation<Error>>(
-                        mk::range<size_t>(options.get("parallelism", 3)),
-                        [=](size_t thread_id) {
-                            return [=](Callback<Error> cb) {
-                                run_next_measurement(thread_id, cb, num_entries,
+    contact_bouncer([=](Error error) {
```
hellais: Will the bouncer lookup be skipped if the user manually overrides the bouncer settings?

doc/api/common/async_runner.md
```diff
@@ -0,0 +1,167 @@
+# NAME
```
hellais: Maybe you want to cleanup the commit message.

test/dns/system_resolver.cpp
bassosimone: Yes, I think this would be better!

bassosimone:
> We should probably also add docs for these methods too though.

I was thinking about not documenting them because they are more an implementation detail to make the code very compact and reusable than something interesting from an API point of view.

include/measurement_kit/nettests/runnable.hpp
bassosimone: Yes, I agree

src/libmeasurement_kit/ooni/bouncer_impl.hpp
bassosimone: Yes, that is true

doc/api/ooni/orchestrate.md
```diff
@@ -0,0 +1,233 @@
...
+The `settings` attribute allows to specify optional settings. By default, an
+empty settings object is used, meaning that defaults will always be used.
+
+The `available_bandwidth` attribute indicates how much bandwidth you would
+like to use for automatic tests run through orchestration. By default, this
+attribute is empty, meaning that no bandwidth quota will be enforced.
+
+The `device_token` attribute uniquely identifies the app inside a device, and
+is only meaningful in the mobile context. This token is generated by the mobile
+OS for the current application and is used to identify it.
```
hellais: Maybe it's useful here to link to the docs of what this means on Android and iOS: 
iOS docs: https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW13

Android docs:
https://firebase.google.com/docs/cloud-messaging/android/client#sample-register

doc/api/ooni/orchestrate.md
```diff
@@ -0,0 +1,233 @@
...
+
+The `events_url` attribute contains the URL used to query the `events` web
```
hellais: Maybe you can link to: https://github.com/TheTorProject/proteus/tree/master/proteus-events

src/libmeasurement_kit/nettests/runnable.cpp
bassosimone: There is currently no way to disable the bouncer with options. There is a `use_bouncer` boolean in the class for tests that don't need the bouncer, but probably all tests need it.

doc/api/ooni/orchestrate.md
```diff
@@ -0,0 +1,233 @@
...
+of the country in which the probe currently is. If not set, this
+attribute will be filled automatically by MeasurementKit when the
+actual request for the orchestrator is prepared.
+
+The `probe_family` attribute allows to bind a set of different
+probes to a common family of probes. This attribute is currently
+reserved for future use.
+
+The `registry_url` attribute contains the URL used to query the
+`registry` web service of the orchestrator. By default, this attribute
```
hellais: Maybe you can link to: https://github.com/TheTorProject/proteus/tree/master/proteus-registry

doc/api/common/async_runner.md
bassosimone: What commit message are you referring to? "Further refine the runner model"?

hellais:
Overall the diff looks good to me. I will do some further local testing and another pass at reading the code and understanding it.

hellais:
> Overall the diff looks good to me. I will do some further local testing and another pass at reading the code and understanding it.

I did the following:

* [x] Compiled the branch locally
* [x] Ran all unittests and checked the passed
* [x] Ran the orchestration tool
* [x] Confirmed that clients are created and updated on the backend admin interface

In doing so I learned about the `orchestrator_secrets.json` store, that is quite neat, though I am not yet fully convinced this library should be responsible for creating a file on disk.

On that I have 2 questions:

1. Is it possible to disable the creation of the `orchestrator_secrets.json` file?
2. Would it maybe make sense to make the default behaviour be that of **not** creating the `orchestrator_secrets.json` file?

bassosimone:
> In doing so I learned about the orchestrator_secrets.json store, that is quite neat, though I am not yet fully convinced this library should be responsible for creating a file on disk.

The approach motivating the creation of the file is the following:

1. you basically only need to tell the library what you want and it does that (this is not totally clear as of this branch, but if you look at #1260 - which is what I am about to merge in here - you see that, for example, you don't even need to register a probe, because that is done automatically by the library, when you express your intent of registering a probe and the library figures out you're not registered yet).

2. the file on disk is *just a mean to the end described in the previous point*: since there is the file on the disk, one doesn't need to pass around parameters (like passwords etc) and, as a consequence, it is possible to compose more easily functions (e.g. you can call `update` and then the code can figure out that what you actually want is instead to register).

That is, the file is an implementation detail that helped me to reach a design goal (making the API super simple to use and allowing the programmer to express intent rather than what exactly to do).

I probably tend to design APIs where one express intent but cannot be more precise and tell the code exactly what to do, which is something that may annoy who uses the library. On the other hand, I often do that to make it *obvious* how to use the code without the need to think too much about it.

> On that I have 2 questions:
>
> 1. Is it possible to disable the creation of the orchestrator_secrets.json file?
> 2. Would it maybe make sense to make the default behaviour be that of not creating the orchestrator_secrets.json file?

I'd rather not do that. In my opinion, either the model based on the file is good, or we'd rather find another way. But, I don't like very much the case in which different behaviors are possible.

My understanding of the rationale of your question, by the way, is that you don't like that you are not given control over exactly to write the state. You probably are thinking that maybe you want to write the state on a database, rather than on a file on disk, and this API does not allow you to do that, and you're pissed.

Is my understanding of your concerns correct? If so, I'll see to change the API to accommodate this need.

hellais:
> My understanding of the rationale of your question, by the way, is that you don't like that you are not given control over exactly to write the state. You probably are thinking that maybe you want to write the state on a database, rather than on a file on disk, and this API does not allow you to do that, and you're pissed.

That is one aspect, the other aspect is that I find it very unusual and unexpected for a library to start writing files on my filesystem without any control over where they are written. This puts me, the developer, in a position where I don't have full control over how the library operates and it makes me concerned that it may have written other files I don't know of elsewhere that I may delete (think for example of the case where I run the application inside of a sandbox or jail and every time the app is launched the jail is deleted, there needs to be a way, at minimum, for the developer to control this).

I would rather we have, if long term state is needed, that we abstract that into a separate class (maybe `MeasurementKitState`) that the user (developer) can call `.save()` on with a location to persist it on disk and then load it via `.load()` with the same file path to restore it.

bassosimone:
> That is one aspect, the other aspect is that I find it very unusual and unexpected for a library to start writing files on my filesystem without any control over where they are written.

This is not true. You can (and on mobile must) specify the path where to write explicitly.

> This puts me, the developer, in a position where I don't have full control over how the library operates

Yeah, this basically reinforces my previous guess that it's a matter of control.

> I would rather we have, if long term state is needed, that we abstract that into a separate class (maybe MeasurementKitState) that the user (developer) can call .save() on with a location to persist it on disk and then load it via .load() with the same file path to restore it.

Do you see this as an abstraction of the file system where you are prompted when the library wants to `read`, `write`, or `append` to a file with a specific name? Because, as of now, there are already a number of cases where we read or write files from the disk. Think, for example, at reports.

This is, anyway, something not very quick to implement. For the specific case of orchestration, there can be some persistent state object that is passed around. The not-so-simple part is how to make it easy to use the library considering both the case where you don't care and the one where you want control.

I will think about it. I guess I'll merge anyway what I already have prepared and then, on top of that, figure out how the programmer can be given full control about making the state persistent.

test/dns/system_resolver.cpp
bassosimone: Do I need to create an issue where to make this happen?

bassosimone:
I have now modified the orchestrator so the programmer can control whether to write on disk. I'd like to merge this branch into `master`, since this is basically what `master` should be.

bassosimone:
Merging!

PR #1259, opened 2017-06-12
Implement send and recv for socket_t

bassosimone:
Depends on #1250.


bassosimone:
I guess we don't need this. We probably need a wrapper for `send` and similar functions that returns an `Error` rather than `-1` and the possibility to combine this with `pollfd`.

PR #1260, opened 2017-06-13
orchestration: guess params and other improvements

bassosimone:


bassosimone:
While this approach is cool, I've sensed from a private discussion with @lorenzoPrimi and @hellais that what the apps really need is to explicitly query for the country and ASN and cache this information. In fact, the app is much better positioned than MK to know when such information are *likely* changed, because it can listen for specific events. As such, the `orchestrator::Client` should probably expose an interface for querying these information and return it back to the app.

bassosimone:
Things to do next:

1. update documentation

2. consider whether it makes sense to update right after register
   in the case where we weren't registered

lorenzoPrimi:
Would be possible for the app to notice a change in the CC or ASN to update the orchestrator?
In case what will be the best way to do it?
- Can you send a notification to the app?
- Or I should set a timer and check it quite often? 
- Or check it every time on app opening would be a good balance

Waiting for @hellais comment on this

bassosimone:
@lorenzoPrimi after this PR you have this operation:

```C++
client.find_location([](Error &&err, std::string &&asn, std::string &&cc) {
    if (!err) {
        // Save somewhere the `asn` and `cc`
    }
});
```

You can store the values and find again the location when you notice app-related events that lead you to suspect that the location is changed. Then, just set these cached values to the client before doing any other operation.

Otherwise, if you don't set these value, MK will always automatically try to infer them.

PR #1262, opened 2017-06-18
Simple implementation of fcompose()

bassosimone:
Subset of #1235 where we don't deal with exceptions yet.


bassosimone:
As of 86ad1ea, it seems the only broken thing is the regress test for errors. I have fixed this in 2776fa0, which means that this pull request can become merge-able tonight.

PR #1263, opened 2017-06-20
orchestrator: do not write on disk

bassosimone:


src/libmeasurement_kit/ooni/orchestrate.cpp
```diff
@@ -29,10 +29,102 @@ std::string production_events_url() {
...
+Error Auth::load(const std::string &filepath) noexcept {
+    ErrorOr<std::string> maybe_data = slurp(filepath);
+    if (!maybe_data) {
+        return maybe_data.as_error();
+    }
+    return loads(*maybe_data);
+}
+
+Error Auth::loads(const std::string &s) noexcept {
+    return json_parse_process_and_filter_errors(s, [&](auto json) {
```
hellais: What happens if some key is missing or the json document is malformed? How will the user learn of an error?

hellais: I see now from the unittests that i handles errors by returning the error.

PR #1272, opened 2017-07-06
@joelanders' captive portal PR plus my patches

bassosimone:
This is #1238 plus small tweaks that I committed on top of it.


bassosimone:
I checked the pull request for correctness. I applied small changes on top of it. These changes are such that they will ease my future work of maintaining this test. I am 90% okay to merge this.

Specific questions:

- @joelanders can you read the final diff and check whether i did not did major mistakes? (I could have asked you to do this changes, but they were mostly trivial, so I felt like it was quicker and less annoying for you if i applied directly these changes to the code and asked for a review later).

- @hellais This pull request uses a data format that is significantly different from the one used by the previous version of the captive portal. I agree with @joelanders that [having the expected and actual values is going to make the UI more interesting](https://github.com/measurement-kit/measurement-kit/pull/1238#issuecomment-308723202). Are you okay with this format change? Can we merge the corresponding spec PR (TheTorProject/ooni-spec#87) as part of merging this PR?

src/libmeasurement_kit/nettests/captive_portal.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <measurement_kit/nettests.hpp>
+#include <measurement_kit/ooni.hpp>
+
+namespace mk {
+namespace nettests {
+
+CaptivePortalTest::CaptivePortalTest() : BaseTest() {
+    runnable.reset(new CaptivePortalRunnable);
+    runnable->test_name = "captive_portal";
+    runnable->test_version = "0.0.1";
```
bassosimone: Given that we're at format `0.4` in the spec, shouldn't this be `"0.4.0"`, @joelanders?

joelanders: whoops, yes!

bassosimone: Implemented in e5af71f

src/libmeasurement_kit/ooni/captive_portal.cpp
```diff
@@ -0,0 +1,231 @@
...
+    i.clear();
+    i["name"] = "Apple HTTP 1";
+    i["url"] = "http://captive.apple.com";
+    i["body"] = "<HTML><HEAD><TITLE>Success</TITLE></HEAD><BODY>Success</"
+                "BODY></HTML>\n";
+    i["ua"] = "CaptiveNetworkSupport/1.0 wispr";
+    is.push_back(i);
+    i.clear();
+    i["name"] = "Apple HTTP 2";
+    i["url"] = "http://captive.apple.com/hotspot_detect.html";
```
hellais: What is the difference between `HTTP 1` and `HTTP 2` (a part from the fact that it's a different URL?). What I mean is, in which case is one URL used vs the other?

hellais:
Some high level comments:

* Why has the 0x20 check been dropped?

* Why has the google_dns_cp test been dropped?

* Why has the W3 Captive Portal test been dropped?

src/libmeasurement_kit/nettests/captive_portal.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include <measurement_kit/nettests.hpp>
+#include <measurement_kit/ooni.hpp>
+
+namespace mk {
+namespace nettests {
+
+CaptivePortalTest::CaptivePortalTest() : BaseTest() {
+    runnable.reset(new CaptivePortalRunnable);
+    runnable->test_name = "captive_portal";
```
hellais: According to the spec the test name is actually `captiveportal` written altogether (https://github.com/TheTorProject/ooni-spec/pull/87/files#diff-2474b11b8ed0dee978a9293a881a3c73R185)

src/libmeasurement_kit/ooni/captive_portal.cpp
```diff
@@ -0,0 +1,231 @@
...
+    mk::parallel(continuations, all_done_cb, 3);
+}
+
+// this is the only test that doesn't follow the pattern of those above.
+// this hostname should always resolve to this IP.
+void dns_msft_ncsi(Var<Entry> entry, Callback<Error> done_cb, Settings options,
+                   Var<Reactor> reactor, Var<Logger> logger) {
+    std::string hostname = "dns.msftncsi.com";
+    // Note: we're setting the nameserver to empty, which is going to work
+    // as long as we're using the `system` DNS resolver
```
hellais: Is there a way to force system DNS resolver? What happens if in the future the MK default changes?

src/libmeasurement_kit/ooni/captive_portal.cpp
```diff
@@ -0,0 +1,231 @@
...
+                                             {"failure", nullptr}};
+                             if (!!err) {
+                                 logger->info("dns_query err: %s",
+                                              err.as_ooni_error().c_str());
+                                 result["failure"] = err.as_ooni_error();
+                                 done_cb(err);
+                             } else {
+                                 for (const auto &a : message->answers) {
+                                     // XXX: This should actually be adapted
+                                     // to also deal with IPv6 replies
```
hellais: What happens now if an answer is ipv6? Will it crash of just not be parsed?

src/libmeasurement_kit/ooni/captive_portal.cpp
```diff
@@ -0,0 +1,231 @@
...
+                                     result["actual_ips"].push_back(a.ipv4);
+                                 }
+                                 if (message->answers.size() != 1) {
+                                     logger->info("maybe captive portal");
+                                     result["result"] = false;
+                                 } else if (message->answers[0].ipv4 !=
+                                            "131.107.255.255") {
+                                     logger->info("probably captive portal");
```
hellais: I see @bassosimone usually prefixes logging calls with the scope in which it's happening. I think that is quite useful especially for debugging otherwise you are lost in the logs.

bassosimone: Yep. I am mostly doing that for `debug` and `warning`, tho. With `info` the assumption is that the log message is what you, as a user, would see usually. In such case, it is probably nicer to see just what the program was doing. (It was not like this earlier, but, after using ooni-probe from command line, I started to appreciate its clean logging in `info` mode.)

src/libmeasurement_kit/ooni/captive_portal.cpp
bassosimone: As the code is currently written, each non-IPv4 entry will appear as an empty string inside of the `actual_ips` list.

src/libmeasurement_kit/ooni/captive_portal.cpp
bassosimone: To force the system resolver, `options["dns/engine"] = "system"`.

If the future default changes, the parsing of an empty endpoint will fail with `value_error`.

src/libmeasurement_kit/ooni/captive_portal.cpp
bassosimone: I don't know. I defer to @joelanders. (I asked him to document the process with which he obtained all this information regarding captive portal detection in the specification of the test.)

joelanders:
# ms tests
https://technet.microsoft.com/en-us/library/cc766017(v=ws.10).aspx
# can't find official apple documentation, but here's an example:
https://forums.developer.apple.com/thread/26867

# kitkat
http://androidxref.com/4.4_r1/xref/frameworks/base/services/java/com/android/server/ConnectivityService.java#4122

# lollipop
http://androidxref.com/5.1.1_r6/xref/frameworks/base/services/core/java/com/android/server/connectivity/NetworkMonitor.java#659

# marshmallow
http://androidxref.com/6.0.1_r10/xref/frameworks/base/services/core/java/com/android/server/connectivity/NetworkMonitor.java#643

# nougat 1
http://androidxref.com/7.1.1_r6/xref/frameworks/base/services/core/java/com/android/server/connectivity/NetworkMonitor.java#88

# nougat 2
http://androidxref.com/7.1.1_r6/xref/frameworks/base/services/core/java/com/android/server/connectivity/NetworkMonitor.java#91

# nougat 3
https://android.googlesource.com/platform/frameworks/base/+/master/services/core/java/com/android/server/connectivity/NetworkMonitor.java#99

Weirdly couldn't find that last one on the androidxref site... Also I should specify these "Google" ones are actually specific to Android and there seem to be others for Chrome, Google TV, etc.

I'll add these to a comment at the top of the test.

bassosimone:
> I'll add these to a comment at the top of the test.

I think it's better to add them to the spec, no?

joelanders:
Ok, keeping the old report format and making the test name match the spec. I'm not changing anything for the default DNS resolver or trying to record ipv6 DNS entries, is that right?

bassosimone:
I have started merging this. I will push couple of more commits tomorrow and then merge, unless testing it and comparing with ooniprobe output I find any blocker.

bassosimone:
I have applied some more changes. In particular, I make the vector `const static` because there was actually no point in copying it back and forth. It is basically a constant artifact.

Even though everything looks good to me, I will not merge the test for now.

Before merging, I'd like to know the following (I remember we discussed this, but cannot find records, so I am not sure what was the conclusion that we reached, and think it's useful to write it down here):

1. @joelanders: I see that you did not implement the W3 captive portal test, why?

2. @joelanders: ooniprobe also does a check with "http://www.apple.com/library/test/success.html" that is _different_ from the one we do here. I guess the one we do here is a more updated version of the Apple test, correct? Why don't we implement also the old one?

3. @joelanders: I see that ooniprobe does a test called `google_dns_cp`. Why did we drop that?

4. @joelanders: I've seen that above @hellais mentioned a test called `0x20` DNS and &mdash; even though I don't understand where it is in the output of ooniprobe &mdash; I guess we should probably also implement that, unless we have a clear reason not to do so. A possible reason not to do so is that all the DNS engines we have aren't low-level enough for us to really understand what happens. Correct?

Additionally, but not related to merging:

4. @hellais: here we do much more HTTP tests than in ooniprobe. I guess there is no point in porting back to ooniprobe such tests, since MK will soon become the engine of ooniprobe, correct?

5. @joelanders has the companion `ooni-spec` PR been updated?

joelanders:
1. I don't see that test documented anywhere, and it doesn't actually work in ooniprobe. (checking for a wrong status code or something)
2. I based this code on what I saw documented around the Internet currently.
3. From what I've read, that was more to detect if you have a DNS server that's responding to non-existent domains (like for typosquatting purposes).
4. That also wasn't captive portal related- there's some cache-poisoning attack you can do to a DNS server where an off-path attacker races the authoritative response. Seems totally unrelated to captive portals, and to client-side detection in general.
5. I guess I should flesh out the spec more (https://github.com/TheTorProject/ooni-spec/pull/87)

hellais:
> 1. I don't see that test documented anywhere, and it doesn't actually work in ooniprobe. (checking for a wrong status code or something)

It's documented inside of the header of the ooniprobe test and inside of the document that we actually fetch to perform the test: https://tools.ietf.org/html/draft-nottingham-http-portal-02.

Why do you say it doesn't actually work in ooniprobe? Do you mean that real captive portals don't actually implement it?

I think it's still useful to have this test be done as it actually is indicative of whether or not the captive portal in questions properly implements the linked RFC.

> 2. I based this code on what I saw documented around the Internet currently.

Do you have a reference for this?
This is a reference for the URL used by ooniprobe (also linked in the header of the captive portal test): https://bugzilla.mozilla.org/show_bug.cgi?id=562917#c16

> 3. From what I've read, that was more to detect if you have a DNS server that's responding to non-existent domains (like for typosquatting purposes).

It's also very common for captive portals to resolve any DNS query (also for non-existent domains) with the IP of the login page. As such this is a useful thing to test for too.

> 4. That also wasn't captive portal related- there's some cache-poisoning attack you can do to a DNS server where an off-path attacker races the authoritative response. Seems totally unrelated to captive portals, and to client-side detection in general.

That is the purpose of `0x20`, however it allows you to detect cases in which the DNS query can be forged by the captive portal.

hellais:
> @hellais: here we do much more HTTP tests than in ooniprobe. I guess there is no point in porting back to ooniprobe such tests, since MK will soon become the engine of ooniprobe, correct?

yeah I would not do that.

hellais:
I would say that it would be ideal to have at least 1 (if not 2) DNS based tests as those are very important features of a captive portal.

The main thing I would like to have answered is:

* Are DNS resolutions getting intercepted and forged by the captive portal? (the `0x20` + non-existent domain resolution do to some extent answer that question in some cases).

joelanders:
1. The ooni-probe test is checking for some error code in a response to the RFC describing the error code. The logic is backwards; its counted as unfiltered if you get a 428.

2. I put some of my refs up in this thread. I should put them in the spec when I write it.

3 + 4. It's weird to intentionally misapply methodology, I think.

joelanders:
https://github.com/TheTorProject/ooni-spec/pull/87/files
Refernces added to spec PR^

joelanders:
I guess the random-hostname DNS lookups tell us something about the network. Think it's more of a DNS Consistency test than a captive portal one, though. But we're phasing that out in favor of Web Connectivity, right?

joelanders:
Ok, I will add the random domain lookup stuff.

joelanders:
Random domain stuff is on this branch: https://github.com/joelanders/measurement-kit/tree/feature/captiveportal

bassosimone:
> Random domain stuff is on this branch: https://github.com/joelanders/measurement-kit/tree/feature/captiveportal

@joelanders: okay, I'll try to pull changes from there.

bassosimone:
@joelanders: okay, I've pulled and merged changes and now I'm dealing with further post-merge fallback after which I will probably merge, unless I see issues in additional changes.

bassosimone:
# TODO list for me
- [ ] pull @joelanders changes and merge with master
- [ ] fix all the conflicts
- [ ] review the whole diff
- [ ] make sure the report file is still okay
- [ ] make sure readme is up to date
- [ ] merge

bassosimone:
Unless travis fails, I am going to merge this! 🎉

bassosimone:
Merging anyway, even with coverage reduced by more than `1%`. I will try to further increase the coverage later as part of the release cycle, by focusing initially on stuff like `common` and `net`.

PR #1273, opened 2017-07-06
Allow to pass nettests input from command line

### LINKED ISSUE #308: Tests should run with a string as input

bassosimone:
Supersedes #1132.

Closes #308.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -10,9 +10,9 @@ AC_DEFUN([MK_AM_ENABLE_COVERAGE], [
 
 AC_DEFUN([MK_AM_ADD_COVERAGE_FLAGS_IF_NEEDED], [
   if test "$enable_coverage" = "yes"; then
-    CFLAGS="--coverage $CFLAGS"
-    CXXFLAGS="--coverage $CXXFLAGS"
-    LDFLAGS="--coverage $LDFLAGS"
+    CFLAGS="$CFLAGS --coverage -g -O0"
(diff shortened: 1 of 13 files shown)


bassosimone:
As soon as travis completes, I will merge this into the integration branch. This is basically a simplified #1273 where input passed manually is randomized like input read from files. This [was requested by @hellais in #1273](https://github.com/measurement-kit/measurement-kit/pull/1132#issuecomment-280669703).

bassosimone:
Uhm, it seems that travis fails running coveralls in a very weird way. Maybe better to look into that before proceeding with the actual merging.

bassosimone:
Okay, travis issues should be fixed in dac500e. I tested them locally. Merging into the integration branch as soon as travis completes successfully. (While there, I also fixed coverage measurements and moved compiler flags &mdash; issues noticed while trying to understand why travis failed.)

bassosimone:
Okay, now it seems that [the build fails](https://travis-ci.org/measurement-kit/measurement-kit/builds/251080383) because `hushmail.com` has changed the way in which it serves results. Plus, we have a transient issue with ooni resources. Will also fix those as part of this PR.

bassosimone:
The travis runs that I did not suppress have passed. As such, I am going to merge.

PR #1271, opened 2017-07-04
Integration branch for v0.7.0-beta

bassosimone:
Integration branch for v0.7.0-beta.

PR #1275, opened 2017-07-14
docker: use libc++ with clang

bassosimone:
As explained in c6303b8, to have a robust MK on Linux, we'd better use libc++.

This PR also includes other fixes to repair failing regress tests that are part of the history of #1271.


bassosimone:
(The coverage in here is low because fixes to improve it are inside #1271 that has not been merged yet.)

bassosimone:
I have moved cherry-picks from the integration branch to #1278, so that merging pull requests like this one creates less noise and, additionally, #1271 is easier to review.

bassosimone:
It is also worth mentioning that I needed to upgrade to `zesty` because with `yakkety` the `libc++` based build was failing. We should probably update all automatic builds to use `zesty`.

bassosimone:
(I suppressed the `travis-ci/pr` build.

bassosimone:
Merging because the builds I did not suppress have passed.

Issue #1274, opened 2017-07-12
web_connectivity: failure due to TLS issues


bassosimone:
There are cases where tests fail not because of censorship but because our TLS engine coupled with our embedded CA (both provided by LibreSSL) somehow fail to recognize valid certificates.

I have started experimenting to improve the situation. I am currently testing with:

- upgrading to LibreSSL v2.5.5

- using [cURL cacert](https://curl.haxx.se/docs/caextract.html)

I am not sure this will be able to land into v0.7.x, because there are some cross-compilation issues to be solved both on iOS and Android. Otherwise, this will most likely be in v0.8.x.

bassosimone:
I think #1216 is possibly related to this bug.

bassosimone:
So, this has now landed in v0.7.x integration branch and testing with Android and iOS showed that there are much less case where we fail because of HTTPS. More specifically, the only case where I did see a failure was indeed the case where the certificate was expired and also cURL was failing likewise.

PR #1279, opened 2017-07-14
Fix: ssl verify errors

bassosimone:
This PR fixes ssl verify errors reported by several users and developers (see #1274), where the test failed because it was not possible to authenticate it, despite the certificate being good.

To fix this issue, I have:

1. switched to use [the same CA bundle used by cURL](https://curl.haxx.se/docs/caextract.html)

2. upgraded to libressl v2.5.5 that correctly handles such CA (with 2.4.5 we had issues where some hosts like `mlab-ns.appspot.com` where not correctly authenticated)

Upgrading to libressl v2.5.5 also forced me to update the related set of patches (actually it was not a difficult job, luckily). Also, I've removed a previously used flag that prevented using assembly on ARM, because now it seems the compile process goes smoothly even when using assembly.

I am now testing this quite intensively with many web_connectivity tests on Android and iOS.


bassosimone:
The build of 9c355c2 was previously stopped. I am now restarting it. I will merge into the integration branch for v0.7.0-beta as soon as the automatic checks terminate. The subsequent b12f0f1 has no impact on automatic tests, and hence is `[ci skip]`.

bassosimone:
(Honestly, I don't think it makes much sense to wait for Travis to complete here. Basically, these are changes that are not tested at all by Travis. this makes me wonder whether we perhaps want also to enable an additional layer of checks where we cross-compile stuff etc.)

PR #1294, opened 2017-08-13
fix(var.hpp): correct dupe in license comment

bassosimone:


bassosimone:
This can be merged without waiting for checks, since it only changes a comment.

PR #1297, opened 2017-08-14
Better life-cycle model for long-running tasks

bassosimone:
Motivation
==========

I recently had a discussion with @hellais where we agreed that writing code for measurement-kit is difficult.

What makes it difficult, are the rules to guarantee that we are deadling with object's life cycle properly.

Current rules
=============

We basically have two cases: the case where state is not needed and the case where state is needed. But there is also code that is not following the current rules.

No state needed
---------------

The former case is the one where you have a one-shot slow operation to which you pass arguments and a callback.

Consider for example DNS:

```C++
mk::dns::query("IN", "A", "www.google.com",
               [](mk::Error error, mk::dns::Message message) {
               // HANDLE ERROR AND PROCESS MESSAGE
           });
```

In this _simple_ case you really don't want and need any more state. This code will create state when you initiate the DNS query and destroy it after the callback.

State needed
------------

There are cases where you cannot get away with a simple function taking arguments and a callback. Those are the cases where you need an object that will emit events throghout its lifecycle. An example of that could be the transport:

```C++
using namespace mk;
using namespace mk::net;

void some_func(Var<Transport> txp, Var<Buffer> buff, ...) {
    read(txp, buff, [](Error error) {
        // HANDLE ERROR OR READ DATA
    });
}
```

In these cases the approach has been to allocate state using the `Var<>` shared pointer and then to have function operations to which you can pass such state. While operating on state, these operations will also keep it alive, by virtue of `Var<>` being a shared pointer. While this seems neat, it is a bit obscure in that you don't actually _see_ the moment where the specific bit of code will be destroyed. And having it die when the last callback uses it _may not_ be optimal, so that you need to use the `call_soon()` trick described below.

Code not following rules
------------------------

An example of code not following rules is `nettests::Runnable` where we do pass `this` to some lambdas.

This is dangerous because it does not guarantee that the object will still be alive when the callback is called, this possibly leading to _use after free_ errors.

However, for `nettests::Runnable` overall the code is safe because we _know_ that the `Runnable` is managed by a `Runner` that receives the `Runnable` as a shared pointer and properly pass such shared pointer around to some lambdas such that it stays alive until it reaches its final state and calls its final callback.

When the final-callback state is reached, the code is careful not to lose the reference to the shared pointer immediately, which may trigger destruction with a stack shaped like this:

```
    +---------------+
    | test::final() |
    +---------------+
    |   test::a()   |
    +---------------+
    | libevent code |
    +---------------+
```

Assuming we're in `test::final()`, calling the final callback and losing the shared pointer reference is dangerous in that this _might_ cause destruction and, hence, when we return to `test::a()`, we can use the object after it has been freed.

The solution of this problem is to use `Reactor::call_soon()` such that, when we call the final callback, the stack is instead shaped like this:

```
    +---------------+
    | test::final() |
    +---------------+
    | libevent code |
    +---------------+

```

So, we just call the final callback, trigger destruction, but since we have only libevent code below on the stack, this means we cannot have a use-after-free condition.

Proposal
========

There is one use case where writing code for measurement-kit is actually simpler and less rules are involved.

That is the case where you write a simple main, allocate on the stack all the objects you need, start the reactor and, in its "initial event" callback, start all the async operations (be them I/O, threads, or delayed calls) you need. This includes setting a final state handler, which calls `reactor->stop()`. E.g.:

```C++
using namespace mk;
Var<Reactor> reactor = Reactor::make();
// Note: nonexistent API here only to make a point
dns::Client client{reactor};
client.set_dns_server("8.8.8.8", 53);
reactor->run_with_initial_event([&]() {
    client.query("IN", "A", "www.google.com",
                 [](Error error, dns::Message message) {
                     // DO WHAT YOU NEED
                     reactor->stop();
                 });
});
```

This is very neat for the following reasons:

- all you need is allocated beforehand on the stack and it will be de-allocated when you leave the scope

- `run_with_initial_event()` is blocking, so you can safely pass all the state using `&` and you don't need to keep it alive

- when you reach the final state, you stop the reactor, and so you basically unblock `run_with_initial_event()` and you leave the scope and cleanup everything

However, currently it's not possible to do that with async code, since you don't find yourself in a blocking context where you can use RAII like above. That's the reason why we came up with rules involving shared pointers and passing them around to make sure we keep them safe until the final state.

This pull request aims to:

1. make handling big async operations (like running a test or orchestration) as easy as it would be run them in a sync fashion using a simple main

2. make sure that, even if `reactor->stop()` is not called because of a bug, the reactor is going to stop automatically when it has no further work do do

Technically, change 2 could have been implemented alone, but I tried that in #1293, only to discover that I needed to do other changes to make it work. So, overall, it would have been more complex to write two diffs than just a single one.

Implementation
==============

To achieve the above objectives, basically, what we do is that we require a specific reactor to be used for running big async operations.

We will run this reactor in a specific thread, and in a condition in which it's basically possible to replicate what you would do when writing a simple main. So, you can adopt more relaxed object rules (objective 1).

Additionally, we will change the way in which the reactor loops so that it leaves the loop when it is out of things to do (objective 2).

For example, we will be able to write:

```C++
/*
 * Hypothetical snippet from API for performing a single DNS query. Unlikely
 * we will implement something like that because it's a waste to use a reactor
 * only to perform this operation (at least currently I think so). Here only
 * to explain my point. More general than example above, because we need to
 * also show what it would entail to move around initial params.
 */
using namespace mk;
// Note that Worker is an _internal_ API
Worker::default_tasks_queue()->run_in_background_thread([
    dns_server_address = std::move(dns_server_address),
    dns_server_port,
    dns_query_class = std::move(dns_query_class),
    dns_query_type = std::move(dns_query_type),
    dns_query = std::move(dns_query)
]() {
    Var<Reactor> reactor = Reactor::make();
    // Note: nonexistent API here only to make a point
    dns::Client client{reactor};
    client.set_dns_server(dns_server_address, dns_server_port);
    reactor->run_with_initial_event([&]() {
        client.query(dns_query_class, dns_query_type, dns_query,
                     [](Error error, dns::Message message) {
                         // DO WHAT YOU NEED - MAYBE CALL A CALLBACK?
                         reactor->stop();
                     });
    });
});
```

Limitations in applicability
============================

This enhancement basically binds an operation with a reactor.

While this makes sense for big operations like tests, orchestration, or downloading resources, it would not make sense, of course, for smaller operations like a single DNS resolution or HTTP request.

So, the external API of smaller operations cannot be changed as a result of this pull request.

However, the internal API can. For example, we can write an internal DNS API where we have a DNS client object that can be either used to implement the external DNS API or can be used by a specific test so that we do not need to allocate and deallocate a DNS client for each request we perform, rather we can reuse the same client, allocated using RAII as a sub-object of a nettest.

And, when we are performing requests with such DNS client in the context of the larger DNS object, we can pass `this` to lambdas and generally be relaxed that the overall object is alive until end of test.

So, to summarize, this has an impact on internal implementation more than on external APIs. For external APIs, another solution should probably be found to achieve the same objective.

One solution I was thinking of was to allow to allocate objects in the context of the reactor, such that they are alive as long as the reactor loop is running and cancelled afterwards.

This is anyway something I will do in another pull request.

Discussion
==========

This introduces the additional constraint that it binds the lifecycle of a reactor with the lifecycle of a big task or test.

This has more overhead than the current model where we use a single reactor loop to serve all the I/O in measurement-kit.

Yet, since measurement-kit is a measurement library, I don't see a use case where we're running many tasks in parallel.

Rather, I'd argue we want, in most cases, to run 1-2 tasks in parallel.

As such, for us nonblocking I/O is not a matter of scalability but rather a way to perform many async operations using a single thread.

So, I don't see binding a reactor to a specific operation as a disadvantage given the purpose of the library.

Moreover, I see the following advantages:

- with small code changes, it will become very easy to interrupt a running test, because it will just suffice to invoke `reactor->stop()` from another thread context

- scheduling of big tasks or tests is already easier because we have a queue from which we pick and what cannot be served currently will stay in the queue until it's possible to serve it

- writing internal code will become much simpler: it if works using a simple main() it will also work in async fashion

- currently, we have an always-running thread for the reactor, while after this diff we will only have reactors running during a test or long task, which is perhaps more suitable to the requirements of mobile devices


bassosimone:
Tests are failing. But looking at the reason of failure, I don't see any dramatic issues. So, I decided to open the pull request and deal with fixing errors with the pull request open.

test/nettests/dash.cpp
```diff
@@ -15,8 +15,4 @@ TEST_CASE("Synchronous dash test") {
     test::nettests::make_test<DashTest>().run();
 }
 
-TEST_CASE("Asynchronous dash test") {
-    test::nettests::run_async(test::nettests::make_test<DashTest>());
-}
-
```
bassosimone: I have removed async execution of tests since now we use the same code for async and non-async execution. There may be a little penalty for doing sync execution in another thread, but our main use case is _really_ the case of async execution in a background thread. As such, this does not worry me much that we're implementing sync execution using async execution plus a future object. Removing async execution for each test, given that it's now redundant, can significantly reduce the time it takes to run all tests on travis, so I see it as a plus.

test/nettests/runner.cpp
```diff
@@ -1,89 +0,0 @@
-// Part of measurement-kit <https://measurement-kit.github.io/>.
```
bassosimone: Interface gone, so also the test has been removed.

bassosimone:
# Tests that are failing and explanation of why they are failing

Let's take [the case where more tests fail](https://travis-ci.org/measurement-kit/measurement-kit/jobs/264352050). We see three failures:

## test/nettests/runnable

```
FAIL: test/nettests/runnable
============================
==19725== Memcheck, a memory error detector
==19725== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==19725== Using Valgrind-3.12.0.SVN and LibVEX; rerun with -h for copyright info
==19725== Command: ./test/nettests/runnable
==19725== 
[snip]

test/nettests/runnable.cpp:229
test/nettests/runnable.cpp:230: FAILED:
with expansion:
  1 == 9
test cases:  8 |  7 passed | 1 failed
assertions: 31 | 30 passed | 1 failed
==19725== 
[snip]
```

So, this case here is that somehow we've broken a test case. *Action item*: investigate.

bassosimone:
# Update regarding failing tests

I was _hoping_ to be able to fix all tests by this evening, but apparently I could not.

The only one that is _still_ failing is `./test/nettests/runnable`. This is weird in that this piece of code _should not_ be affected by that problem and in that I cannot easily reproduce it locally.

First things first. After the changes in this diff are applied, the [failing test](https://github.com/measurement-kit/measurement-kit/blob/5fe25a8c6d91154ba358261041c516a0be946bcb/test/nettests/runnable.cpp#L230) &mdash; a test that verifies whether we can _disable_ shuffling input &mdash; [fails like this on travis](https://travis-ci.org/measurement-kit/measurement-kit/jobs/265213193):

```
test/nettests/runnable.cpp:230: FAILED:
with expansion:
  1 == 9
```

Interestingly, it seems that it fails _only_ when running through Valgrind. In fact, the same test passes for me both on my macOS laptop and on my Ubuntu dev box. However, when running it through valgrind (as in `./build/docker/valgrind-clang`) the test actually do fails like this:

```
test/nettests/runnable.cpp:230
test/nettests/runnable.cpp:231: FAILED:
with expansion:
  3 == 9
```

I am bit puzzled in that I do not _fully_ understand what is happening. I think it is something not obvious to me and out of my current mental model of the change. The fact that it *does not* happen when not using Valgrind seems to confuse me further 😺.

I'll try to do another pass at it this night and hopefully I'll understand what is going on.

bassosimone:
Another datapoint: I _cannot_ reproduce the problem by compiling and running with and without Valgrind on Ubuntu zesty. But the problem appears when running inside docker on Ubuntu zesty. 🍭

bassosimone:
A-ha, I fixed it. It was due to parallelism in running tests that somewhat shuffled the results. This is the super simple diff that fixes the issue:

```
diff --git a/test/nettests/runnable.cpp b/test/nettests/runnable.cpp
index 099cd962d..6e6d887f6 100644
--- a/test/nettests/runnable.cpp
+++ b/test/nettests/runnable.cpp
@@ -192,6 +192,7 @@ TEST_CASE("Make sure that 'randomize_input' works") {
         test.reactor = Reactor::make();
         test.input_filepaths.push_back("./test/fixtures/hosts.txt");
         test.options["randomize_input"] = shuffle;
+        test.options["parallelism"] = 1; // Otherwise they're mixed anyway
         test.needs_input = true;
 
         test.reactor->loop_with_initial_event([&]() {
```

I _think_ that some race conditions are more obvious when running under Valgrind. The test has been probably broken in a very subtle way since the beginning, and after this change that was exposed.

With this fixed, now I only have to cleanup the diff and then this can _really_ be ready for review. Yay!

bassosimone:
Okay, I should have fixed the build by now. This should now be ready for review, @hellais!

include/private/common/async_runner.hpp
```diff
@@ -1,133 +0,0 @@
-// Part of measurement-kit <https://measurement-kit.github.io/>.
```
bassosimone: File removed because this is superseded by running using `Worker::default_tasks_queue()`.

include/private/common/worker.hpp
```diff
@@ -20,7 +20,7 @@
 
 namespace mk {
 
-class Worker : public HasGlobalFactory<Worker> {
+class Worker {
```
bassosimone: Removed because we use now a specifically named factory.

include/private/nettests/runner.hpp
```diff
@@ -1,27 +0,0 @@
-// Part of measurement-kit <https://measurement-kit.github.io/>.
```
bassosimone: Removed because superseded by using `Worker::default_tasks_queue()`.

include/private/ooni/orchestrate_impl.hpp
```diff
@@ -4,12 +4,6 @@
 #ifndef PRIVATE_OONI_ORCHESTRATE_IMPL_HPP
 #define PRIVATE_OONI_ORCHESTRATE_IMPL_HPP
 
-#include "private/common/async_runner.hpp"
-#include "private/common/fcompose.hpp"
-#include "private/common/json.hpp"
-#include "private/common/mock.hpp"
-
-#include "private/common/async_runner.hpp"
```
bassosimone: Remove duplicate headers that were probably created by an automatic merge.

src/libmeasurement_kit/nettests/base_test.cpp
```diff
@@ -98,25 +99,59 @@ BaseTest &BaseTest::on_destroy(Delegate<> cb) {
     return *this;
 }
 
+static void start_internal_(Var<Runnable> &&r, std::promise<void> *promise,
+                            Callback<> &&callback) {
```
bassosimone: The core function now used to run tests. The detailed comment should explain all, er, details.

src/libmeasurement_kit/nettests/base_test.cpp
```diff
@@ -98,25 +99,59 @@ BaseTest &BaseTest::on_destroy(Delegate<> cb) {
...
-                reactor->call_soon([&]() {
-                    reactor->stop();
-                });
-            });
-        });
-    });
+    std::promise<void> promise;
+    std::future<void> future = promise.get_future();
+    // Note: using `std::move` to invalidate the `runnable` such that it is
+    // not possible to start another test from the same object.
```
bassosimone: Before that was not done explicitly, but I always meant for that to be the case.

src/libmeasurement_kit/nettests/runner.cpp
```diff
@@ -1,41 +0,0 @@
-// Part of measurement-kit <https://measurement-kit.github.io/>.
```
bassosimone: Removed because superseded by `Worker::default_tasks_queue()`.

src/libmeasurement_kit/ooni/orchestrate.cpp
```diff
@@ -157,55 +158,62 @@ void Client::register_probe(std::string &&password,
...
-              [ done = std::move(done),
-                cb = std::move(cb) ](Error && error, Auth && auth) mutable {
-                  done([
-                      error = std::move(error), cb = std::move(cb),
-                      auth = std::move(auth)
-                  ]() mutable { cb(std::move(error), std::move(auth)); });
-              });
+    ]() mutable {
+        Var<Reactor> reactor = Reactor::make();
+        reactor->run_with_initial_event([&]() {
```
bassosimone: This is basically the same as we do for running tests. Here in orchestration it has not been factored in a common function because in here it is actually a couple of lines of code, so I didn't see the need.

test/libevent/reactor.cpp
```diff
@@ -81,35 +81,6 @@ TEST_CASE("Reactor: basic functionality") {
...
-        libevent::Reactor<event_base_new, event_base_once,
-                          event_base_dispatch_no_events, event_base_loopbreak,
-                          event_new, event_add_fail>
-              reactor;
-        REQUIRE_THROWS(reactor.run());
-    }
-}
-
-extern "C" {
-
```
bassosimone: Code removed because we don't add the periodic event anymore, so of course this test would fail.

test/nettests/utils.hpp
```diff
@@ -34,14 +34,6 @@ template <typename T> mk::nettests::BaseTest make_test(std::string s) {
...
 }
 
-static inline void run_async(mk::nettests::BaseTest test) {
-    volatile bool done = false;
-    test.start([&]() { done = true; });
-    do {
-        std::this_thread::sleep_for(std::chrono::seconds(1));
-    } while (!done);
-}
-
```
bassosimone: No need to keep this abstraction since we've stopped running all tests both sync and async (see above comment)

hellais:
For sake of history I will summarise my comments and feedback on slack:

I raised the question wether or not goal 2 (automatically stopping the reactor) is desirable. 
Based on discussion with @bassosimone these are the conclusions:

* Since we a measurement library, high concurrency is not a goal, but rather we favour accuracy (we generally want to run tests sequentially).
* This solution will lead to less bugs and make it easier to write tests
* In the example above (Implementation section), you can also drop the `reactor.stop()` call and it will work nonetheless.
* For long running tasks (like long polling or websockets) we either may not need to do anything (because the reactor is waiting on reading on the socket) or you can use a queue.
* The goal of this design is that we should be enforcing a one reactor per test rule
* This is actually the default behaviour of libevent

I think I am satisfied with these responses.

I took a look at the diff and for me this is good to be merged.

I would maybe also file a ticket to update the developer documentation to mention this design change and look for code snippets that are using the old pattern and update them to the new one (these should all be changed in a single commit that can be useful as reference to tell people how to adjust code from one version to the other).

bassosimone:
> I would maybe also file a ticket to update the developer documentation to mention this design change and look for code snippets that are using the old pattern and update them to the new one (these should all be changed in a single commit that can be useful as reference to tell people how to adjust code from one version to the other).

I think I can do that as part of this PR to avoid tickets overhead!

bassosimone:
@hellais wrote:
> This is actually the default behaviour of libevent

Yep! For the records, I checked `libuv` and `boost::asio` and also for them by default the "reactor" automatically exits when it is out of events to wait for and/or dispatch.

bassosimone:
I am going to merge.

PR #1299, opened 2017-08-14
feature(cmake): also build examples

bassosimone:


bassosimone:
Pointless to run regress tests, since they don't use cmake. Tested locally.

PR #1300, opened 2017-08-14
rename(libevent): poller -> reactor

bassosimone:
Use less confusing name. That class implements reactor, so
call it reactor in the mk::libevent namespace.


bassosimone:
Build succeeded. Then, I re-merged master and suppressed the build (I _know_ that merging was compatible, so we can safely suppress the build: the differing commit is #1301 where I just silenced a compiler warning in the worker module).

PR #1301, opened 2017-08-16
fix(t/c/worker.cpp): silence compiler warning

bassosimone:
Now concurrency is unsigned, so checking for it being greater
or equal than zero is clearly redundant.


bassosimone:
Apart from [the not-so-infrequent failure with downloading resources](https://travis-ci.org/measurement-kit/measurement-kit/jobs/265103648#L2565), the build succeeded. Merging!

PR #1302, opened 2017-08-16
cleanup(t/c/worker.cpp): remove useless cast and comment

bassosimone:
Now that attribute is unsigned, so no need to cast to unsigned
as well as no need to justify the reason for the cast.


bassosimone:
No need to wait for travis to complete after updating master since master contains unrelated small changes, so I'll proceed with merging right away: the only build I did not suppress, terminated successfully.

PR #1303, opened 2017-08-16
cleanup: remove all CI systems except travis

bassosimone:
In the past I added other CI systems when travis was not working
properly for us. Now it is working properly for us.

Having two additional build systems with their occasional failures
to track is becoming more an hassle than an advantage.

As such, nuke everything except travis.

While there, attempt to adjust coverall's badge.


bassosimone:
Okay, [there is just a failure in fetching resources](https://travis-ci.org/measurement-kit/measurement-kit/jobs/265174096), which is harmless and happens from time to time. Merging.

PR #1304, opened 2017-08-16
fix(test/nettests): avoid races and optimize duration

bassosimone:
1. in `test/nettests/runnable.cpp` avoid the races that were so
   annoying when I was finishing #1297 up

2. in `test/nettests/utils.hpp` avoid doing collect for tests for
   which we really do not care about doing any collect


bassosimone:
Merging:

- all the builds I did not suppress completed successfully

- no need to build after upcoming merge with master that has only changed README.md

PR #1305, opened 2017-08-16
fix(README.md): use shields.io images

bassosimone:


bassosimone:
Squash this directly: it only changes badges and I've verified they work.

PR #1306, opened 2017-08-16
t/n/runnable.cpp: clang-format -i

bassosimone:


bassosimone:
This can go in without continuous integration, as it's just a `clang-format -i` of a file.

PR #1322, opened 2017-09-02
dash: detect middleboxes preventing the test from running

bassosimone:
This branch contains fixes to let dash work on Vodafone and other mobile networks where the connection is closed after the first request has been sent and the first response received.


bassosimone:
It seems that setting `Cache-Control` like I did in e5f1293 is not enough to fix the behavior.

bassosimone:
[RFC 7230 Sect 6.1](https://tools.ietf.org/html/rfc7230#section-6.1) says:

> the Connection header field provides a declarative way of distinguishing header fields that are only intended for the immediate recipient ("hop-by-hop") from those fields that are intended for all recipients on the chain ("end-to-end")

if I am understanding that correctly (and I think I do) then there is not much to fix dash for networks like Vodafone where there is a middlebox that forces `Connection: close` after the first message.
.
What I can do is to add code to _detect_ that case and fail more gracefully.

I _think_ that using `https` for DASH may be better because in that case the test would go over an encrypted end to end tunnel, so probably by restructuring it that way we would have the possibility of running the whole test.

Another possible fix would be to allow DASH to work with non-persistent connections, but also that IMHO cannot be done without thinking quite a little bit about that, since the idea of the test was to measure performance of retrieving stuff using DASH with a single connection.

bassosimone:
Added @hellais, @darkk, and @joelanders as potential reviewers, mainly to have their feedback related [to my previous comment on how to move forward](https://github.com/measurement-kit/measurement-kit/pull/1322#issuecomment-330179410).

bassosimone:
Also moving this to `v0.8.0-beta` since I think I cannot have that done by today.

bassosimone:
# TODO list for me
- [ ] in this branch, push support for preventing caches from caching responses (will not fix the main issue but is anyway the right thing to do in terms of http)
- [ ] in this branch, add a specific error code for this error condition (a-ha, the libc++ terms!) such that we can report to users what objectively is a functional limitation of the test
- [ ] open another issue for discussing how to fix dash; I currently see two different ways forward:
    1. adding to the neubot-server support for https and run the test over that
    2. change dash to work even if the connection is closed (I think this can be possibly implemented quite easily, but I wonder what effects this could have on measurements, so I should probably run some comparative tests and spend some time thinking _before_ merging this yolo into master)

bassosimone:
I am going to merge this into master as soon as travis completes, so we can test on real devices.

bassosimone:
(Added the hotfix flag, since this clearly qualifies as an hotfix.)

bassosimone:
The [travis-ci/push test that failed](https://travis-ci.org/measurement-kit/measurement-kit/jobs/295337490#L2011) is just a transient failure that we can safely ignore.

bassosimone:
Merging! I explained above because the only test that failed was a transient failure.

PR #1332, opened 2017-09-06
@joelanders #1192 plus my changes

bassosimone:


bassosimone:
Note: I've cancelled the `travis-ci/pr` build.

bassosimone:
I am going to merge this PR even though the regress tests fail. What is actually going on is that a test for `sha256.badssl.com` in `test/net/connect.cpp` fails because we're using the CA certificate chain saved at `test/fixtures/certs.pem`. I have verified locally that thing would work when using system-wide CA. That is, the failure is caused by an unrelated bug in the regress tests with which I'll deal soon.

PR #1333, opened 2017-09-06
ooni: add support for CNAME

### LINKED ISSUE #1270: dns: where possible, gather CNAME
@darkk noticed in https://github.com/TheTorProject/ooni-pipeline/pull/62 that:

>  control_failure [ambiguous] is stamped at ~27% of dns_a rows, that's likely a bug

It seems this is only affecting measurements from the mobile apps, so it's likely a bug in measurement-kit.

We should look into this fact and figure out what is going on.

bassosimone:
Closes #1270.

- `./measurement_kit facebook_messenger`

```JSON
        "queries": [
            {
                "answers": [
                    {
                        "answer_type": "CNAME",
                        "hostname": "z-m.c10r.facebook.com",
                        "ttl": 0
                    },
                    {
                        "answer_type": "A",
                        "ipv4": "31.13.86.37",
                        "ttl": 0
                    }
                ],
                "engine": "system",
                "failure": null,
                "hostname": "b-graph.facebook.com",
                "query_type": "A",
                "resolver_hostname": null,
                "resolver_port": null
            }
        ]
```

- `./measurement_kit web_connectivity -u http://www.facebook.com`

```JSON
        "queries": [
            {
                "answers": [
                    {
                        "answer_type": "CNAME",
                        "hostname": "star-mini.c10r.facebook.com",
                        "ttl": 0
                    },
                    {
                        "answer_type": "A",
                        "ipv4": "31.13.86.36",
                        "ttl": 0
                    }
                ],
                "engine": "system",
                "failure": null,
                "hostname": "www.facebook.com",
                "query_type": "A",
                "resolver_hostname": null,
                "resolver_port": null
            }
        ]
```

(Are there any other OONI tests for which this is relevant?)


src/libmeasurement_kit/ooni/web_connectivity.cpp
```diff
@@ -527,8 +527,6 @@ static void experiment_dns_query(
             for (auto answer : message->answers) {
                 if (answer.ipv4 != "") {
                     addresses.push_back(answer.ipv4);
-                } else if (answer.hostname != "") {
-                    addresses.push_back(answer.ipv4);
```
bassosimone: Note: this caused passing the `CNAME` to the backend to connect to. The result was a failure.

include/private/dns/system_resolver.hpp
```diff
@@ -45,20 +45,32 @@ void system_resolver(QueryClass dns_class, QueryType dns_type, std::string name,
         return;
     }
 
+    /*
+     * When running OONI tests, we're interested to know not only the IPs
+     * associated with a specific name, but also to the CNAME.
+     */
+    ErrorOr<bool> also_cname = settings.get("dns/resolve_also_cname", false);
```
hellais: I am not a fan of this settings name, but I understand why you used it (`resolve_cname` would have been confusing). No concrete suggestion here, but mainly a note to myself.

include/private/libevent/dns.hpp
```diff
@@ -335,6 +335,22 @@ void query(QueryClass dns_class, QueryType dns_type, std::string name,
...
+     *
+     * This is not yet implemented in the libevent engine.
+     */
+    ErrorOr<bool> also_cname = settings.get("dns/resolve_also_cname", false);
+    if (!also_cname) {
+        cb(also_cname.as_error(), nullptr);
+        return;
+    }
+    if (*also_cname == true) {
+        cb(mk::NotImplementedError(), nullptr);
```
hellais: ha, so libevent doesn't support this?

src/libmeasurement_kit/ooni/web_connectivity.cpp
hellais: That's interesting. I should probably look into this as it's probably a bug in ooni-backend. Filed: https://github.com/TheTorProject/ooni-backend/issues/110

include/private/dns/getaddrinfo_async.hpp
```diff
@@ -61,7 +61,11 @@ std::vector<Answer> getaddrinfo_async_parse_response(const std::string &name,
             throw GenericError(); /* Avoid g++ warning */
         }
         if (p->ai_canonname != nullptr) {
-            answer.hostname = p->ai_canonname;
+            Answer cname_ans = answer;
```
hellais: Is this an API change? Was `answer.hostname` ever used? We should at the very least document it in the changelog.

bassosimone: No, there was no use case where we were using this. I'd say it was unexpected behavior (i.e. inconsistent with our API) what we were doing previously.

src/libmeasurement_kit/ooni/web_connectivity.cpp
bassosimone: Well, to be more specific: the backend said it could not connect to and the JSON contained a `null` where a boolean was expected. That triggered a `nlohmann::json` exception.

include/private/libevent/dns.hpp
bassosimone: It is probably possible to do that with some hacking. I recall there is a way to convince it doing so, but decided not to bother implementing it, since we're using the `system` resolver for all cases except when we need to force a specific DNS server. Since the plan is to switch over to `c-ares` in the future, I figured it was more efficient not to spend time on hammering this further.

bassosimone:
I am going to merge. Apart from builds that I suppressed, there is the typical failure with ooni-resources where sometimes we get a `503` and the test actually expects a `200`.

PR #1335, opened 2017-09-06
fix(test): use cURL CA

bassosimone:
The usage of a basic CA broke tests in #1332. Use always the
cURL CA except in the only case where we need two CAs. In such
case, use cURL CA and the basic CA as fallback.

Rename basic CA for clarity.


bassosimone:
I suppressed the `travis-ci/pr` build.

PR #1336, opened 2017-09-09
refactor(net): ongoing work

bassosimone:
This branch contains ongoing work for refactoring net. My (tentative) objectives are:

- simplify the code and make it easier to read
- allow to interact with a `Transport` using a full proactor pattern (i.e. similar to asio's `async_read` and `async_write` and to `libuv`)
- refactor all code related to connecting using libevent into a single file
- fix the bug that we cannot do socks5 over ssl
- allow to measure precisely the amount of bytes sent and received (at ssl level)
- refactor socks5 to be a simple factory
- rationalize usage of errors and be able to measure more ssl errors
- allow to capture the content of the initial ssl handshake
- refactor more code (including algorithms on transports) to be public templates using mocks for testability
- use a single bufferevent for ssl code when possible (should improve performance)
- improve performance of sending by using `evbuffer` reference rather than copying a string

I am structuring this branch as a set of simple, self-contained patches that ideally can cherry-picked and applied to master when they're ready. Probably, at the end, this branch will not even be merged but all of its content will be cherry-picked and merged into master.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,4 @@
 ---
 BasedOnStyle: LLVM
 IndentWidth: 4
-ContinuationIndentWidth: 6
 ---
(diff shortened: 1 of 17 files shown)


bassosimone:
Also this in-progress branch can be closed.

PR #1337, opened 2017-09-10
refactor(common): ongoing work

bassosimone:


bassosimone:
We can close this branch. I merged into master all that I could, the rest is to be considered experiments.

hellais:
Not 1337 enough.

PR #1345, opened 2017-09-11
feature(net): implement `continue_reading()`

bassosimone:
Based on the #1336 integration branch.

Supersedes #1041, whose original log message was:

> Supersedes #929
>
> Related to #741
>
> To complete:
>
> - [ ] make sure that we stop the parser at the end of the first request to prevent it from parsing further requests (otherwise, given two back to back requests, the code would return only the second one) <- see #1363 
> - [ ] add some regress tests
> - [ ] make sure that issues fixed by #1174 and #1175 are still fixed here

Compared to #1041, this PR has a more readable / clear diff.


bassosimone:
I think this can be deferred to v0.8.0-beta, since it will take a little to review it.

bassosimone:
Closing in favor of #1468

PR #1355, opened 2017-09-13
Release: v0.7.2

bassosimone:
Excerpt from the logs of @lorenzoPrimi that led me to think this fix was needed (I could not reproduce the bug, so I needed to code it blindly and @lorenzoPrimi will test the fix):

```
 D/MK: runner: starting reactor in background...
 D/MK: runner: starting reactor in background...
                                                                   
            --------- beginning of crash
 A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 12220 (pool-3-thread-1)
 D/MK: runner: background reactor running...
 D/MK: runner: background reactor running...
 D/MK: orchestrate::find_location: scheduling
 D/MK: orchestrate::find_location: starting
 D/MK: http::request: {
 D/MK: http::request:     http/method => GET
 D/MK: http::request:     http/url => http://geoip.ubuntu.com/lookup
 D/MK: http::request: }
 D/MK: resolve_hostname: geoip.ubuntu.com
 A/libc: /Volumes/Android/buildbot/src/android/ndk-r15-release/external/libcxx/../../external/libcxxabi/src/abort_message.cpp:74: void abort_message(const char *, ...): assertion "terminating with uncaught exception of type std::runtime_error: event_base_dispatch() failed" failed
 A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 12253 (atory.ooniprobe)
```

Now, this was clearly an MK bug. I wonder whether from the app side it is correct that two threads are started basically at the same time and what events could trigger that.


include/private/common/async_runner.hpp
```diff
@@ -106,11 +106,22 @@ class AsyncRunner : public HasGlobalFactory<AsyncRunner>,
...
+        // N threads can race into this function and can both attempt to
+        // start the shared reactor. All the threads but the first one will
+        // fail and an excepton will be thrown. Fix by preventing multiple
+        // threads from entering into this function. This is actually a
+        // blind fix because I cannot reproduce the problem but it happens
+        // quite predictably to @lorenzoPrimi and @hellais. This should
+        // not be an issue with v0.8.x where the Worker class already has
+        // a barrier preventing this and where we will use a private reactor
+        // for each specific operation anyway.
+        std::unique_lock<std::mutex> _{barrier_};
```
hellais: This fix looks reasonable to me

bassosimone:
Thanks, @hellais

PR #1357, opened 2017-09-13
fix(e/n/connect.cpp): use the custom reactor

bassosimone:
Depends on #1356


bassosimone:
I am going to merge the PR since the build that matters has passed and the diff is tiny.

PR #1364, opened 2017-09-16
fix(http/request.cpp): more efficient ignore_body

bassosimone:
If ignore body is set, do not even bother setting a callback
for the body so to avoid another data copy


bassosimone:
Most builds have succeeded at this point, I am going to merge.

PR #1365, opened 2017-09-18
fix(upgrade): correctly deal with upgrade

### LINKED ISSUE #1312: Add test to ensure we deal with upgrade to http2
See #1227.

bassosimone:
While there, add regress tests.

Closes #1312.

### DIFF
--- a/include/private/http/response_parser.hpp
+++ b/include/private/http/response_parser.hpp
@@ -181,9 +181,25 @@ class ResponseParserNg : public NonCopyable, public NonMovable {
     size_t parser_execute(const void *p, size_t n) {
         size_t x =
             http_parser_execute(&parser_, &settings_, (const char *)p, n);
+        // FIX: I initially coded `upgrade` as the following commented out code
+        // does, because I did read [the documentation of http-parser](
+        // https://github.com/nodejs/http-parser#the-special-problem-of-upgrade)
+        // but it seems that is written more with servers in mind.
+        //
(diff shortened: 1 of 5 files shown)


include/private/http/response_parser.hpp
```diff
@@ -181,9 +181,25 @@ class ResponseParserNg : public NonCopyable, public NonMovable {
...
+        //
+        // This fixes for example the `http://aseansec.org/` URL.
+        //
+        // I am going to keep the code below commented out and this comment
+        // so the choices I made are clear and documented.
+        /*
         if (parser_.upgrade) {
             throw UpgradeError();
         }
+        */
```
hellais: Maybe then you want to just delete the code in here and move the comment above into the commit message?

bassosimone:
Note: the unit tests fail because I need to remove (or comment out) the test that was checking that `UPGRADE` was actually thrown by the parser.

include/private/http/response_parser.hpp
bassosimone: Yeah, makes sense!

hellais: Yes I agree with this fix. Actually `Upgrade` is used to "kind of" advertise support for HTTP2 (or spdy) so in those cases the client knows it can use it (like you say "informational"). Maybe in the future we want to also support HTTP2 so maybe it's worth referencing this commit in that ticket to know the point of the code where that logic should/could be hooked.

bassosimone:
Squashing and merging!

PR #1368, opened 2017-09-19
dns: make sure callbacks are delayed

bassosimone:
Rather than having to call `call_soon()` every time, just call it once when we enter into the API.

While there, reindent some bits of code with `clang-format`.


bassosimone:
I am going to proceed in merging this.

PR #1374, opened 2017-09-21
feature(common): add libevent-based proactor

bassosimone:
This branch contains an implementation of the `proactor` model for sockets using libevent. It is unlikely the final pull request will be exactly like this, as I think I would like to make more changes.

### DIFF
--- a/include/measurement_kit/common/libevent/socket.hpp
+++ b/include/measurement_kit/common/libevent/socket.hpp
@@ -1,0 +1,377 @@
+// Part of measurement-kit <https://measurement-kit.github.io/>.
+// Measurement-kit is free software under the BSD license. See AUTHORS
+// and LICENSE for more information on the copying conditions.
+#ifndef MEASUREMENT_KIT_COMMON_LIBEVENT_SOCKET_HPP
+#define MEASUREMENT_KIT_COMMON_LIBEVENT_SOCKET_HPP
+
+#include <cassert>
+#include <deque>
+#include <event2/buffer.h>
+#include <event2/bufferevent.h>
+#include <event2/event.h>
+#include <event2/util.h>
+#include <measurement_kit/common/detail/mock.hpp>
+#include <measurement_kit/common/detail/utils.hpp>
+#include <measurement_kit/common/reactor.hpp>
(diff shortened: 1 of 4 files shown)


include/measurement_kit/common/libevent/socket.hpp
```diff
@@ -0,0 +1,377 @@
...
+static inline void mk_socket_read(bufferevent *, void *);
+}
+
+namespace mk {
+inline namespace libevent {
+
+template <MK_MOCK(evbuffer_new), MK_MOCK(evbuffer_add_reference),
+        MK_MOCK(evbuffer_add_buffer), MK_MOCK(evbuffer_pullup),
+        MK_MOCK(evbuffer_get_length)>
+class LibeventSocketBuffer : public mk::SocketBuffer {
```
bassosimone: This is basically a simplified `mk::net::Buffer`. We should use `mk::net::Buffer` instead.

include/measurement_kit/common/libevent/socket.hpp
```diff
@@ -0,0 +1,377 @@
...
+        logger->debug2("socket %lld: alloc", sock->fd());
+        return sock;
+    }
+
+    static SharedPtr<Socket> attach(socket_t sockfd, SharedPtr<Reactor> reactor,
+            SharedPtr<Logger> logger) {
+        return make_internal_(sockfd, reactor, logger)->self_;
+    }
+
+    static void connect(sockaddr *sa, socklen_t salen, Settings settings,
```
bassosimone: Note: for this to be complete, we also need support for SSL.

include/measurement_kit/common/socket.hpp
```diff
@@ -4,18 +4,72 @@
...
+    virtual evbuffer *as_evbuffer() = 0;
+
+    virtual size_t size() = 0;
+
+    virtual const void *data() = 0;
+
+    virtual ~SocketBuffer();
+};
+
+class Socket {
```
bassosimone: The `proactor` should actually live in the `mk::net` namespace. The `Transport` should be based on that.

bassosimone:
This is another experiment I should have kept private

PR #1375, opened 2017-09-21
OONI whatsapp tests plus @bassosimone changes

bassosimone:


bassosimone:
@joelanders, while trying to merge, I noticed the following issues:

1. The `"whatsapp_endpoints_blocked"` key was not implemented; I tried to do this in 8d0c173

2. I added some assertions and [now they're triggered by travis](https://travis-ci.org/measurement-kit/measurement-kit/jobs/278239094#L2179)

3. apparently, if `dns_many` fails, there is not output in the final JSON

Could you look into this?

@hellais I guess these issues mean that we cannot release a `v0.7.3` with WhatsApp included by today and perhaps it's best to not wait for WhatsApp for making the next mobile release.

joelanders:
Ok, I forgot to fill out the `'whatsapp_endpoints_blocked'` part of the report (list of hostnames we fail to TCP connect to). In adding it now, I'm thinking we should change the inputs to `tcp_many()` here. Currently, we're only trying to connect to the IPs from `dns_many()` that we think are in the Whatsapp network. That's from some old/outdated CIDR listing they posted, so actually I think we should probably try TCP connecting to whatever the endpoint hostnames resolve to.

Thoughts?

bassosimone:
> so actually I think we should probably try TCP connecting to whatever the endpoint hostnames resolve to.

It seems reasonable to me. I wonder what the spec says, tho.

joelanders:
Ok, have the missing report keys filled and am now TCPing to every IP the DNS lookup responds with.

I guess we should figure out what values to put for "whatsapp_endpoints_status" and "whatsapp_endpoints_dns_inconsistent" for now. (like, in the meantime while we don't have a way of checking DNS consistency)

joelanders:
Ok, going to TCP only to "consistent" IPs after all. Closed that issue.

bassosimone:
I am going to merge when travis reports success.

joelanders:
:champagne:

PR #1382, opened 2017-09-23
Repair failing regress tests

bassosimone:
It seems Politecnico di Torino has now applied some filter
that prevents our regress tests from fetching a PDF.

Repair by using other URLs.


bassosimone:
Merging with a successful build, a suppressed build, and coveraged that is reduced because previously it already existed a branch having the same name,

PR #1383, opened 2017-09-23
fix(cmake): define project name

bassosimone:
Should enable one to `find_package(MeasurementKit)`.

Requested by @hellais.

Depends on: #1382


bassosimone:
There is no need to wait for any build to complete, so I suppressed all of them, because the change in this PR is related to a build system (cmake) not tested in travis.

PR #1384, opened 2017-09-23
refactor(net): transform socks5 into a factory

bassosimone:
Should also fix using https over Tor, which previously was broken. It will probably also enable to run NDT over Tor, but this is something I need to check.


bassosimone:
Update: This is actually not working because, when doing SSL over SOCKS, we must first establish the SOCKS channel and then, only later, we must setup the SSL channel.

bassosimone:
This is very old code, at this point. Merging it would be hard. Better to forget about it.

PR #1386, opened 2017-09-24
Modularise cmake

hellais:
* Add MEASUREMENTKIT_ROOT to support overriding the path to where
measurement-kit sources live


cmake/Common.cmake
```diff
@@ -0,0 +1,116 @@
+# Definitions:
+
+if (NOT MEASUREMENTKIT_ROOT)
+    set (MEASUREMENTKIT_ROOT "${CMAKE_SOURCE_DIR}")
+endif (NOT MEASUREMENTKIT_ROOT)
```
hellais: This is added. If allows somebody importing this cmake file to override the `MEASUREMENTKIT_ROOT` with their own path to where measurement-kit lives. This will be clear why below.

cmake/LibMeasurementKit.cmake
```diff
@@ -0,0 +1,27 @@
+## libmeasurement_kit
+
+file(
+  GLOB
+  MK_LIBRARY_SOURCES
+  "${MEASUREMENTKIT_ROOT}/include/measurement_kit/*.hpp"
+  "${MEASUREMENTKIT_ROOT}/include/measurement_kit/*/*.hpp"
+  "${MEASUREMENTKIT_ROOT}/src/libmeasurement_kit/*/*.c"
+  "${MEASUREMENTKIT_ROOT}/src/libmeasurement_kit/*/*.cpp"
```
hellais: Here I use the `MEASUREMENTKIT_ROOT` variable in front. Since by default it's set to `CMAKE_SOURCE_DIR` it should not break anything when used from the root of the repo (`CMAKE_SOURCE_DIR` will always point to the absolute path of where the `CMakeLists.txt` is stored)

cmake/MeasurementKit.cmake
```diff
@@ -0,0 +1,19 @@
+## measurement_kit
+
+file(
+  GLOB
+  MK_PROGRAM_SOURCES
+  "src/measurement_kit/*.cpp"
+  "src/measurement_kit/*/*.cpp"
```
hellais: Maybe you also want this to be prefixed with `MEASUREMENTKIT_ROOT`. I didn't set it since I don't need to compile the binary, but for consistency this could be changed too.

cmake/Tests.cmake
```diff
@@ -0,0 +1,17 @@
+## tests
+
+enable_testing()
+
+file(GLOB MK_TESTS_SOURCES RELATIVE "${CMAKE_SOURCE_DIR}" "test/*/*.cpp")
```
hellais: Idem

cmake/MeasurementKit.cmake
bassosimone: Yes, please, do that

cmake/Tests.cmake
bassosimone: Ditto

bassosimone:
Thank you! I will test this to make sure it does not break other workflows and proceed with merging!

bassosimone:
@hellais this is working for me. Do you want to add the "for consistency" additions that you suggested above, so that I can test again and merge this into master?

cmake/MeasurementKit.cmake
bassosimone: ping @hellais

cmake/Tests.cmake
bassosimone: ping @hellais

PR #1388, opened 2017-09-25
fix(collector_client): recognize build number

bassosimone:
Closes TheTorProject/ooniprobe-ios#131

This needs to be backported in v0.8.x and v0.7.x.

### DIFF
--- a/src/libmeasurement_kit/ooni/collector_client.cpp
+++ b/src/libmeasurement_kit/ooni/collector_client.cpp
@@ -30,7 +30,7 @@ using namespace mk::report;
 
 static const std::regex re_name{R"(^[A-Za-z0-9._-]+$)"};
 static const std::regex re_version{
-    R"(\bv?(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)(?:-[\da-z\-]+(?:\.[\da-z\-]+)*)?(?:\+[\da-z\-]+(?:\.[\da-z\-]+)*)?\b)",
+    R"(^v?(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)(?:-[\da-z\-]+(?:\.[\da-z\-]+)*)?(?:\+[\da-z\-]+(?:\.[\da-z\-]+)*)?$)",
     std::regex::icase};
 
 static std::map<std::string, std::regex> mandatory_re{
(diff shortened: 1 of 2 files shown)


src/libmeasurement_kit/ooni/collector_client.cpp
```diff
@@ -15,7 +15,7 @@ using namespace mk::report;
 
 static const std::regex re_name{"^[A-Za-z0-9._-]+$"};
 static const std::regex re_version{
-    "^[0-9]+.[0-9]+(.[0-9]+(-[A-Za-z0-9._-]+)?)?$"};
+    R"(^[0-9]+\.[0-9]+(\.[0-9]+(-[A-Za-z0-9._-]+)?(\+[0-9]+)?)?$)"};
```
hellais: Can we make this regexp looser at this point? What is the purpose of having validation of it inside of MK?

In theory this is also a valid semver string:
```
0.1.1-alpha.2+23.b.a
```

Which that regexp would fail on.

Maybe stealing this regexp is a good idea: https://github.com/sindresorhus/semver-regex/blob/master/index.js

bassosimone: > What is the purpose of having validation of it inside of MK?

Be strict in what you send, liberal in what you accept.

> Maybe stealing this regexp is a good idea: https://github.com/sindresorhus/semver-regex/blob/master/index.js

Let me see...

hellais: > Be strict in what you send, liberal in what you accept.

Technically what you are doing here is the opposite, though, given the fact that this thing is given to you by the user of the library.

bassosimone: So, I have imported the regex you suggested!

bassosimone:
[Permalink to the regex, useful to understand, test](https://regex101.com/r/7Q589e/1)

hellais:
👍

bassosimone:
>> Be strict in what you send, liberal in what you accept
> Technically what you are doing here is the opposite, though, given the fact that this thing is given to you by the user of the library.

What you say is true from the point of view of the library user. At the same time, it is false from the point of view of sending data to the collector. I adopted the latter point of view.

Since the library and the applications are actually decoupled, one could probably argue that the library should optionally validate what the application sets, rather than enforcing it as a policy.

PR #1394, opened 2017-09-26
fix(logger): force ofile flush for each entry

bassosimone:
Use `std::endl` rather than `\n` when emitting logs. Force flushing logs at end of test and fixes TheTorProject/ooniprobe-ios#80.

This is the forward port of the fix tested for `v0.7.4` to `master`.

Cc: @hellais


bassosimone:
All the builds that I did not suppress have terminated successfully. Merging!

Issue #1397, opened 2017-09-28
mk v0.7.x: iOS crash during connect


bassosimone:
What follows stems from a crash log sent to us by a user. I strongly believe this crash is actually also the correct explanation of the issue originally reported as TheTorProject/ooniprobe-ios#122.

This is a subset of the original crash log shared with us. It is related to `1.2.0(24)`. The related version of MK is `0.7.2`. I found this running `string|grep '0\.7'` on the binary.

```
[snip]
Thread 12 Crashed:
0   ???                           	0000000000 0 + 0
1   ooniprobe                     	0x0028e8b0 0x49000 + 2382000
2   ooniprobe                     	0x0028e196 0x49000 + 2380182
3   ooniprobe                     	0x0028dc88 0x49000 + 2378888
4   ooniprobe                     	0x00285b34 0x49000 + 2345780
[snip]

Binary Images:
0x49000 - 0x4b0fff ooniprobe armv7  <1534c2f313873e4bbcbc9eb9e6996084>
/var/containers/Bundle/Application/DC79AD22-E0DF-4A4A-90D1-1D98D08601C9/ooniprobe.app/ooniprobe
[snip]
```

The process of symbolicating the crash from command line was not easy, but I managed to do it and I am going to document exactly the steps that I followed:

1. @lorenzoPrimi provided me with a `.zip` containing, among other things, a directory named `ooniprobe-$date.xcarchive` that was, I think, the archive submitted to Apple

2. this directory contained `dSYMs/ooniprobe.app.dSYM/` that I `cp -R`-ed into a new empty directory

3. as explained in the [Apple documentation](https://developer.apple.com/library/content/technotes/tn2151/_index.html#//apple_ref/doc/uid/DTS40008184-CH1-SYMBOLICATE_WITH_ATOS), the tool I needed is called `atos` (there are other ways to do that but I couldn't make them work)

4. from the above report I gathered that the architecture was `armv7` and that the load address of the `ooniprobe` module was `0x49000`: with this information I was able to run `atos` to symbolicate

Thus, to symbolicate entry 1 above (`0x0028e8b0`), I run the following command:

```
$ atos -arch armv7 \
  -o ooniprobe.app.dSYM/Contents/Resources/DWARF/ooniprobe \
  -l 0x49000 0x0028e8b0
```

and obtained as output:

```
mk::net::connect(
    std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >,
    int,
[...]

PR #1402, opened 2017-10-03
Fix for systems that don't support thread_local

bassosimone:
There is empirical evidence that MK v0.7.5 crashes on 32 bit iOS systems
because thread_local isn't working properly.

See #1397.

This diff is an attempt to fix the issue in a very light way. Basically,
retain the thread local semantic, but allocate a cache very time if
thread_local is known to misbehave.

This fix makes the code less efficient for those devices where thread_local
is not working, of course. In fact, basically, for those devices any kind of
caching is disabled.

Yet, given that this seems to affect a minority of our users, I've decided
to go for an as-simple-as-possible fix that prevents the crash.

(Also, `armv7` and `armv7s` are going to disappear anyway in the near future,
because Apple is now mostly focusing on `arm64`.)

While there I've refactored `libssl.hpp`'s cache factory to return a shared
pointer rather than a reference, in the hope that, if the shared pointer
is not initialized, this will lead to an exception and not a crash.

### DIFF
--- a/build/ios/cross
+++ b/build/ios/cross
@@ -16,6 +16,9 @@ MINIOSVERSION="9.0"
 if [ "$PLATFORM" = "iphoneos" ]; then
     EXTRA_CONFIG="--host=arm-apple-darwin --target=arm-apple-darwin --disable-shared"
     MINVERSION="-miphoneos-version-min=$MINIOSVERSION"
+    if [ "$ARCH" = "armv7" -o "$ARCH" = "armv7s" ]; then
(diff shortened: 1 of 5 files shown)


bassosimone:
By reading the disassembly, I have verified that for iOS now we only set `thread_local` for `arm64` and we don't set if for `armv7` and `armv7s`. So, I guess we can move on with merging this.

TODO:

- [ ] understand why travis fails (it seems it cannot build `coveralls`?!)
- [ ] forward port this to `master`

Issue #1413, opened 2017-10-10
See if we can use C++11 rather than C++14


bassosimone:
It seems there are use cases where MK can be used in legacy systems. Currently, we're using C++14, which is quite recent. Going back to C++11 is probably feasible, because most of the code base is actually in C++14 and we are not using too many C++14 features. Going back to C++98 is probably very complex, because we depend a lot on lambdas, so I think attempting that would be a very bad idea.

Yet, using C++11 rather than C++14, is probably conducive in the mid term to help people that have legacy constraints to be more happy, because, with time passing, I imagine that legacy systems will maybe start supporting C++11 whereas support for C++14 will probably be added later.

bassosimone:
Related PR: #1432.

I am going to close this issue as WONTFIX. I discussed this matter with @hellais quite at length on a public slack channel (`#ooni-dev`) and we were in agreement that the benefits of C++14 seem to out-weight the potential benefits of converting back all the code to C++11, given that the people with which I talked wanted C++98 and returning this back with the standard is not gonna happen.

PR #1432, opened 2017-10-10
Start returning to C++11

bassosimone:
It is probably better to return back to C++11 since this is less
modern than C++14, and so probably more widely available.

This will take some time to implement, beyond the current, very easy
patch, and specifically we need to address the following:

1. use a safe pattern when we _need_ to move stuff inside callbacks
   rather than just copying because we are using threads and thus
   we must make sure that a callback is owned by a single thread.

2. rewrite serial and parallel callbacks to avoid doing magic
   that requires C++14


bassosimone:
Will close as WONTFIX. See #1413

PR #1445, opened 2017-10-28
reactor: correctly deal with exception in child thread

### LINKED ISSUE #1387: worker: pass exception to reactor
Currently, if an exception occurs inside a worker, it will be just printed:

1. For starters, it would be good if it was printed on the same logger of the test (this is an easy fix).

2. Moreover, when the Worker is used by a Reactor, it would be correct to tell the reactor that an exception occurred. The Reactor in turn should probably leave "as if" the exception occurred in the I/O loop itself.

The purpose of the second change is that I would like to move towards a model where an unhandled exception causes the current reactor to leave and, when we're running `nettests`, stops by the try...catch setup by the thread that is running the specific `nettest`. (In the future I would also like to pass around handlers for routing exceptions, but in the interest of incrementally evolving MK I believe it is saner to proceed piecemeal and for now the most striking fix is indeed 2.)

bassosimone:
<strike>Depends on #1446.</strike>

Closes #1387.

Required for v0.8.0-beta.

### DIFF
--- a/include/measurement_kit/common/reactor.hpp
+++ b/include/measurement_kit/common/reactor.hpp
@@ -6,6 +6,7 @@
 
 #include <measurement_kit/common/callback.hpp>
 #include <measurement_kit/common/error.hpp>
+#include <measurement_kit/common/logger.hpp>
 #include <measurement_kit/common/shared_ptr.hpp>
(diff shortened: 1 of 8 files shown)


hellais:
🐳

bassosimone:
Thanks!

PR #1446, opened 2017-10-29
Refactor libevent reactor (1/2)

bassosimone:
I started working on #1387 and realized I needed actually some refactoring work to be done before implementing that properly. This is the first step towards that end.

Content of the diff:

- refactor the comments
- move all the attributes at the end of the class and make them private
- use the same code path for call_soon / call_later and pollin_once / pollout_once
- rather than using MK_POLLIN / MK_POLLOUT use directly EV_READ / EV_WRITE
- move the code from the C level callback to a static method of the class

PR #1447, opened 2017-10-29
Refactor libevent reactor (2/2)

bassosimone:
Depends on #1446, which should be merged first.

In this diff we basically make sure that we do not leak pending callbacks if we exit abruptly from the reactor loop, either because stop() was called or because of an exception.

With this building block in place, we can then safely route exceptions in child threads without leaking memory, which is what stopped me from finishing #1445.


bassosimone:
After thinking a bit more about it, I am going to close this PR. The direction towards which we're going is the one where all exceptions are fatal in MK. Hence, no point in making code more complex for the purpose of dealing gracefully with exceptions so to clean up memory.

PR #1456, opened 2017-11-01
feature(raw_ptr): smart pointer for raw C pointers

bassosimone:
Makes the management of the lifecycle of raw C pointers much
more structured than it's currently.

The plan is to use it to make more easy and yet robust to
change the underlying bufferevent used by Transport.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,7 @@
 /test/common/maybe
 /test/common/parallel
 /test/common/range
+/test/common/raw_ptr
 /test/common/settings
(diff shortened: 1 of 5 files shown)


hellais:
Nice I like this idea. Good to merge.

hellais:
🐳

bassosimone:
Thanks! I have since refactored the implementation to reuse existing libc++ code. Will merge when travis checks indicate success and full coverage of the newly added code.

PR #1484, opened 2017-11-11
base_test: break circular ref, fix node bindings

### LINKED ISSUE #1483: mk-node hangs when running a test

bassosimone:
Closes #1483


bassosimone:
This could even potentially go in, but I'd like to reflect a little more on matters like:

- I moved the callback down, is this reasonable or should I leave it where it was originally?

- I removed `stop()` because I was assuming it was unnecessary, is that true and moreover what does actually happen when we're in node in terms of quitting the loop? What behavior would be correct?

These require a little bit of reflection. Hence, we can probably even afford to bless v0.8.0 without this fix, since we have not released node bindings yet.

bassosimone:
Okay, I actually sorted out pretty much everything. This can go on master and should also be backported in stable so that we have this functionality working starting from v0.8.0-stable.

bassosimone:
Changed the milestone and epic as result of this being ready.

PR #1487, opened 2017-11-11
fix(dash): enforce random UUID

bassosimone:
This is so hotfixish that we can for sure include it in v0.8.0.


bassosimone:
Moved to v0.8.0 as this is ready.

PR #1488, opened 2017-11-11
dash: include congestion signals into result

bassosimone:
An interesting question when doing streaming like stuff is whether we're generating congestion (i.e. whether we are pushing the envelope) or not. Interestingly, in my current location (D.C.; Comcast) I don't see any congestion.


bassosimone:
I’d say no to this. We can compute this information from already existing data and we do not care about exposing this information to the user at present.

PR #1494, opened 2017-11-16
ooni/utils.cpp: fix search-title segfault

### LINKED ISSUE #1276: web_connectivity: regex segfault with Linux
As mentioned in c6303b8 and #1275:

> We are using very aggressive regular expressions in web_connectivity. The problem is that the GNU C++ library has a very broken regex implementation that uses stack recursion and segfaults when the expression is too big. This happens for me on my development `Mac mini` running Ubuntu zesty for `https://jezebel.com` as input. I spotted this running pre-release checks.
>
> So, I investigated ways to use libc++ also on Linux and not only on macOS.  This diff modifies docker specs such that, when we build with clang, we use libc++.
> 
> This opens up the possibility of packaging MK for debian using libc++ rather than using GNU C++ library, because we can see whether it is working or not.
>
> Also updated the tutorial to explain how to build on a generic Linux system using libc++.
>
> See also: <https://stackoverflow.com/questions/36304204>.

The specific, aggressive regular expression is the one for extracting the title.

In addition to the above issue, evaluating the regexp over long bodies is also very expensive. I have seen MK freezing for seconds when searching for the title (again: on my `Mac mini`). This is probably the correct topic for another issue, however.

### LINKED ISSUE #1277: web_connectivity: slow regex to extract title (when body is huge)
As mentioned in #1276:

> The specific, aggressive regular expression is the one for extracting the title.
>
> In addition [...], evaluating the regexp over long bodies is also very expensive. I have seen MK freezing for seconds when searching for the title (again: on my Mac mini). [...]

Perhaps we can mitigate this issue by trying to evaluate the regexp on an initial slice of the body.

(I think that slicing the body is a good fix for the slowness, but not to prevent the segfault documented in #1276, which instead should be addressed by using a C++ library that works better, see #1275)

bassosimone:
Closes #1276.

Closes #1277.

### DIFF
--- a/src/libmeasurement_kit/ooni/utils.cpp
+++ b/src/libmeasurement_kit/ooni/utils.cpp
@@ -139,7 +139,7 @@ ErrorOr<std::string> GeoipDatabase::resolve_asn(std::string ip,
 }
 
 std::string extract_html_title(std::string body) {
-  std::regex TITLE_REGEXP("<title>([^<]*)</title>", std::regex::icase);
+  std::regex TITLE_REGEXP("<title>([^<]{0,1024})</title>", std::regex::icase);
   std::smatch match;
 
   if (std::regex_search(body, match, TITLE_REGEXP) && match.size() > 1) {
--- a/src/libmeasurement_kit/ooni/utils.cpp
+++ b/src/libmeasurement_kit/ooni/utils.cpp
@@ -139,7 +139,7 @@ ErrorOr<std::string> GeoipDatabase::resolve_asn(std::string ip,
 }
 
 std::string extract_html_title(std::string body) {
-  std::regex TITLE_REGEXP("<title>([\\s\\S]*?)</title>", std::regex::icase);
+  std::regex TITLE_REGEXP("<title>([^<]*)</title>", std::regex::icase);
   std::smatch match;
 
   if (std::regex_search(body, match, TITLE_REGEXP) && match.size() > 1) {
--- a/src/libmeasurement_kit/ooni/utils.cpp
+++ b/src/libmeasurement_kit/ooni/utils.cpp
@@ -139,7 +139,7 @@ ErrorOr<std::string> GeoipDatabase::resolve_asn(std::string ip,
 }
 
 std::string extract_html_title(std::string body) {
-  std::regex TITLE_REGEXP("<title>([^<]{0,1024})</title>", std::regex::icase);
+  std::regex TITLE_REGEXP("<title>([^<]{0,4096})</title>", std::regex::icase);
   std::smatch match;
 
   if (std::regex_search(body, match, TITLE_REGEXP) && match.size() > 1) {


src/libmeasurement_kit/ooni/utils.cpp
```diff
@@ -139,13 +139,25 @@ ErrorOr<std::string> GeoipDatabase::resolve_asn(std::string ip,
...
+  begin_title += sizeof "<title>";
+  assert(begin_title <= zerop); // Must always hold
+  auto end_title = strcasestr(begin_title, "</title>");
+  if (end_title == nullptr) {
+    return "";
   }
-  return "";
+  assert(end_title >= begin_title); // Should really not happen
+  auto siz = static_cast<size_t>(end_title - begin_title);
+  return std::string{begin_title, siz};
```
darkk: IMHO the code serves the goal to get rid of regex engine, but...
- I'd rather ship "known good" regex engine if there are other regexes in the MK (maybe there are none of them, I don't know)
- Seems, `\0` breaks strcasestr()
- Wild-life example that breaks this code is `<title id="pageTitle">Facebook</title>` from view-source:https://www.facebook.com/ :-)

darkk:
lgtm

bassosimone:
Maybe I should also push a regress test that would segfault MK as part of this PR to have in future the possibility to understand if we're introducing regressions.

bassosimone:
Yeah, second thinking at this, I guess it would be better to have also a regress test.

hellais:
Are we sure 4096 is a reasonable number to limit the title to?

I think it may be worthwhile to pick this number based on the historical HTML title. I will do that now.

hellais:
screen shot 2017-11-28 at 18 03 47

darkk:
4k seems to be enough according to http://archive.li/2017.11.28-192421/https://stackoverflow.com/questions/8516235/max-length-of-title-attribute
Chrome has 1k characters limit :)
The thing that worries me a bit is unicode / byte boundary / html entity boundary.

bassosimone:
@darkk Yeah, I _think_ it will be base64 encoded in that case, but not sure whether it would be useful to collect it, then. Perhaps:

- cut the title

- check whether it's valid UTF8 (we have code for doing that in tree)

- if it is, use it on the client side, otherwise let the pipeline decide

?

bassosimone:
I have deployed a mitigation in `v0.9.0-beta`. I believe we _should_ migrate to a more robust regex engine (specifically `re`, which I have already studied a bit), for `v0.10.0`.

bassosimone:
(Also, I think we _should_ probably push some bits of processing to the pipeline or to the test helper rather than computing this bit of information in the probe. Parsing the `<title>` of a `HTML` web page using regular expressions is brittle anyway, so perhaps we want to have a place where we parse the full page and extract the title. We may also want to extract additional URLs from the webpage as well?)

Issue #1492, opened 2017-11-15
Fix the IM test implementations


hellais:
**Whatsapp**

* [x] Calculate the blocking `ok` value based on the result of the tcp connect test

* [x]  Compare which of the consistent IP responses we are able to do a successful TCP connect to

* [x]  Not consider (towards blocking determination) the endpoints to be accessibly solely based on the DNS consistency result (https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/whatsapp.cpp#L307)

* [ ] Only populate the `whatsapp_endpoints_blocked` with the endpoints that are blocked AND consistent: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/whatsapp.cpp#L239

**Facebook Messenger**

* [x] Facebook ASN should not be hardcoded inline, but rather be a DEFINE at the beginning of the file: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/facebook_messenger.cpp#L76

* [x] For the DNS blocking we should use the same logic that we use for the whatsapp test, if "at least 1 IP is consistent" (matches facebook AS) we consider DNS to not be the means of blocking: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/facebook_messenger.cpp#L80. Which is actually already what you do here: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/facebook_messenger.cpp#L86

* [x] Again, as the whatsapp test, we should do a TCP connect to all IPs (not just the consistent ones) and consider it TCP blocked when ALL of them fail: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/facebook_messenger.cpp#L141 and not just when ONE of them fails: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/facebook_messenger.cpp#L150

**Telegram**

* [x] It seems like `telegram_tcp_blocking` is always being set to `true` and is never `false`. See: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/telegram.cpp#L21 and https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/telegram.cpp#L37. I suspect this is a typo.

[...]

PR #1498, opened 2017-11-16
IM test logic fixing

joelanders:
https://github.com/measurement-kit/measurement-kit/issues/1492

This fixes all remaining bugs.

### DIFF
--- a/src/libmeasurement_kit/ooni/telegram.cpp
+++ b/src/libmeasurement_kit/ooni/telegram.cpp
@@ -83,7 +83,7 @@ static void http_many(const std::vector<std::string> urls, std::string type,
                 logger->info(
                     "telegram: success HTTP connecting to %s", url.c_str());
                 if (type == "endpoints") {
-                    (*entry)["telegram_tcp_blocking"] = false;
+                    (*entry)["telegram_http_blocking"] = false;
                 } else if (type == "web") {
                     if (extract_html_title(response->body) != "Telegram Web") {
                         (*entry)["telegram_web_status"] = "blocked";
@@ -132,6 +132,7 @@ void telegram(Settings options, Callback<SharedPtr<report::Entry>> callback,
 
     // if any endpoints are (TCP or HTTP) reachable, switch this to false
     (*entry)["telegram_tcp_blocking"] = true;
+    (*entry)["telegram_http_blocking"] = true;
 
     mk::fcompose(mk::fcompose_policy_async(),
         [=](Callback<> cb) {
--- a/src/libmeasurement_kit/ooni/telegram.cpp
+++ b/src/libmeasurement_kit/ooni/telegram.cpp
@@ -34,7 +34,7 @@ static void tcp_many(const std::vector<std::string> ip_ports, SharedPtr<Entry> e
                 logger->info("telegram: success TCP connecting to %s:%d",
                     ip.c_str(), port);
                 result["status"]["success"] = true;
-                (*entry)["telegram_tcp_blocking"] = true;
+                (*entry)["telegram_tcp_blocking"] = false;
             }
             (*entry)["tcp_connect"].push_back(result);
             txp->close(nullptr);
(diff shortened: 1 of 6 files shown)


joelanders:
Bleh I see the CI's compiler wants me to add a declaration for the new "is this IP in FB's ASN?" function I added.

src/libmeasurement_kit/ooni/telegram.cpp
```diff
@@ -34,7 +34,7 @@ static void tcp_many(const std::vector<std::string> ip_ports, SharedPtr<Entry> e
                 logger->info("telegram: success TCP connecting to %s:%d",
                     ip.c_str(), port);
                 result["status"]["success"] = true;
-                (*entry)["telegram_tcp_blocking"] = true;
+                (*entry)["telegram_tcp_blocking"] = false;
```
hellais: This is not the only logic bug. You are also re-initialising `telegram_tcp_blocking` to `true` in both `http_many` and `tcp_many`. Doing so means that the condition "consider it to work if at least one is working" will not be met in many cases.

hellais:
> CI's compiler wants me to add a forward declaration for the new "is this IP in FB's ASN?"

@bassosimone knows better, but if the function is defined as not defined as `static` you need to declare it inside of the header. You should either mark it as static or a declaration for it.

hellais:
I haven't actually run it, but a cursory look at the diff looks good.

As mentioned on slack there is still one logic bug missing to be squashed:

> Only populate the whatsapp_endpoints_blocked with the endpoints that are blocked AND consistent

Current you are erasing it from the list of `blocked_hostnames` only when it fails to connect and is consistent: https://github.com/measurement-kit/measurement-kit/blob/48f4e3cefdd60806a14e2c5f8dd438d6e6b5df57/src/libmeasurement_kit/ooni/whatsapp.cpp#L241, you should be doing the opposite.
Erasing it from the list of blocked domains when either it succeeds or it fails but is inconsistent.

hellais:
My checklist for testing this is:

* [ ] Do another pass at reading the diff
* [ ] Smoke test by compiling and running it
* [ ] Setup a local system to simulate various forms of telegram, whatsapp and fb blocking and confirm it produces consistent measurements
* [ ] Merge

hellais:
Here are some things I am noticing while comparing the output of `ooniprobe` to the output of the MK test:

## Telegram

* The telegram test output misses the key `telegram_http_blocking`. Grepping for this key in the source I don't find it.

joelanders:
@hellais ready for another review.

hellais:
@joelanders so it seems like the facebook test is still not writing the keys properly.

Here is a sample of the report having set my `/etc/hosts` with what I described above:

```
        "facebook_b_api_dns_consistent": false,
        "facebook_b_api_reachable": false,
        "facebook_b_graph_dns_consistent": false,
        "facebook_b_graph_reachable": false,
        "facebook_dns_blocking": true,
        "facebook_edge_dns_consistent": true,
        "facebook_edge_reachable": true,
        "facebook_external_cdn_dns_consistent": true,
        "facebook_external_cdn_reachable": true,
        "facebook_scontent_cdn_dns_consistent": true,
        "facebook_scontent_cdn_reachable": true,
        "facebook_star_dns_consistent": true,
        "facebook_star_reachable": true,
        "facebook_stun_dns_consistent": false,
        "facebook_stun_reachable": null,
        "facebook_tcp_blocking": true,
```

It states that there is `facebook_tcp_blocking`, but that is not the case as the DNS response is inconsistent. Similarly to the whatsapp and telegram test you should set `tcp_blocking` to true only when CONSISTENT IPs are unreachable (not inconsistent as in this case).

hellais:
I checked the telegram and whatsapp tests and things there are good 👍

joelanders:
@hellais pong.

hellais:
Ok, I did another pass of testing and this looks good to me.

🐳 

I will let @bassosimone do the honors of merging it if it looks good to him as well.

bassosimone:
Yap, will look into this later today!

src/libmeasurement_kit/ooni/telegram.cpp
bassosimone: FTR, it seems that now this has been fixed!

PR #1503, opened 2017-11-24
Make the delta in the every.cpp test less strict

hellais:


bassosimone:
Sure, we can do that.

PR #1506, opened 2017-11-25
refactor(shared_ptr): start enforcing pure-wrapper role

bassosimone:


src/libmeasurement_kit/common/reactor.cpp
```diff
@@ -8,8 +8,7 @@
 namespace mk {
 
 /*static*/ SharedPtr<Reactor> Reactor::make() {
-    return locked_global(
-            []() { return SharedPtr<Reactor>{new LibeventReactor<>}; });
+    return SharedPtr<Reactor>{std::make_shared<LibeventReactor<>>()};
```
bassosimone: locked global seems unnecessary for a factory method.

PR #1526, opened 2017-12-20
fix(README.md): explain how to obtain iOS bitcode

bassosimone:
Depends on #1528


bassosimone:
Since this is really minor and only related to README.md, I am going to self merge this god speed using my administrator privileges to do that before even waiting for travis 😈

PR #1527, opened 2017-12-20
Update Vagrantfile to use zesty instead of yakkety

rschulman:
The current Vagrantfile uses yakkety, but trying to pull the image gives a 404. This PR replaces yakkety with zesty, which does not 404.


bassosimone:
It's unclear why coveralls coverage was never reported. I'll wait a little more and then, if there are no good news from coveralls, proceed with merging using my superpowers. Thank you!

PR #1529, opened 2017-12-28
refactor: remove private headers

bassosimone:
Private headers below `include` were looking more tidy but at the end of
the day it seems simpler to have such headers as close as possible to
implementation files, because then it's more natural to look for files
and to make sure they are in sync, etc.

While there, make sure the project name is correctly spelled. This was
done by using a one-off script that I did write.


bassosimone:
For the records, this is the script I used to make this PR:

```python
#!/usr/bin/env python

import sys, os

def slug(name):
    return name.upper().replace("/", "_").replace(".", "_")

for path in sys.argv[1:]:
    if not os.path.isfile(path):
        continue

    if path.endswith(".h") or path.endswith(".hpp"):
        basepath = os.path.basename(path)
        baseslug = slug(basepath)
        lines = []
        for line in open(path, "rb"):
            if line.startswith("#ifndef PRI"):
                lines.append("#ifndef %s\n" % slug(path))
            elif line.startswith("#define PRI"):
                lines.append("#define %s\n" % slug(path))
            else:
                lines.append(line)
        open(path, "wb").write("".join(lines))

for path in sys.argv[1:]:
    if not os.path.isfile(path):
        continue

    lines = []
    for line in open(path, "rb"):
        if line.startswith("// Part of measurement-kit") or line.startswith(" * Part of measurement-kit ") or line.startswith("// Measurement-kit is") or line.startswith(" * Measurement-kit is"):
            line = line.replace("measurement-kit", "Measurement Kit", 1).replace(
               "Measurement-kit", "Measurement Kit")
        lines.append(line)
    open(path, "wb").write("".join(lines))

for path in sys.argv[1:]:
    if not os.path.isfile(path):
        continue

    lines = []
    for line in open(path, "rb"):
        if line.startswith("#include \"private/"):
            line = line.replace("#include \"private/", "#include \"src/libmeasurement_kit/")
        lines.append(line)
    open(path, "wb").write("".join(lines))

```

bassosimone:
For the records, with @hellais we also concluded the following (that is meant to help us streamline the process while having all commits vetted):

* that, even for hotfixes, having a single commit for each thing is better
* that publishing the script used to perform mechanical changes is a good thing(TM) to do

Thanks for the review!

PR #1530, opened 2017-12-28
fix(error_or): make construction unambiguous

bassosimone:
Depends on #1529. See 93d93f098f02781605f236ef208c48e63193f9ca for the individual commit. Closes #1509.

- - - 

This diff is a response to @joelanders observation that sometimes
making an ErrorOr was leading to ambiguity.

The way in which I implemented it was that I removed all the
partial constructors, so now you always implement returning
ErrorOr like when returning a tuple. Which is more or less the
thing we're doing anyway. And also is more readable.

PR #1531, opened 2017-12-28
fix(error): define virtual destructor and simplify

bassosimone:
Depends on #1530. See 64ca9588c7ee7f9e9a76424f42f85db02596f38c for the individual commit.

- - -

The bug we're trying to fix here is that, since Error derives from
std::exception, its destructor should be virtual.

The proper way of doing that would probably be that of defining a
virtual destructor for every error class that we have.

Adding an out-of-line destructor for each error class would take
time and/or being hackish (by augmenting the existing macro?).

Adding an in-line virtual destructor is bad practice (and indeed
clang warns about that; see `-Wweak-vtables`).

However, it seems we really don't need all these classes.

They were introduced when we were thinking about throwing errors
and then catching them using different catch clauses.

But, during #1235, #1262, and #1447, it occurred to me repeatedly
that we should only use exceptions as fatal errors because to have
exceptions and async code in the same plate is too tricky.

So, if that is the intent, we can get away with a single Error
class -- with proper virtual destructor, of course.

At least in the library.

Then, in the regress tests, we have a few instances where we do
actually check that an exception is of the specified type.

But, it also turn out that doing that is not exactly needed for
our code testing goal. Which is that of making sure that the code
throws `$whatever` in case something unexpected happens.

If that is the end, we can safely just check if the code throws
something and then rely on code coverage to make sure that all
the possible fatal error paths throw.

After all, we already have cases where different error paths lead
to the same exception being thrown, so we already cannot use the
exception type alone to tell whether we cover all paths.

While there, force the single-argument constructor to be called
explicitly, as that is considered good practice.

While there, force moving in the child error.

### DIFF
--- a/include/measurement_kit/common/error.hpp
+++ b/include/measurement_kit/common/error.hpp
@@ -45,25 +45,18 @@ namespace mk {
 /// as part of measurement-kit v0.8.0 as part of a refactoring of common.
 class Error : public std::exception {
   public:
-
     /// \brief The default constructor initializes the error code to zero.
     Error() : Error(0, "") {}
 
     /// \brief The constructor with error initializes the error code to the
     /// specified error code.
-    Error(int e) : Error(e, "") {}
+    explicit Error(int e) : Error(e, "") {}
 
     /// \brief The constructor with error and reason string initializes both.
     Error(int e, std::string r) : code{e}, reason{r} {
         if (code != 0 && reason == "") {
             reason = "unknown_failure " + std::to_string(code);
         }
-    }
-
-    /// \brief The constructor with error code, reason string, and child
-    /// error also initializes the child error.
-    Error(int e, std::string r, const Error &c) : Error(e, r) {
-        child_errors.push_back(c);
     }
 
     /// The bool operator evaluates to true if the error code is nonzero.
@@ -87,8 +80,8 @@ class Error : public std::exception {
     const char *what() const noexcept override { return reason.c_str(); }
 
     /// `add_child_error` allows you tou append additional child errors.
-    void add_child_error(const Error &err) {
-        child_errors.push_back(err);
+    void add_child_error(Error &&err) {
+        child_errors.push_back(std::move(err));
     }
 
     /// `child_errors` contains all the child errors.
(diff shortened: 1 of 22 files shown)


bassosimone:
What I do expect from a reviewer (saying this explicitly after a discussion on `#ooni-dev` involving @darkk and @agrabeli and based on @lorenzoPrimi previous input):

- [ ] a negative comment if my coalescing of all errors into a single class with different status codes appears to be a bad idea in the reviewer's opinion

- [ ] whether I did some obvious mistake (like the one highlighted by @darkk in this comment)

bassosimone:
@pboothe your comments really help me to think at `Error` in a more abstract way, thanks!

Most likely, currently, each `Error` is a type since originally we also wanted to `throw`. Thus, we figured that each error having its own type would have helped to distinguish between them using `catch`. This now strikes me as a bit Java-ish since even the C++ library does differently (see, e.g. the fact that there is a single `std::system_error` class from which one can extract the proper `std::error_code`).

Since this requirement of being able to `throw Error`s is going away, the design that would strike me as more proper for `Error` would be that `Error` is a container for an error code. In this respect, it would probably wise to check whether we could use `std::error_code` for representing error codes. (I'd rather user components from the standard library, to minimize dependencies.)

I am still undecided whether to go down the value road or keep the types.

I have a sense that resubmitting as another PR the subset of this diff that makes `Error`'s single-argument constructor explicit and enforces move semantic would be okay. Then, since we want to stop throwing, it would probably make sense to have a PR where we stop throwing any `Error` anyway. This would allow us to guarantee that no code catching `const std::exception &` was making the assumption that that would catch either stdlib exceptions or `Error`s. Thus, it will be 100% safe to apply your suggestion of not having `Error` inherit from `std::exception`. Finally, if the decision is to use values, more refactoring can come.

bassosimone:
Too many conflicts to think about saving it. Also, this PR has been pivotal to steer me towards a direction where I make MK much more simple than it is now. In itself, it's already a great outcome!

pboothe:
Hooray for simplicity!

PR #1533, opened 2017-12-28
cleanup(build): don't save vendor patches

bassosimone:
It is so bureaucratic to have them around when developing to
improve our dependencies. Considering that one can always get
the diff between a pristine import and subsequent changes,
since everything is now committed, there is no point in keeping
the patches around. Actually, they can lead to mistakes as one
can modify a dependency without saving the patch and this would
later lead to annoyances. Better to dispose of them.


bassosimone:
Things that I want to know from the reviewers:

- [ ] do you feel like this is bad idea? why?

- [ ] I do not remove from `./build/dependency` the functionality of patching sources when building, since this _may_ still be useful for customizing the build for specific archs, okay?

hellais:
If the patches that are removed are in fact already part of the upstream or the dependency that is fetched via `./build/vendor/import foo` (i.e. the end result is the same by going through all the steps) I don't see why this should be a problem.

bassosimone:
> If the patches that are removed are in fact already part of the upstream or the dependency that is fetched via ./build/vendor/import foo (i.e. the end result is the same by going through all the steps) I don't see why this should be a problem

No, the patches are local. `./build/dependency` used to work like this:

1. we fetched dependency `$foo` using `build/spec/$foo`

2. we automatically applied patches

3. we built for a platform

Instead, now, with vendored sources, it works like this:

1. we edit `build/spec/$foo` to specify version

2. we do a pristine import of `$foo` using `./build/vendor/import` and commit

3. we apply and commit patches

Which means that, when we update we do the following:

1. `git diff` from previous pristine import to save patches

2. edit `build/spec/$foo` to update version

3. do a pristine import again using `./build/vendor/import` and commit

4. manually apply and commit patches saved in step 1.

In theory we could also use another workflow where we keep in the repository pristine dependencies and patches and we continue to apply them as part of the build process.

I find more transparent to commit our patches along with the vendored dependencies tho.

bassosimone:
After discussion with @hellais on slack, I implemented 4c932b1. FTR, @hellais evidence a critical flaw of my approach, i.e. that one may want to apply different patches with different purposes, rather than a single big patch. This is what Tor browser does for example. I still think that is more robust and transparent to have changes inside the vendored dependencies. Hence in 4c932b1 I have added a script that gives you all the patches applied to a vendored dependency since latest pristine import.

bassosimone:
So, another good suggestion here is to add a CI script that makes sure that the committed patches are exactly the patches that are in tree. Another opinion is that we should maximize maintainer's comfort rather than minimize strangeness. I will spend some time thinking about all of this.

bassosimone:
Since #1554 has been merged, this PR can now be closed, as an alternative but more complex approach to address the same problem addressed by #1554.

PR #1534, opened 2017-12-28
maint(*): update gitignores

bassosimone:
Updates the `.gitignore` files by re-running `./update/gitignore`. This used to be done automatically with `./autogen.sh` but, due to Windows changes, now it's something to be done manually.

I removed manually the `.gitignore` in `test/libevent` which is empty after recent refactoring.

PR #1535, opened 2017-12-29
feature(cmake): enable Windows build

bassosimone:
<strike>Depends on #1534. See c9c5e75 for the real commit.</strike>

- - -

Radically rework CMake to build for Win32/Win64:

1. since everyone says to avoid using glob, introduce script to
   automatically generate list of things to build

2. use tricks borrowed from the CMake files of others, like defining
   an object library to compile just once

3. rename the script performing the checks but otherwise do not
   modify it dramatically, just add better Windows support

Along with upcoming changes in the library, this allows to build a
MK binary that doesn't make smoke on Windows.

Further tweaks and cleanups are left as future work.


bassosimone:
@hellais This is 1/3 of the work remaining to have Windows support in master. The 2/3 part is to make everything compile. The 3/3 part is to flag specific bits for DLL export (we may ignore this if we're using a static library to create a single bundle with Go).

bassosimone:
I adeed the "work in progress" flag because I still need to see whether this works on Windows.

bassosimone:
Okay, I have now tested (using #1536) that the cmake code in this PR is good for allowing us to build MK not only on macOS and Linux but also on Windows. Ready for review again.

cmake/configure.cmake
```diff
@@ -111,9 +148,29 @@ if(UNIX)
...
+  # seem to be generated by the cmake script used on Windows.
+  list(APPEND MK_LIBS event_pthreads)
+  list(APPEND MK_LIBS event_openssl)
+  list(APPEND MK_LIBS ssl)
+  list(APPEND MK_LIBS crypto)
+  list(APPEND MK_LIBS event_core)
+  list(APPEND MK_LIBS event_extra)
+
+elseif(WIN32)
+  # On Windows we expect a specific version of libressl. We can possibly make
```
willscott: I don't see anything in this PR that provides guidance on how to get this version of these libraries.  Is this just 'gonna work' in that cmake will get the right version? or do i need to have these libs / where can i get them?

build/spec/libressl  has a URL for `libressl-2.5.5.tar.gz` which is not obviously the same

bassosimone: Good question! So, that ABI number is actually equal to v2.5.5. I have discovered anyway that it seems it's not needed to clearly specify the ABI, so it would probably be better to drop the explicit ABI number, to make the thing overall more maintainable.

hellais:
I have been trying to get this to run on windows, but I can't get it to run.

What I did was:

1. Install cmake using choco
2. Run `./autogen.sh --cmake` (I figured this was the thing to run from the diffs above, it should probably be documented somewhere)
3. As suggested by the autogen.sh output I ran: `./scripts/cmake/run`, but this failed to detect my cmake installation
3-bis. I then ran `C:\Application Support\Cmake\bin\cmake.exe .`

The last command failed almost immediately with it failing to detect my C compiled. I tried setting some of the suggested flags (`CC=/usr/bin/cc`), but still to no avail.

@bassosimone can you document somewhere how you setup your windows development environment and what steps I need to follow in order to get it to build?

hellais:
Ok, so I got a bit further along (note: I am actually using this branch to build: https://github.com/measurement-kit/measurement-kit/pull/1536), but I am still stuck at another point.

Here are some notes of what I did to setup my environment:

Notes on windows setup:

1. Install Visual Studio Community

When you are prompted to install packages, be sure to enable "Desktop development with C++".

2. Enable developer mode

To do this you have to go into Settings->For developers (type this to find the menu)

and then choose Developer mode

3. Open developer command prompt and run the `bash` command. You will be prompted to install Ubuntu on windows.

There will be some setup process at the end of the installation where you have to create a unix username and password.

I told it to keep as locale `en_US`.

What follows are command all run inside of `bash` (you should probably re-run bash by typing `bash`).

4. Run `sudo apt-get update`

5. Run `sudo apt install cmake clang libgeoip-dev`

6. Run `./autogen.sh --cmake`

7. Run `./scripts/run/cmake`

The point I am stuck at is that cmake is complaining that it can't find the geoip libraries. 
```
art@DESKTOP-D2TEL7L:/mnt/c/Users/Arturo/Projects/OONI/measurement-kit$ ./script/cmake/run                                                                                                                                                       - Build system: Linux                                                                                                   - You are not on macOS, probably no cmake tweak is needed                                                               - about to invoke: 'cmake  .'                                                                                                                                                                                                                   === BEGIN CMAKE OUTPUT ===                                                                                                                                                                                                                      CMake Error at cmake/configure.cmake:120 (message):                                                                       geoip missing; Use -DMK_GEOIP to specify where it is installed (e.g.                                                    -DMK_GEOIP=/usr/local if it is installed under /usr/local)                                                            Call Stack (most recent call first):                                                                                      CMakeLists.txt:5 (include)                                                                                                                                                                                                                                                                                                                                            -- Configuring incomplete, errors occurred!                                                                             See also "/mnt/c/Users/Arturo/Projects/OONI/measurement-kit/CMakeFiles/CMakeOutput.log".                                See also "/mnt/c/Users/Arturo/Projects/OONI/measurement-kit/CMakeFiles/CMakeError.log".                                                                                                                                                         === END CMAKE OUTPUT ===                                                                                                                                                                                                                        - It seems that cmake failed. Possibly useful options in case it                                                          is complaining about missing dependencies:                                                                                                                                                                                                        -D MK_GEOIP=/geoip/install/path                                                                                         -D MK_LIBEVENT=/libevent/install/path                                                                                   -D MK_OPENSSL=/openssl/install/path                                                                                                                                                                                    
```
@bassosimone how did you install the libraries required by measurement-kit? Did you use apt-get install or something else?

bassosimone:
@hellais it seems this branch become stale and broken. I am going to try to repair it by rebasing onto master and removing the stuff that conflicts. Will let you know when done.

bassosimone:
Step 1: merging master: ✔️

bassosimone:
Step 2: improve `script/build/README.md`: ✔️

bassosimone:
Step 3: Make sure we can _start_ a build: ✔️

bassosimone:
Work to make MK compile will now continue in #1536: 👨‍🏭

bassosimone:
Superseded by #1590.

PR #1545, opened 2018-01-11
cleanup(include): stop throwing Error

bassosimone:
Based on #1544. See 223f9b9 for the standalone diff.

This should also be quite hofix-ish and straightforward based on discussion at #1531.

Text of the commit message follows:

- - -

As discussed in #1531, one of the first steps to start simplifying
`Error` is to make sure we are not going to throw `Error`.

Rather, change the code to throw `std::runtime_error`.

The specific reason why I want to avoid throwing `Error` is that I am
worried that that will break assumptions made by code that was using
`catch (const std::exception &)` to _also_ catch for `Error`.

By never throwing `Error`, we can make this issue disappear.

Then we can move forward by detaching `Error` and `std::exception`.

This diff does not address all the cases where we `throw Error` rather
it starts addressing some of these. More diffs/PR will follow.

I am following this process to keep the workload low for reviewers
and to make sure that I can wrap up work at end of day.

### DIFF
--- a/include/measurement_kit/common/settings.hpp
+++ b/include/measurement_kit/common/settings.hpp
@@ -27,7 +27,7 @@ class Settings : public std::map<std::string, Scalar> {
 
     /// \brief `get()` returns the specified \p key if set; otherwise it
     /// returns the default value \p def_value.
-    /// \throw ValueError if the value associated to \p key cannot be
+    /// \throw std::runtime_error if the value associated to \p key cannot be
     /// converted to the specified type.
     template <typename Type> Type get(std::string key, Type def_value) const {
         if (find(key) == end()) {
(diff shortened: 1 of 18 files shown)


bassosimone:
As of 223f9b9 there is a [Travis build error](https://travis-ci.org/measurement-kit/measurement-kit/jobs/327721934#L2126) that really appears to be caused by the resources test failing because it could not fetch resources from github as expected. This seems to be unrelated with the PR proper.

After #1544 is merged, I'll merge in the master branch so to reduce the diff. I expect the above error to go away as part of the process. (FWIW, resources are a feature that we do not use on OONI and that is scheduled for removal, so I do not think it's worth bothering making the test more robust to such kind of transient failures.)

include/measurement_kit/common/settings.hpp
```diff
@@ -45,8 +45,6 @@ class Settings : public std::map<std::string, Scalar> {
         return at(key).as_noexcept<Type>();
     }
 
-#undef XX
-
   protected:
```
bassosimone: This should have been removed in a previous commit.

bassosimone: I mean: there's no `XX` in this file anymore.

bassosimone:
Thanks, I am going to squash this!

PR #1551, opened 2018-01-26
Update Vagrantfile to use zesty instead of yakkety

rschulman:
Yakkety is no longer supported so Vagrant is unable to download images. Upgrading to zesty.


bassosimone:
This may fail because the build is failing for the same reason. Let me quickly finish the PR to repair it and then I'll ask for your review, and finally we can move forward with this.

bassosimone:
Ah, it seems I fixed this as part of my global find-and-replace strategy in #1552.

bassosimone:
I basically run `git grep yakkety` and replaced it with `$something_else`, which for the Vagrant file actually appears to be `zesty`. Hence, this PR is now empty. Thanks much anyway! :-)

PR #1552, opened 2018-01-26
fix(build): yakkety is no more, find replacements

bassosimone:
- artful for valgrind, where valgrind seems to work okay (no weird issues like in zesty, but correctly like it used to work on yakkety)

- zesty for coveralls, because cpp-coveralls doesn't seem to work on artful

### DIFF
--- a/build/docker/coveralls
+++ b/build/docker/coveralls
@@ -8,4 +8,4 @@ REPOROOT=$(cd $(dirname $(dirname $(dirname $0))) && pwd -P)
 #
 # (Honestly, to use a different distro for builds is not that bit
 # problem, IMHO. What matters is that we can measure coverage.)
-$REPOROOT/build/docker/script/setup ubuntu:zesty coveralls
+$REPOROOT/build/docker/script/setup ubuntu:xenial coveralls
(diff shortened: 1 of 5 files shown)


bassosimone:
@pboothe 

Yeah, it's annoying and a bit weird the type of error. [From what I understand](https://travis-ci.org/measurement-kit/measurement-kit/jobs/332298813#L2463) it's unrelated with cpp-coveralls proper and instead it seems that the certificate-validation callback in the Python `OpenSSL` package, as invoked by `requests`, fails because it cannot find an attribute using `ffi`.

I am anyway doing some extra lazy experiments (does it work with the `bionic` beta) so to report the bug to `$someone`. Maybe I'll open a ticket with `cpp-coveralls`, in case someone else notices the same issue, and then possibly I'll try to understand what the proper venue is.

(All this need for docker forced me to abandon my beloved Windoze machine; I'm so day).

bassosimone:
All checks have passed. It looks like I can merge (even though the interface is inconsistent and tells me that not all checks have passed). Whatever, I'll hit the big green button.

eddyxu:
@bassosimone Not sure whether it is fixed on your site.

It looks like {{requests}} could not find this ssl module. Have you installed https://users-cs.au.dk/sortie/sortix/release/nightly/man/man3/X509_up_ref.3.html ?

PR #1553, opened 2018-01-30
include: reduce places where we throw Error

bassosimone:
This diff reduces the places where we throw `Error` below `include`. This is part of the work required to decouple `Error` and `std::exception`, as discussed with @pboothe in #1531.

### DIFF
--- a/src/libmeasurement_kit/dns/getaddrinfo_async.hpp
+++ b/src/libmeasurement_kit/dns/getaddrinfo_async.hpp
@@ -62,7 +62,7 @@ std::vector<Answer> getaddrinfo_async_parse_response(const std::string &name,
             answer.type = MK_DNS_TYPE_AAAA;
             aptr = &((sockaddr_in6 *)p->ai_addr)->sin6_addr;
         } else {
-            throw GenericError(); /* Avoid g++ warning */
+            throw std::runtime_error("unexpected_family"); // Avoid g++ warning
         }
         if (p->ai_canonname != nullptr) {
             Answer cname_ans = answer;
@@ -73,7 +73,7 @@ std::vector<Answer> getaddrinfo_async_parse_response(const std::string &name,
         }
         char abuf[128];
         if (inet_ntop(p->ai_family, aptr, abuf, sizeof(abuf)) == nullptr) {
-            throw InetNtopFailureError();
+            throw std::runtime_error("inet_ntop_failed"); // Should not occur
         }
         if (p->ai_family == AF_INET) {
             answer.ipv4 = abuf;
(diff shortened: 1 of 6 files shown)


src/libmeasurement_kit/dns/getaddrinfo_async.hpp
```diff
@@ -106,11 +106,7 @@ void getaddrinfo_async(std::string name, addrinfo hints, SharedPtr<Reactor> reac
                       name.c_str(), error.code, error.what());
         std::vector<Answer> answers;
         if (!error && rp != nullptr) {
-            try {
-                answers = getaddrinfo_async_parse_response<inet_ntop>(name, rp);
-            } catch (const Error &err) {
-                error = std::move(err);
-            }
+            answers = getaddrinfo_async_parse_response<inet_ntop>(name, rp);
```
bassosimone: This is probably the hunk that needs most attention. One needs to make sure that `getaddrinfo_async_parse_response` is not `throw`ing anymore.

bassosimone:
Nice! The build is now failing in an interesting way: since I removed the try...except, now we exit with an exception with a thread still active. I think the proper way to fix this is to modify the thread pool used for doing this kind of stuff so that, rather than having a detached thread (which is probably bad) we have a non-detached thread, so that RAII is enforced in case of any exception. This would make the thread pool safer from many points of view (TBH, the fact that I was using a detached thread was already worrying me, so I am actually happy this popped up, because it allows me to more correctly address the way in which the thread pool works and make it safer).

bassosimone:
So, it turns out that my understanding of what happens when an exception reaches the bottom of a thread stack was naive. Of course `std::abort` is called. To make the code testable, then, I need to refactor the code so that exceptions are not throw in that code path, except for exceptional cases.

src/libmeasurement_kit/dns/getaddrinfo_async.hpp
```diff
@@ -73,18 +73,18 @@ std::vector<Answer> getaddrinfo_async_parse_response(const std::string &name,
...
-            throw InetNtopFailureError();
+            return {GenericError{"inet_ntop_failed"}, {}}; // Unexpected
         }
         if (p->ai_family == AF_INET) {
             answer.ipv4 = abuf;
         } else if (p->ai_family == AF_INET6) {
             answer.ipv6 = abuf;
         } else {
-            /* Case excluded above */;
+            throw std::runtime_error("case_excluded_above");
```
pboothe: Why should this throw when the others don't?

bassosimone: Because the case has been, in theory, excluded above. So, if this even happens, I would like to know in a very noisy way. Would you prefer to have `assert(false)` here?

pboothe: Yes. It seems like this is not designed to be catchable, so engaging in the "throw/catch" formalism feels like overkill.

pboothe:
LGTM

PR #1532, opened 2017-12-28
feature(third_party): enable Windows builds

bassosimone:
Depends on #1531. See 9e3c510 for individual commit.

- - -

This diff introduces build system changes that allow one to build
all the default dependencies for Win32 and Win64 targets.

Tested with Visual Studio Community 2017 and CMake 3.10.0.

Will publish the binaries created in here in the dependency
repositories quite soon.

After this commit, the diffs to be applied after updating the
repository start become dangerously stale. I'll fix that.

PR #1554, opened 2018-02-03
Simpler build system

bassosimone:
As mentioned in #1532, after negative feedback pretty much from everyone regarding the way in which I was structuring build scripts, I spent time trying to find out a simpler way.

This PR introduces low-level build scripts that are feature equivalent to the ones in `./build` [*], except that they are simpler (I flattened the architecture and disposed of some gratuitous complexity).

Regarding Windows proper, I removed the requirement that one should not use the Ubuntu shell on Windows to start this build. Without this requirement, things are much more easy than before.

Note that these build scripts are in `./script/build` rather than in `./build` because there are many CMake guides suggesting to create a directory named `./build`. So, I'd like to move away from `./build`.

- - -

What I would like a reviewer to do:

1. comment on the overall approach and way in which I'm structuring things (is the optional complexity reduced? Is it difficult to understand the structure of the build scripts, also given the README.md?)

2. read the scripts and check whether they're reasonable (below I'll publish diff wrt previous version of the scripts, which is still inside of `./build` &mdash; I did this to start moving away from `./build`)

3. optionally run _some_ builds (I think it would especially be useful if @hellais can run the Windows build and generally try to build dependencies using these scripts and provide feedback)

4. comment on the plan I drafted to move forward, which is:

    1. make a full Android and iOS build using these scripts and test them with our apps

    2. if this works, edit existing scripts inside of `./build` to point to the new scripts, but leave the old scripts as _wrappers_ / _aliases_ for now

    3. also, remove vendored code, which now becomes unnecessary

    4. also, since we are assuming we'll be running Unix scripts in all scenario, remove from the repository the autogenerated stuff that we committed and make sure we autogenerate all we need in `./autogen.sh` to avoid having the problem of keeping autogenerated things consistent

    5. do not perform more changes for now, but note that more refactoring is pending (I think more of this can be done after we bless v0.9.0, which is probably something to do ASAP once we have support for Windows and an experimental new API)

    6. rebase the existing PR to build MK on Windows on top of this branch (or better master once this branch is merged) and get a review of the CMake and code changes

- - -

[*] There are also high level build scripts in `./build` but there is no point in touching them and making things simpler until the low-level ones have been vetted, IMHO.


bassosimone:
# Diff between selected files

```diff
--- build/android/cross	2017-12-26 13:19:25.935899700 +0100
+++ script/build/cross-android	2018-02-02 18:39:22.713212100 +0100
@@ -1,10 +1,13 @@
 #!/bin/sh
+
+# Adapted from build/android/cross
+
 set -e
 
-ROOTDIR=$(cd $(dirname $0) && pwd -P)
+ROOTDIR=$(cd $(dirname $0)/../../ && pwd -P)
 
 if [ $# -lt 1 ]; then
-    echo "Usage: $0 ARCH" 1>&2
+    echo "Usage: $0 ARCH [command [arguments...]]" 1>&2
     echo "" 1>&2
     echo "  ARCH: armeabi, armeabi-v7a, arm64, mips, mips64, x86, x86_64" 1>&2
     echo "" 1>&2
@@ -20,7 +23,7 @@
 echo "Cross compiling for $ARCH"
 
 # Place where make-toolchain creates a standalone toolchain
-export ANDROID_TOOLCHAIN=${ROOTDIR}/toolchain/${ARCH}
+export ANDROID_TOOLCHAIN=${ROOTDIR}/MK_BUILD/android/toolchain/${ARCH}
 
 # Directory under the toolchain where we have headers and libs
 export SYSROOT=${ANDROID_TOOLCHAIN}/sysroot
@@ -117,20 +120,17 @@
 #export LIBS="-lm"
 
 #
-# Flags for ./build/dependency (cross-)build system
+# Flags for ./script/build/<package> (cross-)build system
 #
 
-# Tell configure to use the toolchain and build only static libs
-export pkg_configure_flags="--host=${TOOLCHAIN_NAME} --disable-shared $CONFIG_EXTRA"
-
-# Where to install headers and libraries
-export pkg_prefix="${ROOTDIR}"
+# Tell configure to use the toolchain
+export pkg_configure_flags="--host=${TOOLCHAIN_NAME} $CONFIG_EXTRA"
 
-# Tell ./build/dependency that we are cross compiling
-export pkg_cross="jni"
+# Tell ./script/build/<package> that we are cross compiling
+export pkg_os="android"
 
-# Tell ./build/dependency the name of the cross architecture
-export pkg_cross_arch="${DESTDIR_NAME}"
+# Tell ./script/build/<package> the name of the cross architecture
+export pkg_arch="${DESTDIR_NAME}"
 
 if [ $# -gt 0 ]; then
     $@
--- build/ios/cross	2017-12-26 13:19:25.942918000 +0100
+++ script/build/cross-ios	2018-02-03 14:00:08.951298800 +0100
@@ -1,4 +1,7 @@
 #!/bin/sh
+
+# Adapted from ./build/cross/ios
+
 set -e
 
 if [ $# -lt 2 ]; then
@@ -14,16 +17,16 @@
 
 MINIOSVERSION="9.0"
 if [ "$PLATFORM" = "iphoneos" ]; then
-    EXTRA_CONFIG="--host=arm-apple-darwin --target=arm-apple-darwin --disable-shared"
+    EXTRA_CONFIG="--host=arm-apple-darwin --target=arm-apple-darwin"
     MINVERSION="-miphoneos-version-min=$MINIOSVERSION"
     if [ "$ARCH" = "armv7" -o "$ARCH" = "armv7s" ]; then
-        CPPFLAGS="-DMK_NO_THREAD_LOCAL"
+        CPPFLAGS="-DMK_NO_THREAD_LOCAL" # See #1404
     fi
 elif [ "$ARCH" = "i386" ]; then
-    EXTRA_CONFIG="--host=i386-apple-darwin --target=i386-apple-darwin --disable-shared"
+    EXTRA_CONFIG="--host=i386-apple-darwin --target=i386-apple-darwin"
     MINVERSION="-mios-simulator-version-min=$MINIOSVERSION"
 elif [ "$ARCH" = "x86_64" ]; then
-    EXTRA_CONFIG="--host=x86_64-apple-darwin --target=x86_64-apple-darwin --disable-shared"
+    EXTRA_CONFIG="--host=x86_64-apple-darwin --target=x86_64-apple-darwin"
     MINVERSION="-mios-simulator-version-min=$MINIOSVERSION"
 else
     echo "$0: unsupported configuration" 1>&2
@@ -33,10 +36,6 @@
 echo ""
 echo "Cross compiling for $PLATFORM and $ARCH"
 
-ROOTDIR=$(cd $(dirname $0) && pwd -P)
-SOURCEDIR="$ROOTDIR/../../"
-DESTDIR="$ROOTDIR/tmp"
-
 export CC="$(xcrun -find -sdk ${PLATFORM} cc)"
 export CXX="$(xcrun -find -sdk ${PLATFORM} g++)"
 export CPPFLAGS="-arch ${ARCH} -isysroot $(xcrun -sdk ${PLATFORM} --show-sdk-path) $CPPFLAGS"
@@ -45,10 +44,8 @@
 export LDFLAGS="-arch ${ARCH} $MINVERSION -isysroot $(xcrun -sdk ${PLATFORM} --show-sdk-path)"
 
 export pkg_configure_flags="$EXTRA_CONFIG"
-export pkg_make_flags="V=0 $pkg_make_flags"
-export pkg_prefix="$DESTDIR"
-export pkg_cross="${PLATFORM}"
-export pkg_cross_arch="${ARCH}"
+export pkg_os="${PLATFORM}"
+export pkg_arch="${ARCH}"
 
 if [ $# -gt 0 ]; then
     $@
```

bassosimone:
(I asked a review to all the people who provided me feedback on #1532 plus @rschulman who has recently played around with the build scripts for Raspberry and may hence have some insightful comments)

script/build/libressl.asc
```diff
@@ -0,0 +1,182 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
willscott: why is this needed?

bassosimone: To verify the signature of the libressl tarball we download.

script/build/README.md
```diff
@@ -0,0 +1,110 @@
...
+1. you have installed Visual Studio 2017 Community
+
+2. you have installed CMake and opted to have `cmake.exe` inside your path
+   during the install process (important, as that is not the default
+   choice)
+
+3. you have installed Ubuntu and `ubuntu.exe` is available from your
+   path (that should be the case), and you have installed into Ubuntu
+   on Windows all the Linux packages mentioned above when discussing
+   building for Linux or macOS
```
hellais: This is a pretty strong requirements for windows. Keep in mind that this is only available since Windows 10.

I am OK with keeping it like this for the moment, but we should see if it's possible to do it what we need with standard windows tooling.

script/build/libressl.asc
hellais: It may be easier (and equivalent from a security) to have a hash of the specific version you are downloading. I suspect getting that gpg to work properly on windows may not be that easy.

script/build/mk
```diff
@@ -0,0 +1,43 @@
...
+mkdir -p ${pkg_dist_dir}
+
+pkg_build_dir=${pkg_topdir}/MK_BUILD/${pkg_os}/${pkg_arch}/${pkg_name}
+rm -rf ${pkg_build_dir}
+mkdir -p ${pkg_build_dir}
+
+(
+    set -e
+
+    ./autogen.sh # Must run in the toplevel directory
```
hellais: Maybe you should add some check to ensure this is actually run in the top level directory (you probably want the same pattern for every build script).

script/build/mk
```diff
@@ -0,0 +1,43 @@
...
+
+    ../../../../configure --prefix=${pkg_dist_dir} --disable-shared           \
+        --disable-examples                                                    \
+        --disable-binaries                                                    \
+        --with-geoip=${pkg_dist_dir}/../geoip                                 \
+        --with-libevent=${pkg_dist_dir}/../libevent                           \
+        --with-openssl=${pkg_dist_dir}/../libressl                            \
+        ${pkg_configure_flags}
+
+    make V=0 -j4 install
```
hellais: Maybe the parallelism should be something configurable?

script/build/mk
```diff
@@ -0,0 +1,43 @@
...
+
+    # No `.la` files around because they will confuse other `./configure`.
+    rm -rf ${pkg_dist_dir}/lib/*.la
```
hellais: Just as a extra safety measure (in the case `pkg_dist_dir` for whatever reason ends up evaluating to the empty string (or is undefined)), I would suggest checking that it's a non-empty string, so that you never by mistake do `rm -rf /lib/*.la`.

script/build/libressl.asc
bassosimone: > I suspect getting that gpg to work properly on windows may not be that easy.

Once you get `ubuntu.exe` you get everything, including `gpg`. Since I scrapped the assumption that we will not run _any_ Unix script, I'd like to do things exactly as I would do them on any other Unix system.

script/build/README.md
bassosimone: Please: no. If we want to support _building_ on previous versions of Windows, then we would better have everything in the repository and minimise _our_ pain in maintaining.

Having `ubuntu.exe` seems to me a _good_ compromise between having to learn to develop on Windows like a pro and using tooling that we know to do the job.

bassosimone:
> I am assuming the next step is to do a cleanup of ./build in the root.
>
> I would suggest to minimise confusion, if possible, you just delete the build directory as it's redundant at this point.

This is not possible immediately. What you see here is the low level part of the build scripts being rewritten. Migrating all of build would take more time. I do not want to do everything in a single PR since that would be too much chaos and fallback. I'd rather do it incrementally.

script/build/README.md
bassosimone: (Also, to the maximum extent as possible, I'd like to minimise the amount of work entailed by maintaining more than one build system. To this end, for example, I would like to make sure that in the future we _only_ build Measurement Kit using CMake. No more autotools.)

script/build/libressl.asc
bassosimone: (To be more precise, I am doing things so that working on Windows would be functionally equivalent to working in any other Unix system, except for ~10 lines of code that are Windows specific and configure and make the project using Windows tooling.)

script/build/README.md
bassosimone: Also: why should we support _building_ on a version of Windows that we don't have. Really, I do not understand why we would want doing that. Please, let's try to keep the building on Windows simple _for us_, given that we have limited resources. Also: building on Windows 10 is not impairing in any way _using_ Measurement Kit in earlier Windows versions.

script/build/geoip
```diff
@@ -38,11 +39,11 @@ mkdir -p ${pkg_build_dir}
     (
         set -e
         cd libGeoIP
-        make V=0 -j4 install
+        make ${pkg_make_flags} install
```
hellais: It's generally more usual in bash to use `UPPERCASE` variable names for user-overridable variables. This also ensures there are no conflicts with user-overridable variables and those used internally by the script.

script/build/README.md
hellais: Sounds reasonable.

script/build/libressl.asc
hellais: Ok, though I think that verifying the package using a hardcoded hash that you have to update, is in any case more robust than importing a key (and adding it to the default keyring) and calling `gpg --verify`. Using this scheme you will consider a package is valid even if it's signed by ANY key that is present inside of the users keyring (and you are also adding a new key that they may not want to their keyring).

darkk: > verifying the package using a hardcoded hash ... is ... more robust than importing a key

<bikeshed>Technically, I'd say those are **equally** robust as having compromised key in the system keyring brings you compromised CC with `apt-get update && apt-get upgrade`. I don't think that there is major difference between backdoored source code and backdoored compiler.</bikeshed> I'd say both solutions are equally good in terms of _robustness_ and _security_.

script/build/mk
darkk: I'd say, it should be either hardcoded or configurable with default [`MAKEFLAGS`](https://github.com/darkk/home/blob/master/.bashrc#L61-L78). Making additional configuration option to `make` sounds like unnecessary complexity to me.

script/build/libressl.asc
hellais: > Technically, I'd say those are equally robust as having compromised key in the system keyring brings you compromised CC with apt-get update && apt-get upgrade

Well... A part from the fact that the `apt-get` does not use the users keyring, but even if that were the case it would still only be relevant to linux users.

That said, I think the main argument for not using this method is that I am not super happy to have my users keyring contain random keys that I don't care to have in it and some build script modifying the setup of my users home.

If you do decide to keep the pattern of using GPG to verifying (which I don't actually think is a good idea), I would at least provide it some temporary throw-away keyring path with `--keyring --no-default-keyring`.

script/build/mk
darkk: Dropping libtool files from /lib/ is not a tragedy, those are just libtool files used for linking.
MAYBE it should be checked earlier (to understand if the variable is deduced from the environment (pwd and "$0") in a sane way), but checking it just before `rm` sounds like nothing-but-anxiety to me.

script/build/android-toolchain
```diff
@@ -0,0 +1,54 @@
...
+if [ "$ARCH" = "armeabi" -o "$ARCH" = "armeabi-v7a" ]; then
+    ARCH="arm"
+fi
+
+if [ "$API" = "" ]; then
+    if [ "$ARCH" = "arm64" -o "$ARCH" = "mips64" -o "$ARCH" = "x86_64" ]; then
+        API=21
+    else
+        API=14
+    fi
```
darkk: Is some comment on default API level useful here? Why is it 14-by-default and 21 for some platforms?

rschulman:
I am going to bow out of doing code review for this change. I actually do not know very much about Unix build tools (this is why I like Rust and Cargo so much!), except for the bits and pieces I picked up in doing some work on the Raspberry Pi cross-compile.

script/build/libressl.asc
bassosimone: >  I think the main argument for not using this method is that I am not super happy to have my users keyring contain random keys that I don't care to have in it and some build script modifying the setup of my users home.

This is a good concern, and something I did not think about. We could opt for using `signify` &mdash; the native OpenBSD tool for doing these things, which actually works better than GPG as far as solving the problem at hand is the objective. But probably it is simpler if I verify the signtature and then just add the `sha256sum` of such signature in the build system.

I will do that during my next iteration of work on this branch.

script/build/geoip
bassosimone: I didn't know. I'll see whether it's possible to avoid user overridable variables (maybe it suffices to let `MAKEFLAGS` do the job?), and, if not, I'll make the ones that user can override uppercase.

script/build/android-toolchain
bassosimone: Yes, I can add a comment. Basically: 14 is the minimum for 32 bit and 21 the minimum for 64 bit, and these limits are enforced by the NDK script for building the standalone toolchain. I do not remember whether, if passed a too low value, such script bails or just produces a toolchain that does not work. Anyway, I can probably find an issue or PR where I fixed this and point to that.

bassosimone: Look, I've figured out how I've discovered these numbers:

```
$ git diff
diff --git a/script/build/android-toolchain b/script/build/android-toolchain
index 2a869652..65779959 100755
--- a/script/build/android-toolchain
+++ b/script/build/android-toolchain
@@ -40,7 +40,7 @@ fi
 
 if [ "$API" = "" ]; then
     if [ "$ARCH" = "arm64" -o "$ARCH" = "mips64" -o "$ARCH" = "x86_64" ]; then
-        API=21
+        API=14
     else
         API=14
     fi
$ ./script/build/android-toolchain ~/Library/Android/sdk/ndk-bundle arm64
Creating toolchain for API 14 and ARCH arm64 in /Users/sbasso/src/github.com/measurement-kit/measurement-kit/MK_BUILD/android/toolchain/arm64
14 is less than minimum platform for arm64 (21)
$ echo $?
1
```

I am going to reference this comment in the code :-).

bassosimone:
[Pushed new changes](https://github.com/measurement-kit/measurement-kit/pull/1554#commits-pushed-ff487e9) addressing previous feedback from @darkk and @hellais.

hellais:
:shipit:

bassosimone:
So, I am about to hit the big green button. But, before I am doing that, I think it would be smart to do some `pragma` to make sure we can simplify as much as we can as a result of this change.

bassosimone:
> before I am doing that, I think it would be smart to do some pragma to make sure we can simplify as much as we can as a result of this change.

I started doing that as part of #1561, which is currently not complete but touches upon most of the things that needs to be updated as a result of this PR. Going on with merge.

PR #1555, opened 2018-02-06
Sketch out new, simpler API

bassosimone:
We discussed recently about improving the MK API with @hellais, @darkk, @willscott, @lorenzoPrimi, @pboothe, and others. The main issues with the existing [nettests.hpp](https://github.com/measurement-kit/measurement-kit/blob/v0.8.1/include/measurement_kit/nettests/nettests.hpp) API are:

1. it is nonblocking, hence it is difficult to write wrappers (you need to keep state alive)

2. the API surface is O(N) in the number of callbacks (i.e. on the number of objects that we want to share with the caller)

Since recently we acknowledged that we were not going to run more than one test at a single time, we can actually implement a _simpler_ API that is blocking and moves around JSON-like objects, hence solving the two problems identified above, which we creating lots of burden.

In sketching out this new API, I am following a suggestion laid out by @willscott during a video chat, i.e. to make the API as simple as routing around JSON objects. This is exactly what `mk::engine` does. To simplify my life to the maximum extent, I do not event wrap `nlohmann::json`. I just pass it around.

The FFI wrappers around all of this are quite obvious. So I added them.

Then I also added SWIG wrappers. Again this is quite simple.

- - -

# For reviewers

1. please, carefully check the threading code in src/libmeasurement_kit/engine.cpp

2. please, also look into the buffer handling in mk_start_ex() in src/libmeasurement_kit/ffi.cpp

3. I would recommend to avoid discussing the format of events, since this is being worked on by @hellais with help from me, so I would say it's okay to merge this and then finalise the format

4. is the code clear and understandable?

5. any other comment or suggestion?


example/swig/ndt.cpp
```diff
@@ -0,0 +1,29 @@
+// Public domain 2017, Simone Basso <bassosimone@gmail.com.
+
+#include <measurement_kit/swig.hpp>
+
+#include <stdlib.h>
+
+#include <iostream>
+
+#include <measurement_kit/common/nlohmann/json.hpp>
```
bassosimone: not needed

example/swig/ndt.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+int main() {
+    std::string settings = R"({
+        "type": "Ndt",
+        "verbosity": "INFO"
+    })";
+    mk::swig::Task task;
+    bool okay = task.initialize(settings);
+    if (!okay) {
+        std::clog << "cannot parse json" << std::endl;
```
bassosimone: be more informative

include/measurement_kit/engine.h
```diff
@@ -0,0 +1,152 @@
+/* Part of Measurement Kit <https://measurement-kit.github.io/>.
+   Measurement Kit is free software under the BSD license. See AUTHORS
+   and LICENSE for more information on the copying conditions. */
+#ifndef MEASUREMENT_KIT_ENGINE_H
+#define MEASUREMENT_KIT_ENGINE_H
+
+/*
+ * Low level engine for running Measurement Kit tests.
```
bassosimone: explain why not exposed by default

include/measurement_kit/engine.h
```diff
@@ -0,0 +1,152 @@
...
+
+/// Task is a task that Measurement Kit can run. You create a task with Task()
+/// by passing it the desired settings as a nlohmann::json. The minimal settings
+/// JSON must include the task type (see MK_ENUM_TASK for all types).
+///
+/// Creating a Task also creates the thread that will run it. Altough you can
+/// construct more than one Task at a time, Measurement Kit will make sure that
+/// tasks run sequentially in the order in which they were created.
+///
+/// If you enabled task events (see MK_ENUM_EVENT), a Task will emit
```
bassosimone: fixme

include/measurement_kit/engine.h
```diff
@@ -0,0 +1,152 @@
...
+///
+/// You can also interrupt a running task using interrupt().
+///
+/// The destructor, ~Task(), will join on the Task thread. That is, it will
+/// wait for the Task to complete before destroying all the resources.
+///
+/// See example/engine/ndt.cpp for an example of usage.
+class Task {
+  public:
+    /// Task() creates and starts a Task configured according to @p settings.
```
bassosimone: provide more info?

include/measurement_kit/engine.h
```diff
@@ -0,0 +1,152 @@
...
+#endif // MK_ENGINE_INTERNALS
+#endif /* __cplusplus */
+
+/**
+ * MK_ENUM_SETTINGS enumerates the possible keys that the settings object
+ * passed to a task should have. The XX macro takes three arguments: the
+ * name of the setting, the type that the key should have, and whether the
+ * specified setting must be present in the JSON object.
+ */
+#define MK_ENUM_SETTINGS(XX)                                                   \
```
bassosimone: misleading macro name

include/measurement_kit/engine.h
```diff
@@ -0,0 +1,152 @@
...
+    XX(INFO)                                                                   \
+    XX(DEBUG)                                                                  \
+    XX(DEBUG2)
+
+/**
+ * MK_ENUM_EVENT enumerates all possible event types. To specify the
+ * events you want to receive from a task, you should pass the desired event
+ * types as strings (e.g. "LOG") when you configure the task's enabled events.
+ * For example, the minimal JSON to run NDT with the "LOG" and "PERFORMANCE"
+ * events enabled is:
```
bassosimone: fix

include/measurement_kit/ffi.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef MEASUREMENT_KIT_FFI_H
+#define MEASUREMENT_KIT_FFI_H
+
+/*
+ * Measurement Kit Foreign-Function-Interface (FFI) API.
+ *
+ * See include/measurement_kit/engine.h for more documentation.
+ *
+ * See example/ffi/ndt.cpp for example usage.
+ */
```
bassosimone: add notes regarding ownership

include/measurement_kit/swig.hpp
```diff
@@ -0,0 +1,137 @@
...
+       caller takes ownership of the value returned by wait_for_next_event().
+
+    2. Move semantic is not wrapped by SWIG. We replace it with passing a
+       constant reference for the settings of the Task.
+
+    3. Parsing of the stringified JSON can throw. We have therefore created a
+       specific method for initializing, given that throwing in general is
+       something that makes life more complicate for bindings. (We may still
+       fail and abort() in case of bad error conditions, but a parse error
+       is clearly an avoidable error.
```
bassosimone: missing )

include/measurement_kit/swig.hpp
```diff
@@ -0,0 +1,137 @@
...
+    #include <measurement_kit/swig.hpp>
+    %}
+    %include <measurement_kit/swig.hpp>
+    ```
+*/
+
+/// Task is something that Measurement Kit can do.
+class Task {
+  public:
+#ifndef SWIG
```
bassosimone: explain

src/libmeasurement_kit/ffi.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+struct mk_task_ : mk::engine::Task {
+    using mk::engine::Task::Task;
+};
+
+mk_task_t *mk_task_start(const char *settings) noexcept {
+    if (settings == nullptr) {
+        return nullptr;
+    }
```
bassosimone: add note about returning null

test/engine/ndt_example.cpp
```diff
@@ -0,0 +1,2 @@
+// We want to run the unmodified example through Valgrind
+#include "example/engine/ndt.cpp"
```
bassosimone: fixme: should wrap with `ENABLE_INTEGRATION_TESTS`

example/ffi/ndt.cpp
```diff
@@ -0,0 +1,46 @@
...
+            fprintf(stderr, "ERROR: cannot wait for next event\n");
+            exit(1);
+        }
+        const char *serio = mk_event_serialize(event);
+        if (serio == NULL) {
+            fprintf(stderr, "ERROR: cannot serialize event\n");
+            exit(1);
+        }
+        // A bit ugly but this is `null` as a serialized JSON
+        done = (strcmp(serio, "null") == 0);
```
willscott: It seems somewhat awkward to have the task completion come out as a not-explicit event on the event stream.

How hard would it be to implement a `mk_task_is_active(task)` or equivalent that could be used as a condition in a while loop?

bassosimone: Not hard. It was part of a previous draft. I think I omitted it because I was going down the path of reducing complexity recursively and probably exaggerated a bit :-).

src/libmeasurement_kit/engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+        return !pimpl_->running || !pimpl_->deque.empty();
+    });
+    // must be first so we drain the queue before emitting the final null
+    if (!pimpl_->deque.empty()) {
+        auto rv = std::move(pimpl_->deque.front());
+        pimpl_->deque.pop_front();
+        return nlohmann::json{std::move(rv)};
+    }
+    assert(!pimpl_->running);
+    return nlohmann::json(); // this is a `null` JSON object
```
lorenzoPrimi: A null json object is a null object or a json object with empty field?

lorenzoPrimi:
Looks great to me too! Can't wait to test it!

src/libmeasurement_kit/engine.cpp
bassosimone: The former thing!

bassosimone: Based on @willscott's suggestion, now one is not required to check whether the returned JSON is null, but rather he/she can use `is_running()` to check whether things are working. Yet, I believe a `null` JSON string is returned in any case when the queue hits the end. I am not sure how I can avoid this, tho, since I need anyway to return something to the caller if I'm blocked in reading the queue.

src/libmeasurement_kit/engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+  public:
+    Semaphore() = default;
+
+    void wait() {
+        std::unique_lock<std::mutex> lock{mutex_};
+        cond_.wait(lock, [this]() { return !active_; });
+        active_ = true;
+    }
+
+    void signal() {
```
hellais: <Bikeshed>
I generally prefer the naming of `acquire` and `release` for semaphore (and find it more commonly used, for ex `golang`, `twisted`, etc.), but I don't have a strong opinion on this.
</Bikeshed>

src/libmeasurement_kit/engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+Task::Task(nlohmann::json &&settings) {
+    pimpl_ = std::make_unique<TaskImpl>();
+    std::promise<void> barrier;
+    std::future<void> started = barrier.get_future();
+    pimpl_->thread = std::thread([this, &barrier,
+                                         settings = std::move(
+                                                 settings)]() mutable {
+        pimpl_->running = true;
+        barrier.set_value();
+        static Semaphore semaphore;
```
hellais: If I understand this correctly, the key here is that the `Semaphore` is static and therefore it will be "shared" by every instance of `Task:Task`, as such it's going to be useful to "protect" the critical section. Is this true?

src/libmeasurement_kit/engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+        semaphore.wait(); // block until a previous task has finished running
+        task_run(pimpl_.get(), settings);
+        pimpl_->running = false;
+        pimpl_->cond.notify_all(); // tell the readers we're done
+        semaphore.signal();        // allow another task to run
+    });
+    started.wait(); // guarantee Task() completes when the thread is running
```
hellais: See my bikeshed comment above, but I think it's a bit confusing to have the semaphore and a `future` have similar interface. I had to grep through the code to be sure this was not another semaphore.

src/libmeasurement_kit/engine.cpp
bassosimone: Yes. Only one thread at a time can be running a test [if I did my homework correctly].

src/libmeasurement_kit/engine.cpp
bassosimone: I didn't consider this. I think changing the names of the `Semaphore` can make the code more easy to read, so I will do that. Thanks for pointing out this.

src/libmeasurement_kit/engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+    // Actually emit the event.
+    {
+        std::unique_lock<std::mutex> _{pimpl->mutex};
+        pimpl->deque.push_back(std::move(event));
+    }
+    pimpl->cond.notify_all(); // more efficient if unlocked
+}
+
+Task::Task(nlohmann::json &&settings) {
+    pimpl_ = std::make_unique<TaskImpl>();
```
hellais: Whenever I read this I can't stop myself from thinking of:

bassosimone: LOL

bassosimone: I didn't know that `pimpl` had a meaning in English, I thought it was just an acronym for "pointer to the implementation". I can perhaps `s/pimpl_/impl_/g` as now I find it disgusting.

src/libmeasurement_kit/engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+        semaphore.signal();        // allow another task to run
+    });
+    started.wait(); // guarantee Task() completes when the thread is running
+}
+
+bool Task::is_running() const {
+    return pimpl_->running; // is an atomic var
+}
+
+void Task::interrupt() {
```
hellais: From looking at the code it seems like the `interrupt` method is only something that is callable before the test starts, as such it's doesn't seem possible to interrupt a test mid-way. Or am I missing something?

bassosimone: When the test is running, it's running on `pimpl_->reactor`. Calling `pimpl_->reactor->stop()` should cause the reactor to exit and so `run_with_initial_event()` should return and the code continue from there.

hellais: Ah ok I understand now. I guess I was confused by there being some `begin` and `end` methods being called later, but I assume that is something that tis managed by the reactor.

bassosimone: Yes, correct! I am planning to remove these `begin` and `end` methods as part of future refactoring as they're named most like this for historical reasons.

src/libmeasurement_kit/ffi.cpp
```diff
@@ -0,0 +1,99 @@
...
+    return mk_task_start_ex(settings, nullptr, -1);
+}
+
+static void safe_copy(const char *reason, char *errbuf, size_t errbuf_siz) {
+    if (errbuf != nullptr && errbuf_siz > 0) {
+        size_t size = strlen(reason);
+        if (size > SIZE_MAX - 1) {
+            return;
+        }
+        size += 1; // account for the final '\0'
```
hellais: Maybe it's cleaner (and more readable) to increase the size and then do the `if` statement. This way you can avoid adding and removing the - 1.

Though perhaps you are doing that to avoid int overflow in `size`.

bassosimone: > Though perhaps you are doing that to avoid int overflow in size.

Yes, indeed!

src/libmeasurement_kit/ffi.cpp
```diff
@@ -0,0 +1,99 @@
...
+        if (size < errbuf_siz) {
+            memcpy(errbuf, reason, size);
+            return;
+        }
+        // deal with truncation
+        memcpy(errbuf, reason, errbuf_siz);
+        errbuf[errbuf_siz - 1] = '\0'; // zero terminate
```
hellais: Does this mean that if you have a target buffer that is smaller than what you copy into it, you truncate it to the maximum allowed length?

I think it's probably better to handle this as an unexpected behaviour and do some sort of error handling.

bassosimone: Yes, I am truncating it. What sort of error handling do you suggest?

hellais: I am not sure, but I think it's better to make it clear that we failed to write to the error buffer (maybe by writing in it "FAIL" or "ERR"), than truncating the error message (which is more likely to go undetected).

bassosimone: @hellais I understand. So, for now I am going to document as a know limitation of this API that the error will be silently truncated in case of too-short buffer. I am however going to either open a small PR just to address that or a ticket. I do not think this issue is actually high priority, because in the end our primary use case is that of a JSON generated by a real JSON serializer and not manually, so I do not expect that there will be many cases where parsing fails.

src/libmeasurement_kit/engine.cpp
bassosimone: OTOH is a super used C++ idiom, so it makes sense to keep the name in place.

src/libmeasurement_kit/engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+static void emit_settings_warning(TaskImpl *pimpl, const char *reason) {
+    emit(pimpl, make_log_event(MK_LOG_WARNING, reason));
+}
+
+static bool validate_known_settings_shallow(
+        TaskImpl *pimpl, const nlohmann::json &settings) {
+    auto rv = true;
+
+#define VALIDATE(name, type, mandatory)                                        \
```
pboothe: This is either a good argument for or a good argument against the use of macros.

Which is to say: this macro usage is getting complex, and while I think this code is correct, it is worryingly brittle. There's likely not a good way to work around it here, but please start a long-running process imagining a way to accomplish the same goal without using macros-in-macros.

Again: it's okay here for now. But let's try and figure out a better/simpler/clearer way. Ideally one that does not use macros at all.

src/libmeasurement_kit/engine.cpp
```diff
@@ -0,0 +1,480 @@
...
+#include "src/libmeasurement_kit/nettests/runnable.hpp"
+
+namespace mk {
+namespace engine {
+
+// # Multi-thread stuff
+//
+// Comes first because it needs more careful handling.
+
+class Semaphore {
```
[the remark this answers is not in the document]
bassosimone: Would this implementation be better? (My understanding is that a semaphore with only one active thread at a time is actually equivalent to a mutex):

```
diff --git a/src/libmeasurement_kit/engine.cpp b/src/libmeasurement_kit/engine.cpp
index 07f9604d..aebe1820 100644
--- a/src/libmeasurement_kit/engine.cpp
+++ b/src/libmeasurement_kit/engine.cpp
@@ -42,25 +42,18 @@ class Semaphore {
   public:
     Semaphore() = default;
 
-    void acquire() {
-        std::unique_lock<std::mutex> lock{mutex_};
-        cond_.wait(lock, [this]() { return !active_; });
-        active_ = true;
-    }
+    void acquire() { mutex_.lock(); }
 
-    void release() {
-        {
-            std::unique_lock<std::mutex> _{mutex_};
-            active_ = false;
-        }
-        cond_.notify_one();
-    }
+    void release() { mutex_.unlock(); }
+
+    Semaphore(const Semaphore &) = delete;
+    Semaphore &operator=(const Semaphore &) = delete;
+    Semaphore(Semaphore &&) = delete;
+    Semaphore &operator=(Semaphore &&) = delete;
 
     ~Semaphore() = default;
 
   private:
-    bool active_ = false;
-    std::condition_variable cond_;
     std::mutex mutex_;
 };
 
```

src/libmeasurement_kit/engine.cpp
bassosimone: > But let's try and figure out a better/simpler/clearer way. Ideally one that does not use macros at all.

So, an alternative possibility would be to specify in a JSON/YAML file the structure of the messages and then automatically generate code to process them. I have already done some private experiments where I was using YAML, Python, and jinja2 to implement this.

include/measurement_kit/swig.hpp
```diff
@@ -0,0 +1,167 @@
...
+                mk_event_destroy(p);
+            }
+        }
+    };
+#endif
+
+    /// Task() creates an empty task.
+    Task() {}
+
+    /// initialize() initializes a Task with @p settings. @param settings is
```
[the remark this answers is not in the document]
bassosimone: Thanks, it didn't occur to me that this could happen! So, I just checked with Doxygen and it seems that it is actually able to handle this case:

screen shot 2018-02-27 at 12 08 07

src/libmeasurement_kit/engine.cpp
bassosimone: @pboothe does the above proposal (Python and jinja2) sounds an improvement over X-macros to you?

src/libmeasurement_kit/engine.cpp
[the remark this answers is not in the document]
bassosimone: (But, I wonder whether this will preserve ordering, I should check. Otherwise, another possible implementation is probably one where there is a singleton thread with a work queue. In the latter case, we will still use a condition variable, a mutex, a queue. Also, in such case, I think - but need to check - that we may need to register an atexit handler to quit the thread.)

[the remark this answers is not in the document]
bassosimone: > But, I wonder whether this will preserve ordering, I should check.

`Notify_one()` does not guarantee that order is preserved. The documentation says ["unblocks one of the waiting threads"](http://en.cppreference.com/w/cpp/thread/condition_variable/notify_one). So, it does not matter whether using a mutex preserves ordering or not, because we were not preserving it already.

Do we want to enforce strict ordering at this level? Probably, we don't.

Rationale:

1. I would say that, if an app wants to run tests in a _specific_ order, it should run them in a specific order. Since the API is _blocking_, this is the most natural outcome you get if you use a single worker thread for running tests. Running tests in parallel, entails more work. It would require you to spawn several threads and to arrange to concurrently update the UI for many tests. The current OONI probe for {iOS,Android} does that. But the new version will not, since we do not want to encourage users to _attempt_ to run tests in parallel.

2. The way I see it, this `Semaphore` is just an internal safeguard to prevent the app developer from mis-using the library. As such, I would say it's okay to make its implementation _as simple as possible_ (hence a single mutex). We probably want to specify in the documentation that, in the event in which one attempts to run tests in parallel, they will be serialised, and the order will not be preserved. Do we also want to issue a warning in such case?

- - -

BTW, I also checked whether perhaps, _at the implementation level_, we can rely on the system dependent code to actually enforce FIFO behaviour. To this end, I looked into the sources of [llvm-mirror/libcxx](https://github.com/llvm-mirror/libcxx). Long story short: on Windows a SRW lock is used, which [does not preserve order](https://msdn.microsoft.com/en-us/library/windows/desktop/aa904937(v=vs.85).aspx). On Unix, pthread_mutex_unlock is used, where [the order in which the lock is released depends on the scheduling policy](http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_mutex_lock.html), which is not totally clear to me (I think it means that it depends on the kernel scheduler, tho, since there is no reference to any other man page).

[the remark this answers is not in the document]
bassosimone: So, I also pinged @lorenzoPrimi and @hellais on this. It's not clear whether we want to enforce FIFO order or not. But, in any case, it seems to me that, if we don't want FIFO order, then the implementation using a single mutex is functionally equivalent to the existing one and simpler.

src/libmeasurement_kit/engine.cpp
bassosimone: So, I just had a video chat with @pboothe where he suggests to avoid code generation and instead see whether I can refactor the code to use a factory approach to validate incoming and outgoing messages. The main issue seems to be combining X-macros with long macros with side effects. Perhaps we can combine X-macros with a factory-like code, so that we get more precise type checking and the overall code is more robust to refactoring.

PR #1557, opened 2018-02-09
Sketch out synchronous HTTP API

bassosimone:
In recent discussions we realized that some parts of MK can be
synchronous, because they can also run in threads, since we are
not using thousands of connections, so threads will do.

This would easy contributing to MK and reviewing others' patches.

Since most of these parts deal with interactions with backend
servers, and since this interaction mostly involves HTTP, I
come to the conclusion that the lowest-effort path to improve
the situation is to allow for synchronous HTTP requests.

This diff adds a simple wrapper around the existing HTTP engine, so
we can use the current code also to make sync requests.


bassosimone:
This is actually not the correct approach. Changing the code in this ways leads to too many things to keep track of. It's actually better to write synchronous blocks of code and then replace them with their asynchronous counterpart one by one.

PR #1558, opened 2018-02-12
fix(ooni/templates.cpp): no TTL w/ system DNS

### LINKED ISSUE #1556: DNS engine=system reports TTL=0
TTL=0 is a valid TTL value for DNS packet. Yes, [it's bad idea to have TTL=0](http://mark.lindsey.name/2009/03/never-use-dns-ttl-of-zero-0.html), but seems [there are packets like that](https://00f.net/2011/11/17/how-long-does-a-dns-ttl-last/). I think, the statement may be verified if it matters.

Anyway, there are reports claiming TTL=0 for engine=system while this engine does not support TTL at all. Affected version are 0.7.6 and 0.8.1.
```
  "test_keys": {
    "queries": [
      {
        "engine": "system",
        "resolver_hostname": null,
        "query_type": "A",
        "hostname": "www.rutracker.org",
        "answers": [
          {
            "ipv4": "88.208.38.202",
            "answer_type": "A",
            "ttl": 0
          }
        ],
        "failure": null,
        "resolver_port": null
      }
    ],
  },
```

I suggest to omit `ttl` field if it's not known. Having a sentinel value is a bad idea as any UINT32 value is a valid TTL value (but resolvers may enforce min/max TTL). Try `dig $(( 0xffffffff )).42.rnd.darkk.net.ru @rnd-zone.darkk.net.ru` :)

bassosimone:
Example report snippet from Web Connectivity of www.google.com:

```JSON
    {
        "queries": [
            {
                "answers": [
                    {
                        "answer_type": "CNAME",
                        "hostname": "www.google.com",
                        "ttl": null
                    },
                    {
                        "answer_type": "A",
                        "ipv4": "216.58.201.196",
                        "ttl": null
                    }
                ],
                "engine": "system",
                "failure": null,
                "hostname": "www.google.com",
                "query_type": "A",
                "resolver_hostname": null,
                "resolver_port": null
            }
        ]
    }
```

Closes #1556. To also be backported into the stable branch.


bassosimone:
@darkk pointed out that it would probably be better to add a test checking that `null` is actually emitted by this code. I will do that and, if the test passes, proceed with merging.

PR #1562, opened 2018-02-22
fix(ffi.cpp): tell user about start errors

bassosimone:
It was recommended to do something towards this direction
by @hellais in #1555.


test/ffi/start.cpp
```diff
@@ -0,0 +1,41 @@
...
+        REQUIRE(err == MK_TASK_EGENERIC);
+        REQUIRE(task == nullptr);
+    }
+
+    SECTION("With invalid settings") {
+        mk_task_t *task = (mk_task_t *)0x1234;
+        auto err = mk_task_start_ex(&task, "{");
+        REQUIRE(err == MK_TASK_EPARSE);
+        REQUIRE(task == nullptr);
+    }
```
hellais: It may be worthwhile to add a test case for valid settings too.

bassosimone: Right. I actually didn't add this here because the common case is already tested by integration tests, see e.g. [test/ffi/ndt_example.cpp](https://github.com/measurement-kit/measurement-kit/blob/master/test/ffi/ndt_example.cpp).

bassosimone:
Thanks, merging!

PR #1571, opened 2018-03-02
Implement emitting events and fix their format

bassosimone:
A rebased version of the branch @hellais is using to develop go-measurement-kit.

This is more of an integration branch than a single diff. The whole diff is still very readable, but the rationale of specific diff lines is more easily grokked by looking at individual commits, I think.


src/libmeasurement_kit/engine.cpp
```diff
@@ -133,10 +134,16 @@ nlohmann::json Task::wait_for_next_event() {
     if (!pimpl_->deque.empty()) {
         auto rv = std::move(pimpl_->deque.front());
         pimpl_->deque.pop_front();
-        return nlohmann::json{std::move(rv)};
+        return rv;
```
bassosimone: I forgot to document this. It's a bugfix: previous code wrapped the event into a vector object.

bassosimone:
For the records, diff wrt the original branch used by @hellais:

```diff
diff --git a/example/engine/.gitignore b/example/engine/.gitignore
index b02d7e24..7ed8ca80 100644
--- a/example/engine/.gitignore
+++ b/example/engine/.gitignore
@@ -1,2 +1,3 @@
 # file generated by './build/gitignore'; do not edit
 /ndt
+/web_connectivity
diff --git a/example/engine/web_connectivity.cpp b/example/engine/web_connectivity.cpp
new file mode 100644
index 00000000..3964c762
--- /dev/null
+++ b/example/engine/web_connectivity.cpp
@@ -0,0 +1,25 @@
+// Public domain 2017, Simone Basso <bassosimone@gmail.com.
+
+#define MK_ENGINE_INTERNALS
+#include <measurement_kit/engine.h>
+
+#include <iostream>
+
+#include <measurement_kit/common/nlohmann/json.hpp>
+
+int main() {
+    nlohmann::json settings{
+        {"inputs", {
+            "https://www.google.com/",
+            "https://ooni.torproject.org/",
+        }},
+        {"name", "WebConnectivity"},
+        {"verbosity", "INFO"},
+    };
+    std::clog << settings.dump() << "\n";
+    mk::engine::Task task{std::move(settings)};
+    while (!task.is_done()) {
+        auto event = task.wait_for_next_event();
+        std::clog << event << "\n";
+    }
+}
diff --git a/example/ffi/.gitignore b/example/ffi/.gitignore
index b02d7e24..7ed8ca80 100644
--- a/example/ffi/.gitignore
+++ b/example/ffi/.gitignore
@@ -1,2 +1,3 @@
 # file generated by './build/gitignore'; do not edit
 /ndt
+/web_connectivity
diff --git a/example/ffi/web_connectivity.cpp b/example/ffi/web_connectivity.cpp
new file mode 100644
index 00000000..8bdb1bc1
--- /dev/null
+++ b/example/ffi/web_connectivity.cpp
@@ -0,0 +1,47 @@
+/* Public domain 2017, Simone Basso <bassosimone@gmail.com. */
+
+#include <measurement_kit/ffi.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+/*
+ * Ideally it would be nice to have this written in C. For this to work,
+ * however, we would need to specifically compile Measurement Kit to link
+ * statically with its own version of libc++.
+ */
+int main() {
+    mk_task_t *task = nullptr;
+    mk_task_error_t err =
+            mk_task_start_ex(&task, "{\n"
+                                    "    \"inputs\": [\n"
+                                    "        \"https://www.google.com\","
+                                    "        \"https://www.x.org\""
+                                    "    ],\n"
+                                    "    \"name\": \"WebConnectivity\",\n"
+                                    "    \"verbosity\": \"INFO\"\n"
+                                    "}\n");
+    if (err != MK_TASK_ENONE) {
+        fprintf(stderr, "ERROR: cannot create/start task: %d\n", err);
+        exit(1);
+    }
+
+    while (!mk_task_is_done(task)) {
+        mk_event_t *event = mk_task_wait_for_next_event(task);
+        if (event == NULL) {
+            fprintf(stderr, "ERROR: cannot wait for next event\n");
+            exit(1);
+        }
+        const char *serio = mk_event_serialize(event);
+        if (serio == NULL) {
+            fprintf(stderr, "ERROR: cannot serialize event\n");
+            exit(1);
+        }
+        printf("%s\n", serio);
+        mk_event_destroy(event);
+    }
+
+    mk_task_destroy(task);
+    return 0;
+}
diff --git a/example/swig/.gitignore b/example/swig/.gitignore
index b02d7e24..7ed8ca80 100644
--- a/example/swig/.gitignore
+++ b/example/swig/.gitignore
@@ -1,2 +1,3 @@
 # file generated by './build/gitignore'; do not edit
 /ndt
+/web_connectivity
diff --git a/example/swig/web_connectivity.cpp b/example/swig/web_connectivity.cpp
new file mode 100644
index 00000000..cff03376
--- /dev/null
+++ b/example/swig/web_connectivity.cpp
@@ -0,0 +1,28 @@
+// Public domain 2017, Simone Basso <bassosimone@gmail.com.
+
+#include <measurement_kit/swig.hpp>
+
+#include <stdlib.h>
+
+#include <iostream>
+
+int main() {
+    std::string settings = R"({
+        "inputs": [
+            "https://slashdot.org/",
+            "http://www.microsoft.com"
+        ],
+        "name": "WebConnectivity",
+        "verbosity": "INFO"
+    })";
+    mk::swig::Task task;
+    auto rv = task.initialize_ex(settings);
+    if (!rv.result) {
+        std::clog << "ERROR: " << rv.reason << std::endl;
+        exit(1);
+    }
+    while (!task.is_done()) {
+        std::string event = task.wait_for_next_event();
+        std::clog << event << "\n";
+    }
+}
diff --git a/src/libmeasurement_kit/common/logger.cpp b/src/libmeasurement_kit/common/logger.cpp
index b26093bb..19dbec1c 100644
--- a/src/libmeasurement_kit/common/logger.cpp
+++ b/src/libmeasurement_kit/common/logger.cpp
@@ -235,6 +235,13 @@ class DefaultLogger : public Logger, public NonCopyable, public NonMovable {
                 /* Suppress */;
             }
         }
+        assert(!!s);
+        // Note that the mutex is recursive
+        // TODO(bassosimone): improve the API to allow emitting more context
+        emit_event_ex("status.progress", {
+            {"percentage", progress_offset_ + progress_relative_},
+            {"message", s},
+        });
     }
 
     void set_progress_offset(double offset) override {
diff --git a/src/libmeasurement_kit/nettests/captive_portal.cpp b/src/libmeasurement_kit/nettests/captive_portal.cpp
index 12796bf5..64c4cb90 100644
--- a/src/libmeasurement_kit/nettests/captive_portal.cpp
+++ b/src/libmeasurement_kit/nettests/captive_portal.cpp
@@ -12,8 +12,11 @@ namespace nettests {
 
 CaptivePortalTest::CaptivePortalTest() : BaseTest() {
     runnable.reset(new CaptivePortalRunnable);
-    runnable->test_name = "captiveportal";
-    runnable->test_version = "0.4.0";
+}
+
+CaptivePortalRunnable::CaptivePortalRunnable() noexcept {
+    test_name = "captiveportal";
+    test_version = "0.4.0";
 }
 
 void CaptivePortalRunnable::main(std::string input, Settings options,
diff --git a/src/libmeasurement_kit/nettests/dash.cpp b/src/libmeasurement_kit/nettests/dash.cpp
index 0a2bc41d..3e858500 100644
--- a/src/libmeasurement_kit/nettests/dash.cpp
+++ b/src/libmeasurement_kit/nettests/dash.cpp
@@ -11,9 +11,12 @@ namespace nettests {
 
 DashTest::DashTest() : BaseTest() {
     runnable.reset(new DashRunnable);
-    runnable->test_name = "dash";
-    runnable->test_version = "0.7.0";
-    runnable->needs_input = false;
+}
+
+DashRunnable::DashRunnable() noexcept {
+    test_name = "dash";
+    test_version = "0.7.0";
+    needs_input = false;
 }
 
 void DashRunnable::main(std::string /*input*/, Settings options,
diff --git a/src/libmeasurement_kit/nettests/dns_injection.cpp b/src/libmeasurement_kit/nettests/dns_injection.cpp
index 001c72a1..600616bb 100644
--- a/src/libmeasurement_kit/nettests/dns_injection.cpp
+++ b/src/libmeasurement_kit/nettests/dns_injection.cpp
@@ -12,9 +12,12 @@ namespace nettests {
 
 DnsInjectionTest::DnsInjectionTest() : BaseTest() {
     runnable.reset(new DnsInjectionRunnable);
-    runnable->test_name = "dns_injection";
-    runnable->test_version = "0.0.1";
-    runnable->needs_input = true;
+}
+
+DnsInjectionRunnable::DnsInjectionRunnable() noexcept {
+    test_name = "dns_injection";
+    test_version = "0.0.1";
+    needs_input = true;
 }
 
 void DnsInjectionRunnable::main(std::string input, Settings options,
diff --git a/src/libmeasurement_kit/nettests/facebook_messenger.cpp b/src/libmeasurement_kit/nettests/facebook_messenger.cpp
index f9e9582b..64a51508 100644
--- a/src/libmeasurement_kit/nettests/facebook_messenger.cpp
+++ b/src/libmeasurement_kit/nettests/facebook_messenger.cpp
@@ -12,9 +12,12 @@ namespace nettests {
 
 FacebookMessengerTest::FacebookMessengerTest() : BaseTest() {
     runnable.reset(new FacebookMessengerRunnable);
-    runnable->test_name = "facebook_messenger";
-    runnable->test_version = "0.0.2";
-    runnable->needs_input = false;
+}
+
+FacebookMessengerRunnable::FacebookMessengerRunnable() noexcept {
+    test_name = "facebook_messenger";
+    test_version = "0.0.2";
+    needs_input = false;
 }
 
 void FacebookMessengerRunnable::main(std::string /*input*/, Settings options,
diff --git a/src/libmeasurement_kit/nettests/http_header_field_manipulation.cpp b/src/libmeasurement_kit/nettests/http_header_field_manipulation.cpp
index 09dbfdec..db9de9c0 100644
--- a/src/libmeasurement_kit/nettests/http_header_field_manipulation.cpp
+++ b/src/libmeasurement_kit/nettests/http_header_field_manipulation.cpp
@@ -12,10 +12,14 @@ namespace nettests {
 
 HttpHeaderFieldManipulationTest::HttpHeaderFieldManipulationTest() : BaseTest() {
     runnable.reset(new HttpHeaderFieldManipulationRunnable);
-    runnable->test_name = "http_header_field_manipulation";
-    runnable->test_version = "0.0.1";
-    runnable->needs_input = false;
-    runnable->test_helpers_data = {{"http-return-json-headers", "backend"}};
+}
+
+HttpHeaderFieldManipulationRunnable::HttpHeaderFieldManipulationRunnable()
+        noexcept {
+    test_name = "http_header_field_manipulation";
+    test_version = "0.0.1";
+    needs_input = false;
+    test_helpers_data = {{"http-return-json-headers", "backend"}};
 }
 
 void HttpHeaderFieldManipulationRunnable::main(std::string input,
diff --git a/src/libmeasurement_kit/nettests/http_invalid_request_line.cpp b/src/libmeasurement_kit/nettests/http_invalid_request_line.cpp
index 946ee23a..8365dc55 100644
--- a/src/libmeasurement_kit/nettests/http_invalid_request_line.cpp
+++ b/src/libmeasurement_kit/nettests/http_invalid_request_line.cpp
@@ -12,9 +12,12 @@ namespace nettests {
 
 HttpInvalidRequestLineTest::HttpInvalidRequestLineTest() : BaseTest() {
     runnable.reset(new HttpInvalidRequestLineRunnable);
-    runnable->test_name = "http_invalid_request_line";
-    runnable->test_version = "0.0.3";
-    runnable->test_helpers_data = {{"tcp-echo", "backend"}};
+}
+
+HttpInvalidRequestLineRunnable::HttpInvalidRequestLineRunnable() noexcept {
+    test_name = "http_invalid_request_line";
+    test_version = "0.0.3";
+    test_helpers_data = {{"tcp-echo", "backend"}};
 }
 
 void HttpInvalidRequestLineRunnable::main(std::string, Settings options,
diff --git a/src/libmeasurement_kit/nettests/meek_fronted_requests.cpp b/src/libmeasurement_kit/nettests/meek_fronted_requests.cpp
index 0e95e73b..a3a1611e 100644
--- a/src/libmeasurement_kit/nettests/meek_fronted_requests.cpp
+++ b/src/libmeasurement_kit/nettests/meek_fronted_requests.cpp
@@ -12,9 +12,12 @@ namespace nettests {
 
 MeekFrontedRequestsTest::MeekFrontedRequestsTest() : BaseTest() {
     runnable.reset(new MeekFrontedRequestsRunnable);
-    runnable->test_name = "meek_fronted_requests";
-    runnable->test_version = "0.0.1";
-    runnable->needs_input = true;
+}
+
+MeekFrontedRequestsRunnable::MeekFrontedRequestsRunnable() noexcept {
+    test_name = "meek_fronted_requests";
+    test_version = "0.0.1";
+    needs_input = true;
 }
 
 void MeekFrontedRequestsRunnable::main(std::string input, Settings options,
diff --git a/src/libmeasurement_kit/nettests/multi_ndt.cpp b/src/libmeasurement_kit/nettests/multi_ndt.cpp
index de32324f..f7a664e1 100644
--- a/src/libmeasurement_kit/nettests/multi_ndt.cpp
+++ b/src/libmeasurement_kit/nettests/multi_ndt.cpp
@@ -14,10 +14,13 @@ namespace nettests {
 
 MultiNdtTest::MultiNdtTest() : BaseTest() {
     runnable.reset(new MultiNdtRunnable);
-    runnable->options["save_real_probe_ip"] = true;
-    runnable->options["dns/engine"] = "system";
-    runnable->test_name = "multi_ndt";
-    runnable->test_version = "0.1.0";  /* Forked from `ndt` v0.0.4 */
+}
+
+MultiNdtRunnable::MultiNdtRunnable() noexcept {
+    options["save_real_probe_ip"] = true;
+    options["dns/engine"] = "system";
+    test_name = "multi_ndt";
+    test_version = "0.1.0";  /* Forked from `ndt` v0.0.4 */
 }
 
 static void write_simple_stats(report::Entry &entry, SharedPtr<Logger> logger) {
diff --git a/src/libmeasurement_kit/nettests/ndt.cpp b/src/libmeasurement_kit/nettests/ndt.cpp
index 6a095857..7db138a4 100644
--- a/src/libmeasurement_kit/nettests/ndt.cpp
+++ b/src/libmeasurement_kit/nettests/ndt.cpp
@@ -14,8 +14,11 @@ namespace nettests {
 
 NdtTest::NdtTest() : BaseTest() {
     runnable.reset(new NdtRunnable);
-    runnable->test_name = "ndt";
-    runnable->test_version = "0.1.0";
+}
+
+NdtRunnable::NdtRunnable() noexcept {
+    test_name = "ndt";
+    test_version = "0.1.0";
 }
 
 void NdtRunnable::main(std::string, Settings settings,
diff --git a/src/libmeasurement_kit/nettests/runnable.hpp b/src/libmeasurement_kit/nettests/runnable.hpp
index 257666fd..d0c64c59 100644
--- a/src/libmeasurement_kit/nettests/runnable.hpp
+++ b/src/libmeasurement_kit/nettests/runnable.hpp
@@ -71,8 +71,10 @@ class Runnable : public NonCopyable, public NonMovable {
 #define MK_DECLARE_RUNNABLE(_name_)                                            \
     class _name_ : public Runnable {                                           \
       public:                                                                  \
+        _name_() noexcept;                                                     \
+                                                                               \
         void main(std::string, Settings,                                       \
-                  Callback<SharedPtr<report::Entry>>) override;                      \
+                  Callback<SharedPtr<report::Entry>>) override;                \
     }
 
 MK_DECLARE_RUNNABLE(DashRunnable);
@@ -91,7 +93,9 @@ MK_DECLARE_RUNNABLE(WhatsappRunnable);
 // Separate definition because it contains extra methods
 class WebConnectivityRunnable : public Runnable {
   public:
-    void main(std::string, Settings, Callback<SharedPtr<report::Entry>>) override;
+    WebConnectivityRunnable() noexcept;
+    void main(
+            std::string, Settings, Callback<SharedPtr<report::Entry>>) override;
     void fixup_entry(report::Entry &) override;
 };
 
diff --git a/src/libmeasurement_kit/nettests/tcp_connect.cpp b/src/libmeasurement_kit/nettests/tcp_connect.cpp
index e7de8163..cd717798 100644
--- a/src/libmeasurement_kit/nettests/tcp_connect.cpp
+++ b/src/libmeasurement_kit/nettests/tcp_connect.cpp
@@ -12,9 +12,12 @@ namespace nettests {
 
 TcpConnectTest::TcpConnectTest() : BaseTest() {
     runnable.reset(new TcpConnectRunnable);
-    runnable->test_name = "tcp_connect";
-    runnable->test_version = "0.1.0";
-    runnable->needs_input = true;
+}
+
+TcpConnectRunnable::TcpConnectRunnable() noexcept {
+    test_name = "tcp_connect";
+    test_version = "0.1.0";
+    needs_input = true;
 }
 
 void TcpConnectRunnable::main(std::string input, Settings options,
diff --git a/src/libmeasurement_kit/nettests/telegram.cpp b/src/libmeasurement_kit/nettests/telegram.cpp
index 4a686dc0..0a2f19ad 100644
--- a/src/libmeasurement_kit/nettests/telegram.cpp
+++ b/src/libmeasurement_kit/nettests/telegram.cpp
@@ -12,9 +12,12 @@ namespace nettests {
 
 TelegramTest::TelegramTest() : BaseTest() {
     runnable.reset(new TelegramRunnable);
-    runnable->test_name = "telegram";
-    runnable->test_version = "0.0.2";
-    runnable->needs_input = false;
+}
+
+TelegramRunnable::TelegramRunnable() noexcept {
+    test_name = "telegram";
+    test_version = "0.0.2";
+    needs_input = false;
 }
 
 void TelegramRunnable::main(std::string /*input*/, Settings options,
diff --git a/src/libmeasurement_kit/nettests/web_connectivity.cpp b/src/libmeasurement_kit/nettests/web_connectivity.cpp
index d9b09506..eb05247c 100644
--- a/src/libmeasurement_kit/nettests/web_connectivity.cpp
+++ b/src/libmeasurement_kit/nettests/web_connectivity.cpp
@@ -12,10 +12,13 @@ namespace nettests {
 
 WebConnectivityTest::WebConnectivityTest() : BaseTest() {
     runnable.reset(new WebConnectivityRunnable);
-    runnable->test_name = "web_connectivity";
-    runnable->test_version = "0.0.1";
-    runnable->needs_input = true;
-    runnable->test_helpers_data = {{"web-connectivity", "backend"}};
+}
+
+WebConnectivityRunnable::WebConnectivityRunnable() noexcept {
+    test_name = "web_connectivity";
+    test_version = "0.0.1";
+    needs_input = true;
+    test_helpers_data = {{"web-connectivity", "backend"}};
 }
 
 void WebConnectivityRunnable::main(std::string input, Settings options,
diff --git a/src/libmeasurement_kit/nettests/whatsapp.cpp b/src/libmeasurement_kit/nettests/whatsapp.cpp
index 79156c44..b8fb8e2a 100644
--- a/src/libmeasurement_kit/nettests/whatsapp.cpp
+++ b/src/libmeasurement_kit/nettests/whatsapp.cpp
@@ -11,9 +11,12 @@ namespace nettests {
 
 WhatsappTest::WhatsappTest() : BaseTest() {
     runnable.reset(new WhatsappRunnable);
-    runnable->test_name = "whatsapp";
-    runnable->test_version = "0.6.1";
-    runnable->needs_input = false;
+}
+
+WhatsappRunnable::WhatsappRunnable() noexcept {
+    test_name = "whatsapp";
+    test_version = "0.6.1";
+    needs_input = false;
 }
 
 void WhatsappRunnable::main(std::string /*input*/, Settings options,
diff --git a/test/engine/.gitignore b/test/engine/.gitignore
index 130c8bcb..203674ae 100644
--- a/test/engine/.gitignore
+++ b/test/engine/.gitignore
@@ -1,2 +1,3 @@
 # file generated by './build/gitignore'; do not edit
 /ndt_example
+/web_connectivity_example
diff --git a/test/engine/web_connectivity_example.cpp b/test/engine/web_connectivity_example.cpp
new file mode 100644
index 00000000..d53a88a6
--- /dev/null
+++ b/test/engine/web_connectivity_example.cpp
@@ -0,0 +1,6 @@
+// We want to run the unmodified example through Valgrind
+#ifdef ENABLE_INTEGRATION_TESTS
+#include "example/engine/web_connectivity.cpp"
+#else
+int main(){}
+#endif
diff --git a/test/ffi/.gitignore b/test/ffi/.gitignore
index ec1a0e0c..0bc270cf 100644
--- a/test/ffi/.gitignore
+++ b/test/ffi/.gitignore
@@ -1,3 +1,4 @@
 # file generated by './build/gitignore'; do not edit
 /ndt_example
 /start
+/web_connectivity_example
diff --git a/test/ffi/web_connectivity_example.cpp b/test/ffi/web_connectivity_example.cpp
new file mode 100644
index 00000000..1c075031
--- /dev/null
+++ b/test/ffi/web_connectivity_example.cpp
@@ -0,0 +1,6 @@
+// We want to run the unmodified example through Valgrind
+#ifdef ENABLE_INTEGRATION_TESTS
+#include "example/ffi/web_connectivity.cpp"
+#else
+int main(){}
+#endif
diff --git a/test/swig/.gitignore b/test/swig/.gitignore
index c3c0afec..31df9048 100644
--- a/test/swig/.gitignore
+++ b/test/swig/.gitignore
@@ -1,3 +1,4 @@
 # file generated by './build/gitignore'; do not edit
 /initialize
 /ndt_example
+/web_connectivity_example
diff --git a/test/swig/web_connectivity_example.cpp b/test/swig/web_connectivity_example.cpp
new file mode 100644
index 00000000..7ee34b0e
--- /dev/null
+++ b/test/swig/web_connectivity_example.cpp
@@ -0,0 +1,6 @@
+// We want to run the unmodified example through Valgrind
+#ifdef ENABLE_INTEGRATION_TESTS
+#include "example/swig/web_connectivity.cpp"
+#else
+int main(){}
+#endif

```

willscott:
This looks reasonable, and the `is_done` api is used by the measurement-kit-go interface. It would be good to get this merged so that the golang code path doesn't end up relying on an unmerged API a couple dependencies down.

src/libmeasurement_kit/engine.cpp
```diff
@@ -104,17 +100,22 @@ Task::Task(nlohmann::json &&settings) {
         pimpl_->running = true;
         barrier.set_value();
         static Semaphore semaphore;
-        semaphore.acquire(); // block until a previous task has finished running
-        task_run(pimpl_.get(), settings);
+        task_run(pimpl_.get(), settings, [&]() {
+            semaphore.acquire(); // block until a previous task has finished
```
willscott: my only comment on this is that there's a lot of complexity in the concurrency here. semaphores, barriers, lambdas, waiters. 

a bit more comments explaining the state machine, or some tests to try to limit the potential of race conditions entering this would be useful

bassosimone:
@willscott @hellais comments addressed, rebased, and pushed!

bassosimone:
So, it seems some tests are failing for 267dca4. This seems to be more related with the overall build system than it is to this branch. Regardless, I'm investigating in #1575.

src/libmeasurement_kit/common/logger.cpp
```diff
@@ -188,6 +195,34 @@ class DefaultLogger : public Logger, public NonCopyable, public NonMovable {
...
+        nlohmann::json event{
+            {"key", key},
+            {"value", std::move(value)}
+        };
+        std::unique_lock<std::recursive_mutex> _{mutex_};
+        // TODO(bassosimone): other logging functions filter all the
+        // exceptions. We cannot change this behavior until that is part
+        // of our public API. But here we deliberately choose not to do
+        // any exception handling. The callee must behave.
+        handlers_.at(key)(std::move(event));
```
pboothe: `.at()` throws when the key is missing. Is that the desired behavior?

example/ffi/ndt.cpp
```diff
@@ -15,15 +15,15 @@ int main() {
     mk_task_t *task = nullptr;
     mk_task_error_t err =
             mk_task_start_ex(&task, "{\n"
```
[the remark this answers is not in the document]
bassosimone: The reason why I don't use raw literals in that file is that I am using the `ffi` API and ideally I would even want to write that file in C to show you can really get away with a simple API. What prevents me from doing so is that I cannot force statically linking libc++ on macOS. Otherwise, I would really want to have that show case to proof that it is possible.

> But maybe there are object lifetime issues preventing this approach?

I think I am lacking some knowledge about raw literals and life cycles so that I do not understand very well what you mean here. Hence, please could you:

supreme_dalek_gif__by_sallyfriceiii-d7qozkq

?

src/libmeasurement_kit/common/logger.cpp
bassosimone: Throwing is not desired, but the above `handles_.count(key) <= 0` check should guarantee that, if we arrive down this line, there is a registered handler, right?

bassosimone:
Before I can merge this branch, the following should happen:

1. I should have #1575 approved and merged (`master` is currently not running unit tests because of a regression introduced in refactoring the build system and fixed in #1575)

2. I should understand why the code was not throwing at runtime even without 31cbb28 applied

3. I should make sure the event format is documented

example/ffi/ndt.cpp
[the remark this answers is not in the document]
pboothe: `char *` constant typically live in some static allocation. `std::str` variables may or may not. So, if you were to save the passed-in `char *` without copying the string, and it came from a `std::str("fsdfds").c_str()` that was only allocated for that one call, then later use of that pointer could cause problems.

src/libmeasurement_kit/common/logger.cpp
pboothe: Yup! I can't believe I missed that.

example/ffi/ndt.cpp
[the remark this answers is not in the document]
bassosimone: Ah, right! So, since the pointer is used _immediately_ and not stored or use later, I believe C++ guarantees that the temporary `std::str` would be alive for long enough. (Still, as I mentioned above, I'd like that file to be written in C++ close to C as much as possible.)

bassosimone:
Rebased. Checked that there were no differences since before. Going to land this!

Issue #1390, opened 2017-09-26
Make sure we can run NDT over Tor


bassosimone:
It seems this may be potentially of interest, because it allows to measure the end-to-end speed using a well known performance test that provides lots of metrics. Things to consider:

1. make sure we can establish non-encrypted and encrypted sessions over socks5 (this should be fixed soon in master but currently we cannot do ssl over socks5)

2. make sure we can torify every step of the test, so we don't leak information around

As a bonus, doing that means that we know that we can run also web_connectivity and other OONI tests over Tor, which is probably one of the ways in which we'll soon run tests in Desktop apps.

bassosimone:
Cc: @hellais

bassosimone:
I have started working on it in measurement-kit/libndt#15.

bassosimone:
This is implemented in libndt. When we'll migrate to using it, we'll be done.

bassosimone:
This is possible with libndt. I believe this is also possible with ndt7-client if we override the dialer used by gorilla/websocket to connect. No further action required here.

PR #1572, opened 2018-03-02
Start making some public headers private

bassosimone:
Now that we have the new API (see #1555), we can start moving existing API headers in private space, which in turn will enable for more refactoring. The rough plan looks like this:

1. move as much headers as possible in `src/libmeasurement_kit/$module`

2. start using `iwyu` more consistently (this is needed)

3. some header will need to stay public because they are part of the part of current API that is still used by OONi and other integrations, so those will just be deprecated; for these headers, I'll make backward compatible changes until we decide to retire the old API (I already performed large part of this refactoring in a personal branch and so I know just few headers will remain public)

4. perform more cleanups and simplifications (the first aspect to work on is `net` and especially the way in which we establish connections, that now has some bugs like the fact that one cannot really establish a SSL connection using Tor as proxy; see #1390).


lorenzoPrimi:
🐳

PR #1561, opened 2018-02-14
Cleanup build system

bassosimone:
This PR cleans the build system up after #1554.

PR #1573, opened 2018-03-02
Start cleaning up the build system

bassosimone:
After #1554, there is need to perform lots of cleanups on the old, more complex build system. Like for #1572, I have decided to proceed in small steps rather than submitting large diffs. This is the starting point of the cleanup: remove the most obvious duplicate files and modify the old build script so that the new build system is invoked when the new build system supports a specific build.

After this change, the main purposes of the old build system are allowing to get specific sources that we vendor inside MK (e.g. `catch.hpp`) and cross building for Android and iOS. In this regard, I already know that the new build system successfully builds for both. As part of the next steps, I am going to also adapt the cross build scripts to call directly the new build system rather than the old build script, so we will reduce one more use case for such script. Soon, I'll also update dependencies and run a full Android and iOS build, so we can start testing the new build system and new code on devices.

(Patches cherry-picked from the integration branch #1561.)


lorenzoPrimi:
🐳

willscott:
It looks like the remaining dependencies not migrated in this PR are
* Catch
* builtin_ca_bundle
* citrus
* http-parser
* json
* strtonum
* tls_verify

is there a bug tracking this / do you have expectations around their migration?

bassosimone:
> is there a bug tracking this / do you have expectations around their migration?

I have a branch called `cleanup/build_system` where I am incrementally cleaning up stuff. It also corresponds to a PR (#1561), but I am actually using it more as an integration branch.

willscott:
just worth having a reasonable sense of when it's okay to remove the
'deprecated' code path entirely. if you suspect that the larger cleanup
branch will be resolved in relatively short order, that seems like
sufficient process.

bassosimone:
> if you suspect that the larger cleanup branch will be resolved in relatively short order, that seems like 
sufficient process

Yeah, I think so: this is one of the tasks I am working on!

PR #1574, opened 2018-03-06
fix(nettests.hpp): repair JNI bindings

bassosimone:
1. I deprecated and then removed this API to clean up

2. but actually this was used by JNI bindings

3. so, readd it

4. do not bother with deprecation because it only creates
   more annoyances when compiling Java bindings


bassosimone:
@willscott: "Yeah, looks fine". Going to merge as soon as travis completes. 🐳

PR #1575, opened 2018-03-15
Fix travis, coveralls, and broken tests

bassosimone:
# Objectives

- make sure we build and run tests

- make sure we submit coverage through coveralls

- repair a broken test making sure we do not leak IP addresses

# Implementation

- do not use the rule for cross-compiling MK also for running tests, as this is too fragile

- <strike>comment-out the broken test (we have been blacklisted by whatismyipaddress.com, so I've commented out the test; I wonder what would be the proper fix to this)</strike> fixed by replacing an integreation test where it was difficult to find an IP leaking service that did not ban us [or behave weirdly](https://travis-ci.org/measurement-kit/measurement-kit/jobs/355558210#L1854) with an unit test where we control input and output


build/docker/script/run
```diff
@@ -26,7 +26,14 @@ subr_start_over() {
...
 subr_build() {
-    ./build/dependency mk
+    ./autogen.sh
+    ./configure --disable-shared $CONFIGURE_FLAGS
+    make V=0 -j$(nproc) $MAKE_CHECK_TARGET || {
+        if [ -f ./test-suite.log ]; then
+            cat ./test-suite.log
+        fi
+        exit 1
+    }
```
bassosimone: It was backwards to use the script for building `mk` on Android/iOS to drive CI. Since changes in the former may lead to changes in the latter, like in this case. Rather, just roll out the command line for testing MK.

build/docker/spec/coveralls
```diff
@@ -19,6 +16,9 @@ debian_deps="$debian_deps python-pip"
 debian_deps="$debian_deps wget"
 pip_deps="$pip_deps cpp-coveralls"
 
+export CONFIGURE_FLAGS="--enable-coverage"
+export MAKE_CHECK_TARGET="check"
+
```
bassosimone: Style: use ALL_UPPERCASE for overridable variables. Stop assuming `pkg_configure_flags` can be passed around as that is not what is going to happen in this specific case. Reduces complexity.

build/docker/script/setup
```diff
@@ -10,8 +10,6 @@ fi
 shift
 
 docker run                                                                 \
-    -e "pkg_configure_flags=$pkg_configure_flags"                          \
-    -e "pkg_make_flags=$pkg_make_flags"                                    \
```
bassosimone: Variables not passed around. So don't forward them. Reduces complexity.

build/docker/spec/coveralls
```diff
@@ -1,9 +1,6 @@
 #!/bin/sh
 set -e
 
-export pkg_configure_flags="$pkg_configure_flags --enable-coverage"
-export pkg_make_check_rule="check"
-
```
bassosimone: See below.

build/docker/spec/valgrind-clang
```diff
@@ -7,7 +7,7 @@ export CC="clang"
 export CFLAGS="$CFLAGS -fstack-protector-strong -Wformat -Werror=format-security -O2"
 export CXX="clang++"
 export CXXFLAGS="-stdlib=libc++ $CXXFLAGS -fstack-protector-strong -Wformat -Werror=format-security -O2"
-export pkg_make_check_rule="run-valgrind-docker"
+export MAKE_CHECK_TARGET="run-valgrind-docker"
```
bassosimone: See above.

build/docker/spec/valgrind-gcc
```diff
@@ -7,7 +7,7 @@ export CC="gcc"
 export CFLAGS="$CFLAGS -fstack-protector-strong -Wformat -Werror=format-security -O2"
 export CXX="g++"
 export CXXFLAGS="$CXXFLAGS -fstack-protector-strong -Wformat -Werror=format-security -O2"
-export pkg_make_check_rule="run-valgrind-docker"
+export MAKE_CHECK_TARGET="run-valgrind-docker"
```
bassosimone: See above.

test/fixtures/scrub.txt
```diff
@@ -1 +1 @@
-http://whatismyipaddress.com/
+http://geoip.ubuntu.com/lookup
```
bassosimone: Use the same service we use for getting the user IP. Rationale: the former service is not blocked. In the event in which `geoip.ubuntu.com` blocks travis, it's unclear whether this test breaks, tho.

test/fixtures/scrub.txt
```diff
@@ -1 +1,2 @@
...
+https://ip-info.ff.avast.com/v1/info
```
bassosimone: Replace the originally-used ip-leaking service with other two services. Motivation for the replacement: the one that we're removing has apparently banned github and we need to use other services. One of the two services is also the one we use for detecting the user IP (i.e. geoip.ubuntu.com). However, it also returns a web page rather than a JSON. So, I want basically to test for scrubbing in more cases, and I'm fine with using geoip.ubuntu.com also for testing scrubbing because we are using more than one service for testing scrubbing anyway.

build/docker/script/run
bassosimone: Motivation for the change: rolling out explicitly what to do is much more robust.

src/libmeasurement_kit/ooni/templates.cpp
```diff
@@ -100,103 +95,8 @@ void dns_query(SharedPtr<Entry> entry, dns::QueryType query_type,
 void http_request(SharedPtr<Entry> entry, Settings settings, http::Headers headers,
                   std::string body, Callback<Error, SharedPtr<http::Response>> cb,
                   SharedPtr<Reactor> reactor, SharedPtr<Logger> logger) {
-
```
bassosimone: Implementation moved in `./src/libmeasurement_kit/ooni/templates_impl.hpp` so we can unittest it.

test/nettests/web_connectivity.cpp
```diff
@@ -16,86 +16,6 @@ TEST_CASE("Synchronous web connectivity test") {
                                                    test::nettests::run_test);
 }
 
-TEST_CASE("Make sure that IP address scrubbing works") {
```
bassosimone: Test transformed into an unittest. Too much flapping and weird issues in depending from external services. The goal is to see if we scrub the address when requested. To do so, it suffices to see if we scrub it. No need to use the net.

test/ooni/templates.cpp
```diff
@@ -173,6 +175,73 @@ TEST_CASE("http requests template works as expected") {
...
-#else
-int main(){}
 #endif
+
+static void mocked_request(Settings settings, http::Headers,
+        std::string, Callback<Error, SharedPtr<http::Response>> cb,
+        SharedPtr<Reactor>, SharedPtr<Logger>,
+        SharedPtr<http::Response>, int) {
+    std::string probe_ip = settings.get("real_probe_ip_", std::string{});
+    REQUIRE(probe_ip != "");
```
bassosimone: If we ever change the way in which we pass around the probe_ip, this will break.

bassosimone:
Thanks!

PR #1583, opened 2018-03-23
Feature/cross mingw

hellais:
This is based on top of events-rebased and is a work in progress to add support for cross compilation of mk on macos for windows.


hellais:
Currently stuck due to std::thread support. This could be a solution to it: https://github.com/meganz/mingw-std-threads

bassosimone:
Superseded by #1606.

PR #1585, opened 2018-03-25
Fix test checking for user IP scrubbing

bassosimone:
Backports part of the diff from #1575. While there also use the
Settings::get_noexcept() method rather than the Settings::get() one,
thus avoiding a possible exception if conversion is not possible:

```diff
-    ErrorOr<int> max_redirects = settings.get("http/max_redirects", 0);
+    ErrorOr<int> max_redirects = settings.get_noexcept("http/max_redirects", 0);
```

Strictly speaking, the above change was not part of #1575, since
it was introduced in #1530, but it seems reasonable to me to also
pull this improvement as part of this PR.


bassosimone:
Since this change is just a backport of changes that have already been approved for `master`, I am going to proceed with merging this branch into `stable` once CI terminates.

PR #1591, opened 2018-03-29
Make cross building much simpler

bassosimone:
Put everything into the same directory. This is what existing script except. Rather than patching them to support a new style, better to use the previously existing convention.


bassosimone:
I suppressed `travis-ci/pr` build. One build is enough.

bassosimone:
The builds that I didn't suppress terminated successfully. Merging!

PR #1593, opened 2018-03-29
Improve cmake-based build

bassosimone:
1. use custom docker image to avoid downloading from `apt-get` every time;

2. use `tc` to avoid overloading m-lab servers.


bassosimone:
Merging: the failed build was interrupted by me.

PR #1599, opened 2018-04-04
net: make all headers private

bassosimone:


bassosimone:
Merging. [The only build that did not terminate was stuck while downloading dependencies](https://travis-ci.org/measurement-kit/measurement-kit/jobs/362094904#L642).

PR #1617, opened 2018-05-04
Make src/libmeasurement_kit compile with MSVC

bassosimone:
I am currently building this in a working branch using Ninja, but it
seems simple enough to be merged without thinking whether I do prefer
using Ninja or CMake for Windows (still undecided...)


bassosimone:
Merging since the last commit only updates the changelog.

PR #1618, opened 2018-05-05
Allow users to disable the bouncer

### LINKED ISSUE #1523: Provide option to bypass Ooni bouncer
Measurement Kit should provide an option to the user to bypass the Ooni bouncer.

bassosimone:
Closes #1523.


bassosimone:
I discussed this with @hellais and he suggested that we should make sure that all tests should be able to work also in the event in which the bouncer is down. This PR is perhaps a good place to do it.

PR #1620, opened 2018-05-07
Allow compiling measurement_kit.exe with MSVC

bassosimone:


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 project(MeasurementKit)
-cmake_minimum_required(VERSION 3.0.0)
+cmake_minimum_required(VERSION 2.8.0)
```
bassosimone: I lowered the CMake minimum version because apparently this yields more useful warnings.

cmake/Common.cmake
```diff
@@ -4,10 +4,10 @@ if (NOT MK_ROOT)
     set (MK_ROOT ${CMAKE_SOURCE_DIR})
 endif (NOT MK_ROOT)
 
-set(CMAKE_BUILD_TYPE Debug CACHE STRING "Set the build type")
+set(CMAKE_BUILD_TYPE Release CACHE STRING "Set the build type")
```
bassosimone: I changed the build type because our dependencies are using `Release` only, for now, and not doing that leads us to library mismatches when we're linking.

cmake/Common.cmake
```diff
@@ -4,10 +4,10 @@ if (NOT MK_ROOT)
...
 
 set(CMAKE_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}" CACHE PATH
-    "Where to install MK" FORCE)
+    "Where to install MK")
```
bassosimone: This was a mistake.

cmake/Common.cmake
```diff
@@ -59,25 +59,19 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
...
+elseif(${CMAKE_C_COMPILER_ID} STREQUAL "GNU")
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${MK_UNIX_CFLAGS}")
 endif()
 
-# We compile our dependencies using `/MT`, so we must use `/MT` as well
-string(REPLACE "/MD" "/MT" CMAKE_C_FLAGS_RELEASE ${CMAKE_C_FLAGS_RELEASE})
-string(REPLACE "/MDd" "/MT" CMAKE_C_FLAGS_DEBUG ${CMAKE_C_FLAGS_DEBUG})
-string(REPLACE "/MD" "/MT" CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE})
-string(REPLACE "/MDd" "/MT" CMAKE_CXX_FLAGS_DEBUG ${CMAKE_CXX_FLAGS_DEBUG})
-
```
bassosimone: Not needed. On Windows the right thing to do, now, is to use `/MD` which is gonna be the library of the universal runtime, which we assume to be present on modern systems that do updates.

example/ooni/oorchestrate.cpp
```diff
@@ -1,94 +0,0 @@
...
-
-#include <future>
-
-using namespace mk::ooni::orchestrate;
-using namespace mk::ooni;
-using namespace mk;
-
-#define USAGE "oorchestrate [-v]"
-
-int main(int argc, char **argv) {
```
bassosimone: Removed the example client because (i) it was complicating the Windows build and (ii) we're redoing this API.

test/common/libevent_reactor.cpp
```diff
@@ -49,13 +49,19 @@ TEST_CASE("libevent_init_once") {
 
 extern "C" {
 
-static event_base *event_base_new_fail() { return nullptr; }
+#ifndef _MSC_VER
+#define STATIC static
+#else
+#define STATIC /* Nothing */
+#endif
```
bassosimone: Required to allow MSVC to find the moked function. (It seems in general MSVC works much better when we provide it with a mocked interface rather than just a mocked function.)

PR #1623, opened 2018-05-16
Run tests with MSVC at AppVeyor

bassosimone:
As part of this PR, make the `CMakeLists.txt` file much more correct and fix _most_ regress tests when they run on Windows. I am going to merge this PR _even if_ AppVeyor is going to file for a single test because fixing that single test should be submitted as a separate PR. (I would rather do that separately because I've started doing it here and it was quickly becoming a mess.)

It is important to remark that running unit tests I was able to catch a bug in Win32 orchestration that I previously introduced (<- yeah, as we all know, tests are _really_ useful).

Another follow-up commit will migrate to Catch2 and reorganize the unit tests such that the process of building them is most likely becoming much faster. After that, I will enable support for running tests also using Mingw and appveyor. When all these steps are done, we'll have made a significant step towards the next stable release of Measurement Kit.


bassosimone:
> I am going to merge this PR even if AppVeyor is going to file for a single test because fixing that single test should be submitted as a separate PR.

(But maybe it's better to _open_ that subsequent PR before merging this one.)

bassosimone:
>> I am going to merge this PR even if AppVeyor is going to file for a single test because fixing that single test should be submitted as a separate PR.
> (But maybe it's better to open that subsequent PR before merging this one.)
See #1624

PR #1633, opened 2018-05-24
Fix failing test

bassosimone:
Use more realiable, better maintained endpoints. Discussed with @hellais.


bassosimone:
Issue fixed by nexa.polito.it folks, fix for now unnecessary.

PR #1640, opened 2018-06-07
C++14 API based on FFI and nettests based on C++14

### LINKED ISSUE #1636: Evolving MK C++ API for OONI Probe iOS needs
Yesterday, I had a discussion with @lorenzoPrimi where he voted for extending rather than fully deprecating the `nettests.hpp` API. To this end, here are the steps I should implement:

- [ ] make sure that _only_ deprecated callbacks are marked as such, not the whole class
- [ ] implement more callbacks for the events that are not routed

In implementing more callbacks, I should probably favour duplicating stuff, rather than touching existing callbacks, because that is the cleanest path towards avoiding to break existing APIs.

The end result will be an API (but certainly not ABI) compatible `nettests.hpp` file.

bassosimone:
This PR aims to roll out a new, C++14-friendly API. This API is based on the FFI and is fully inline. I am also rewriting the nettests API to be based on the new C++14 API. This seems to me the smoothest path of migration from nettests to a new C++ API as required for OONI Probe iOS.

I originally postulated that the optimal solution was to extend the nettests API. However, I realised there was a flaw in nettests design, by which the API was not enforcing that, once a test is configured, you must not touch it again, and you can run it only once. For this reason, I decided to write a new C++14 API rather than extending the previous one.


include/measurement_kit/nettests/ffi.hpp
```diff
@@ -96,103 +100,193 @@ class MK_NETTESTS_DEPRECATED BaseTest {
...
+        impl_.set_log_filepath(s);
         return *this;
     }
 
     BaseTest &on_logger_eof(std::function<void()> &&fn) {
-        impl_->logger_eof_cbs.push_back(std::move(fn));
+        impl_.on_status_terminated([fn = std::move(fn)](
+                const cxx14::StatusTerminated &) noexcept {
+            MK_NETTESTS_CALL_AND_SUPPRESS(fn, ());
+        });
```
bassosimone: Note to self: in theory, the `terminated` event could be emitted more than once. In practice, this would only happen when we leave the event loop and then re-enter it. However, given the circumstances, it would probably be best to somehow guarantee that the callback isn't called more than once. In fact, this callback is used by the Java bindings to free resources that are explicitly allocated, so maybe some extra _paranoia_ here is good.

include/measurement_kit/nettests/ffi.hpp
```diff
@@ -204,215 +298,18 @@ class MK_NETTESTS_DEPRECATED BaseTest {
...
-    void run() { run_static(std::move(impl_)); }
-
-    void start(std::function<void()> &&fn) {
-        std::thread thread{[tip = std::move(impl_), fn = std::move(fn)]() {
-            run_static(std::move(tip));
-            fn();
-        }};
-        thread.detach();
+    void run() {
+        (void)cxx14::TaskRunner::global().run(std::move(impl_));
```
bassosimone: Maybe, rather that ignoring the error, find a way to emit a warning?

bassosimone: Also: worth thinking whether we want to return a `bool` or a bool-like instance with more info on the actual error that occurred (having more info can help to respond to bug reports etc.)

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+    re_spec = r'^- `"([._/A-Za-z0-9]+)"`: (\((array|float|int|object|string)(;\ *(mandatory|optional))*\))?'
+    sections = {}
+    with open("include/measurement_kit/README.md", "rb") as filep:
+        section = None
+        for line in filep:
+            line = line.rstrip()
+            # Change section we're in
+            if (line == "## Tasks" or line == "## Settings" 
+                or line == "## Log levels" or line == "## Options"
+                or line == "## Events"):
```
bassosimone: Maybe the section names should not be hardcoded

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+                line = line.replace("## ", "").replace(" ", "_").lower()
+                section = line
+            # Extract documentation related to a section
+            if section and line.startswith('- `"'):
```
bassosimone: Maybe just check the regex?

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+                if not match:
+                    continue
+                name = match.group(1)
+                datatype = "string"
+                if match.group(3):
+                    datatype = match.group(3)
+                mandatory = False
+                if match.group(5):
+                    mandatory = match.group(5) == "mandatory"
+                spec = name, datatype, mandatory
```
bassosimone: Here I should explain what kind of data structure I'm creating and why, maybe

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+            elif isinstance(pattern, float):
+                filep.write("double %s = 0.0;\n" % name)
+            else:
+                raise RuntimeError("unexpected type for: %s" % pattern)
+        filep.write("};\n\n")
+
+def write_events_callbacks(filep, events):
+    """ Writes events callbacks in the header file """
+    for key, value in sorted(events.items()):
+        filep.write("/// Callback taking `const %s &` as unique argument\n" %
```
bassosimone: All these doxygen comments lead me to think that:

1. probably the main README should point to doxygen rather than replicating synopsis

2. more explicit documentation can be moved from engine.hpp to ffi.h

3. there should be automatic codedocs documentation also for sub APIs

4. we should not install README.md:s as part of `make install`

5. we probably need other README.md:s for other sub APIs

6. we should make sure the list of other APIs in the main README.md for include is up-to-date

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+    filep.write("/// Info on a generic task\n")
+    filep.write("class TaskInfo {\n")
+    filep.write("  public:\n")
+    for name in sorted(settings.keys()):
+        value, omit_nl = settings[name], False
+        filep.write("    /// '%s' field of settings\n" % name)
+        if isinstance(value, list):
+            filep.write("    std::vector<std::string> %s;\n" % name)
+            filep.write("\n")
+            filep.write("    /// Allows to append to '%s' using a fluent style\n" % name)
```
bassosimone: maybe use "fluent interface" rather than "fluent style" (need to check for the canonical name of the thing)

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+            singular_name = name
+            if singular_name.endswith("s"):
+                singular_name = singular_name[:-1]
+            filep.write("    TaskInfo &add_%s(const std::string &s) noexcept {\n" % singular_name)
+            filep.write("        %s.push_back(s);\n" % name)
+            filep.write("        return *this;\n")
+            filep.write("    }\n")
+        elif isinstance(value, dict):
```
bassosimone: It should probably make sense to justify the choice of using a JSON in case of dictionary and not using a JSON in case of list. What do we gain by doing that? Is that consistent or uselessly inconsistent? Would using _always_ use a JSON for all non-integral types be a less surprising way of organising the code?

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+            filep.write("    nlohmann::json %s;\n" % name)
+            filep.write("\n")
+            for var in ("double v", "int64_t v", "const std::string &v"):
```
bassosimone: Using `int64_t` is suboptimal here in that most often one uses `int`. Maybe the approach of nettests where we explicitly have two cases, integral and string, leads to easier to use code?

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+                    filep.write("    /// Allows to set '%s' option using a fluent style\n" % oname)
+                    if otype == "float":
+                        otype = "double v"
+                    elif otype == "int":
+                        otype = "int64_t v"
+                    elif otype == "string":
+                        otype = "const std::string &v"
+                    else:
+                        raise RuntimeError("unexpected option type: %s" % otype)
+                    filep.write("    TaskInfo &set_%s_option(%s) noexcept {\n" % (
```
bassosimone: Also here there is the `int64_t` problem

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+    filep.write("\n")
+
+def write_nlohmann_json_exception(filep):
+    """ Writes the NlohmannJsonException class """
+    filep.write("/// Exceptions emitted by JSON library\n")
+    filep.write("#if NLOHMANN_JSON_VERSION_MAJOR >= 3\n")
+    filep.write("using NlohmannJsonException = nlohmann::json::exception;\n")
+    filep.write("#else\n")
+    filep.write("using NlohmannJsonException = std::exception;\n")
+    filep.write("#endif\n\n")
```
bassosimone: Since `nlohmann::json::exception` derives from `std::exception`, this matter is actually much simpler than I initially anticipated, and it makes sense to update comments and see if we can update ASAP.

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+    filep.write("                event = nlohmann::json::parse(pstr);\n")
+    filep.write("            } catch (const NlohmannJsonException &) {\n")
+    filep.write("                assert(false);\n")
+    filep.write("                rv = false;\n")
+    filep.write("                break;\n")
+    filep.write("            }\n")
+    filep.write("            rv = process_event(info, event);\n")
+    filep.write("        }\n")
+    filep.write("        if (!rv) {\n")
+    filep.write("            mk_task_interrupt(ptask.get());\n")
```
bassosimone: This should probably be tested.

script/gen/cxx14/all
```diff
@@ -0,0 +1,419 @@
...
+    filep.write("\n")
+    filep.write("    /// Start a task in a background thread\n")
+    filep.write("    void start(TaskInfo &&info, std::function<void()> &&cb) const noexcept {\n")
+    filep.write("        std::thread thread{[this, info = std::move(info), cb = std::move(cb)]() mutable noexcept {\n")
+    filep.write("            run(std::move(info));\n")
+    filep.write("            cb();\n")
+    filep.write("        }};\n")
+    filep.write("    }\n")
+    filep.write("\n")
+    filep.write("    /// Factory to obtain the global runner\n")
```
bassosimone: What do we gain with this?

src/measurement_kit/cmd/multi_ndt.cpp
```diff
@@ -23,7 +23,7 @@ int main(std::list<Callback<BaseTest &>> &initializers, int argc, char **argv) {
         case 'u':
             // By default only the download phase is performed
             test.set_option("single_test_suite",
-                    MK_NDT_DOWNLOAD | MK_NDT_UPLOAD);
+                    (int64_t)(MK_NDT_DOWNLOAD | MK_NDT_UPLOAD));
```
bassosimone: Ideally, this should not be needed.

script/gen/cxx14
```diff
@@ -0,0 +1,490 @@
...
+
+def gather_events_info():
+    """ Gathers info on events """
+    events = {}
+    with open("include/measurement_kit/README.md", "r") as filep:
+        current = []
+        inside = False
+        for line in filep:
+            line = line.rstrip()
+            if line == "```JSON":
```
hellais: I would consider having a custom language descriptor (!= `JSON`) to indicate event_info JSON documents.

You may want in the future to enrich the Readme.md with other JSON formatted data and will be sad you assigned it to this (or at least add a note to the Readme.md saying that this is a strict assumption).

Alternatively maybe you can put the JSON documents for the event definitions in another place.

script/gen/cxx14
```diff
@@ -0,0 +1,490 @@
...
+from __future__ import print_function
+
+import json
+import re
+import sys
+
+if sys.version_info[0] == 3:
+    STR = str
+elif sys.version_info[0] == 2:
+    STR = unicode
```
hellais: I would suggest (since this is dev tooling) to assume just one version of python and not worry about trying to support both.

On which python version did you run this? I would pick that and exit if it doesn't match the desired version.

script/gen/cxx14
bassosimone: Yeah, this is a solid concern. I am not sure I want to add other kind of JSON formatted data to the README, since that is meant to document this kind of JSONs. I hear your concern loud and clear, tho, and I think that saying explicitly that the documentation is used to generate code is a more robust thing to do. (I am a bit uncomfortable with not using JSON in the use case, since when you run the code you seen JSONs, so showing you what you see is helpful.)

bassosimone: (I am also not comfortable with moving the JSONs to another place, since that is the place where one needs to read to understand the JSONs, basically. So having links to outside stuff will probably strike me as something that would make people nervous: too many indirections)

script/gen/cxx14
bassosimone: Mostly Python3. I am not sure I would call this a "dev-only" tooling. It is currently run everytime you run `./autogen.sh` to make sure, among other things, that one cannot forget about changes made in the docs and not see the side effects. As such, I wrote it with also Python2 in mind, since there are versions of Ubuntu shipping Python2 as default. (The approach of generating everything on `./autogen.sh` is not something I've married with, but I should not that it did serve me well over the years, because it's one less thing to worry about.)

hellais: I would suggest not regenerating code when you run autogen.sh if the code is checked into the source tree (which it is).

The generated code may change in subtle and unexpected ways and it can be troublesome to debug and ensure that the user is compiling the same code that is checked into the repo.

bassosimone: The problem I want to solve (TM) is: I do not want a developer to apply changes to the format and forgetting to regenerate the `cxx14.hpp` file (or other generated stuff). An alternative solution to solve this problem, rather than generating everything is `autogen.sh`, is to have a script that runs on travis before compiling that (1) regenerates everything and (2) fails if anything that is generated is changed. In principle, this does not necessarily avoid the need to be compatible with Python 2, but seems more consistent as a solution, because it guarantees that, when you compile MK as a user, you get the same code that is checked into the repo.

hellais:
We had a discussion with @bassosimone related to the new MK API.

This started from a question I raised related to how we are going to implement measurement upload in the new API and wether this should be using the `task` based API or it's going to be some new function.

I raised the concern that I get the impression that the API being `task` oriented is perhaps too high level, given the fact that probably we want to separate things more based on the specific purpose.

For example the lifecycle of calling a `task` to run a `nettest` is going to be very different from that of calling a `task` to upload a measurement or do a geoip lookup.

Based on this reflection it seemed like it would be better, to give more guarantees to a user of the ffi interface, to have separate calls based on the particular purpose (i.e. have a separation between a `task` an a `nettest` call).

This has the benefit of having greater assurance (and make it easer to understand) which particular events are relevant to particular API call.

We concluded that it's probably best to go in a direction where the structure of the new API is more modular and has even a greater per test based separation.

How we are going to do this is that we will:

1. Proceed with the 0.9 version of MK how we are doing it now (i.e. use the task based abstraction)
2. Work towards a 0.10 that takes into account what is said above
3. To the extent that it's possible the high level API (the ones for java and iOS) will follow this philosophy, although it will still be implemented in the 0.9 way of doing things, so that users of the high level API should not have to change things

bassosimone:
> We had a discussion with @bassosimone related to the new MK API.
[...]

Thanks! I'd like to add some more low-level details for future self reference. We concurred that it may be a wise thing to do to expose a FFI-like API in libraries like measurement-kit/libndt and measurement-kit/web-api-client, because that allows anyone to only pick the components they want and, more specifically, allows us to invoke directly specific functionality. So, in moving forward, we are probably going to scorporate more pieces out of this repository and use them directly in our integrations, rather than integrating pieces into this repository. (As a matter of fact, this seems also a more agile way of proceeding, where smaller independent pieces interact.)

bassosimone:
It would be too complex to really merge this PR at this point, since conflicting changes have been merged and included into the `master` branch. It's safe to just dispose of it.

PR #1645, opened 2018-07-11
autoapi: automatically generate C++11 API headers

bassosimone:
The problem that we want to solve here is the following. OONI mobile needs Java on Android and ObjectiveC (or Objective C++) on iOS. OONI desktop needs Go. The <strike>existing</strike> old API is not suitable for such wrapping, except for ObjectiveC++, because it is asynchronous and relies heavily on lambdas and closures. Making this right on a cross language fashion is too much complexity and work. I can even manage to do that, but then I will regret when there will be further changes.

To start solving this problem, since January we introduced a new FFI API. Since this API is basically C it can be easily used from basically wherever. Including also places like Python that use libffi. Yet, this FFI API has the drawback that it is unclear what JSON structures will be accepted in input and emitted at any given time. The original plan was to document stuff and to write code in the target language for generating the input JSONs (settings) and for processing the emitted JSONs (events). This has already been done for Go. The original plan for Java was to write Java code using Gson for doing this job. My initial approach to solve the problem was to extract the JSON messages from the documentation and generate code. This has been already partially merged and the bulk of the work for Java was meant to be made as part of #1640. I started experimenting with this idea in https://github.com/bassosimone/ffi-utils and then started applying changes back to #1640.

More or less at the same time, I realised that, for a simple API like the one of libndt, it was actually quite simple to generate very useable automatic bindings in  [Java](https://github.com/bassosimone/libndt-java) and [Golang](https://github.com/bassosimone/libndt-go).

This pull request is the result of this reasoning. Here we generate blocking code with a design similar to the one of libndt. This allows us, in principle, to easily use SWIG in the same way. This generated code is C++11 inline code, based on the FFI API. If anything, the simplicity of the FFI API was great because it allowed us to layer several different library-consumer oriented APIs on top of it. One of them is the new API introduced here. The other is a backward compatible one.

Some highlights:

1. There is a integration test checking whether generated code is current and fails the build otherwise.

2. Code generation uses `jinja2` (rater than regexes and documentation parsing as in #1640).

3. As mentioned, the structure of the generated C++ code is suitable for wrapping through SWIG.

4. The API is expressed by a simple DSL in Python.

I am looking for feedback on:

1. The design of the C++11 API (`./include/measurement_kit/nettest.hpp`)

2. Whether the code generation approach looks clean and maintainable (`./script/autoapi`)

3. Whether the safeguards implemented against people editing generated code are enough

4. Whether it makes sense as a trajectory (not now!) to only have this C++ API as the external API since it seems that at this point the FFI API adds lots of auto-generated complexity. (In this scenario the FFI API would have been a _tool_ for migrating away from an API that we could not easily embed and was unnecessarily asynchronous in too many places to an API that is much more easily doable with automatic tools &mdash; even though some more pleasant wrapper around the SWIG generated code is probably something to keep in mind as well.)

5. Whether this work adds or removes metres of rope around my neck :-).


include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,783 @@
...
+class WebConnectivitySettings : public CommonSettings {
+  public:
+    static constexpr const char *name = "WebConnectivity";
+    std::string serialize();
+};
+
+class WhatsappSettings : public CommonSettings {
+  public:
+    static constexpr const char *name = "Whatsapp";
+    bool all_endpoints = false;
```
bassosimone: I should probably add options for other tests are well.

.gitignore
```diff
@@ -44,6 +44,7 @@
 /test-driver
 /tmp
 /tor-bridges.tor-bridges-ip-port.csv
+/venv
```
bassosimone: Adding to `.gitignore` the directory used for the `virtualenv` consistently with `./script/autoapi` docs.

.travis.yml
```diff
@@ -2,6 +2,16 @@ language: cpp
...
+    # Verify that autoapi generated stuff is current
+    # ----------------------------------------------
+    #
+    # Just generate the files on travis and then make sure that no actual
+    # content has been changed into the work tree using the `-c` flag.
+    - env:
+      script:
+        - pip install --user jinja2
+        - ./script/autoapi/autoapi -c
+
```
bassosimone: Adding an integration test that ensures that `autoapi` specification is consistent with committed headers.

Doxyfile
```diff
@@ -1,6 +1,6 @@
 BUILTIN_STL_SUPPORT = YES
 EXTRACT_ALL = YES
-INPUT = include/measurement_kit/ffi.h
+INPUT = include/measurement_kit/ffi.h include/measurement_kit/nettest.hpp
```
bassosimone: Adding `nettest.hpp` to the public documentation.

autogen.sh
```diff
@@ -75,8 +75,6 @@ gen_executables() {
     done
 }
 
-./script/gen-ffi-macros
-
```
bassosimone: In this PR I removed the script previously used to generate FFI macros. These macros now are not used anymore and we're using instead a more understandable approach based on using `jinja2`.

cmake/AutoGeneratedTargets.cmake
```diff
@@ -1,11 +1,11 @@
...
+# Autogenerated by ./script/cmake/autogen on date Sun Jul 15 21:23:58 CEST 2018
 
 mk_declare_example("example/ffi" "ndt")
 mk_declare_example("example/ffi" "web_connectivity")
+mk_declare_example("example/nettest" "ndt")
+mk_declare_example("example/nettest" "web_connectivity")
 mk_declare_example("example/nettests" "multithread")
 mk_declare_example("example/nettests" "simple")
-mk_declare_example("example/swig" "ndt")
-mk_declare_example("example/swig" "web_connectivity")
```
bassosimone: This PR removes the `swig` API. The `nettest` API is designed to be easily integrable using SWIG for both Java and Golang (even though we're not planning on doing that _now_ for golang). These changes reflect the fact that the `swig` API examples are gone and the new `nettest` API examples have been added.

cmake/LibSources.txt
```diff
@@ -23,16 +23,15 @@ include/measurement_kit/common/settings.hpp
 include/measurement_kit/common/shared_ptr.hpp
 include/measurement_kit/common/version.h
 include/measurement_kit/ffi.h
-include/measurement_kit/ffi_macros.h
```
bassosimone: Removed because a better approach (generation using `jinja2` is now used).

cmake/LibSources.txt
```diff
@@ -23,16 +23,15 @@ include/measurement_kit/common/settings.hpp
...
 include/measurement_kit/ndt.hpp
 include/measurement_kit/ndt/defs.hpp
+include/measurement_kit/nettest.hpp
```
bassosimone: New public API header.

cmake/LibSources.txt
```diff
@@ -23,16 +23,15 @@ include/measurement_kit/common/settings.hpp
...
 include/measurement_kit/nettests.hpp
 include/measurement_kit/nettests/ffi.hpp
 include/measurement_kit/ooni.hpp
 include/measurement_kit/ooni/error.hpp
 include/measurement_kit/ooni/orchestrate.hpp
 include/measurement_kit/ooni/utils.hpp
-include/measurement_kit/swig.hpp
```
bassosimone: Removed public API header.

example/nettest/ndt.cpp
```diff
@@ -0,0 +1,23 @@
+// Public domain 2017, Simone Basso <bassosimone@gmail.com.
```
bassosimone: New example showing how to run NDT with the nettest API.

example/nettest/web_connectivity.cpp
```diff
@@ -0,0 +1,27 @@
+// Public domain 2017, Simone Basso <bassosimone@gmail.com.
```
bassosimone: Example showing how to run web connectivity.

example/swig/ndt.cpp
```diff
@@ -1,27 +0,0 @@
-// Public domain 2017, Simone Basso <bassosimone@gmail.com.
```
bassosimone: Removed example because API is gone.

example/swig/web_connectivity.cpp
```diff
@@ -1,31 +0,0 @@
-// Public domain 2017, Simone Basso <bassosimone@gmail.com.
```
bassosimone: Removed example because API is gone.

include/measurement_kit/ffi.h
```diff
@@ -4,15 +4,17 @@
 #ifndef MEASUREMENT_KIT_FFI_H
 #define MEASUREMENT_KIT_FFI_H
```
bassosimone: Just made the documentation more beautiful for Doxygen.

include/measurement_kit/ffi_macros.h
```diff
@@ -1,100 +0,0 @@
-/* Auto-generated by ./script/gen-ffi-macros; don't edit. */
```
bassosimone: Header gone. We now use `jinja2` rather than `X-macros`.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+/*-
+ *  __      __  _____ __________  _______  .___ _______    ________
+ * /  \    /  \/  _  \\______   \ \      \ |   |\      \  /  _____/
+ * \   \/\/   /  /_\  \|       _/ /   |   \|   |/   |   \/   \  ___
+ *  \        /    |    \    |   \/    |    \   /    |    \    \_\  \
+ *   \__/\  /\____|__  /____|_  /\____|__  /___\____|__  /\______  /
+ *        \/         \/       \/         \/            \/        \/
+ *
+ * Autogenerated by `./script/autoapi/autoapi`. DO NOT EDIT!!!
+ */
```
bassosimone: Added big warning at top. Coupled with new integration test, that's probably enough to make it difficult for people to make changes to the header rather than to the template and not notice it.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+
+/// \file include/measurement_kit/nettest.hpp
+///
+/// \brief C++11 API for running nettests.
+///
+/// This API is a C++11 adaptation of MK's FFI API. You should probably read
+/// about such API first. \see https://github.com/measurement-kit/measurement-kit/blob/master/include/measurement_kit/README.md.
```
bassosimone: I am not super happy about referring to the FFI API. In theory the auto-generated documentation in the Doxgen comments of this file should be more meaningful than the one that we have below. 😒

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+///
+/// Usage is as follows:
+///
+/// ### 1. Instantiate settings
```
bassosimone: The following describes the steps required to use this API. I think this is nice to have, even though there is a risk that this may become outdated information if changes are not performed in sync. The lazier alternative is to refer directly to the examples folder. However, this documentation is also helpful to understand some of the decisions that I made (even though not probably all), so I'd rather keep it and risk it becomes outdated.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+#include <measurement_kit/ffi.h>
+
+/// Measurement Kit namespace.
+namespace mk {
+
+/// Namespace containing the nettest API.
+namespace nettest {
+
+/// Contains events definitions.
+namespace events {
```
bassosimone: This namespace contains mechanically generated classes using `jinja2`. The documentation of the classes is not very informative, to be honest, as I mentioned above. One possible alternative would be to also include the documentation strings in the information passed to the `jinja2` template; too much complexity?

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+    static constexpr const char *log_level_warning = "WARNING";
+
+    /// String representation of the "info" log level.
+    static constexpr const char *log_level_info = "INFO";
+
+    /// String representation of the "debug" log level.
+    static constexpr const char *log_level_debug = "DEBUG";
+
+    /// String representation of the "debug2" log level.
+    static constexpr const char *log_level_debug2 = "DEBUG2";
```
bassosimone: The FFI API uses strings for verbosity. Since one typically needs to use such strings when setting the `Settings` I have decided to put them as static attributes of the `CommonSettings` class itself.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+    /* No attributes */
+};
+
+} // namespace events
+
+/// Contains settings classes.
+namespace settings {
+
+/// Settings common to all nettests.
+class CommonSettings {
```
bassosimone: In the FFI API there are two layers of options. The `settings` and inside of them the `options`. This actually happens for historical reasons. That is, because of how stuff is implemented internally. Logically, the distinction between settings and options does not make much sense. In general both contain stuff that you may not even want to configure unless you have specific needs. For this reason here all the common options and settings are part of the same class rather than nesting options inside settings.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+    std::vector<std::string> inputs = {};
+
+    /// The "input_filepaths" setting.
+    std::vector<std::string> input_filepaths = {};
+
+    /// The "log_filepath" setting.
+    std::string log_filepath = {};
+
+    /// The "log_level" setting.
+    std::string log_level = log_level_err;
```
bassosimone: The default value is the one actually used by MK. As I mentioned, I am not happy about the docstrings.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+
+    /// The "software_name" setting.
+    std::string software_name = {};
+
+    /// The "software_version" setting.
+    std::string software_version = {};
+};
+
+/// Settings of the "CaptivePortal" nettest.
+class CaptivePortalSettings : public CommonSettings {
```
bassosimone: In the FFI API, one needs to "know" what are nettest specific settings. Here there is a class for each nettest that contains nettest specific settings where available. This should help people understand what are the available settings for each given test. Inheritance is used such that one can pass a nettest-specific settings class to some common code that initialises all the common settings.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+};
+
+/// Settings of the "Whatsapp" nettest.
+class WhatsappSettings : public CommonSettings {
+  public:
+    /// Unique name of this nettest.
+    static constexpr const char *name = "Whatsapp";
+    
+    /// The "all_endpoints" setting of this nettest.
+    bool all_endpoints = false;
```
bassosimone: (There are actually more nettest specific settings than this. For this PR I just added the Whatsapp nettest settings such that it was clear the potential of having nettest specific settings.)

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+};
+
+} // namespace settings
+
+// Contains the routers
+namespace routers {
+
+/// Routes nettest events to virtual methods.
+class Router {
```
bassosimone: This class is really designed to be SWIG friendly. I have separated events routing and managing the nettest lifecycle because it occurred to me that event routing can have a longer lifecycle than just a single nettest.

One potential issue is that the previous API used lambdas rather than virtual methods. However, virtual methods are much more easily wrappable using SWIG. So, virtual methods. However, with a little more templating it is possible to define a derived class of router than takes one or more lambda per events. I am not planning on adding this class now, but I will do that next, because we use lambdas on iOS.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+  public:
+    // Implementation note: virtual methods cannot be `noexcept` because the
+    // task stable version of SWIG does not handle that correctly.
```
bassosimone: (This seems to be fixed in master but does not work for me locally. I can probably investigate more.)

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+    virtual void on_status_update_websites(const events::StatusUpdateWebsitesEvent &event);
+
+    /// Handles the "task_terminated" event.
+    virtual void on_task_terminated(const events::TaskTerminatedEvent &event);
+
+    virtual ~Router() noexcept;
+};
+
+/// Router that logs everything on std::clog.
+class NoisyRouter : public Router {
```
bassosimone: Just a convenience class that one can use to see stuff happening.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+};
+
+} // namespace routers
+
+/// Manages the lifecycle of a nettest. This API mirrors as closely as
+/// possible to FFI API provided by <measurement_kit/ffi.h>.
+class Nettest {
+  public:
+    // C++ object model
+    // ````````````````
```
bassosimone: The object model is such that one can use stack allocation or `new`. I have decided not to do initialisation in the constructor because I don't want to throw (throwing complicates stuff when using SWIG and, say, Golang). As such, there is a specific method one needs to call to "start" the test.

I have made this class explicitly non-copyable and non-movable for safety reasons. If one needs to have unique or shared ownership, the design allows to use `std::make_unique` or `std::make_shared` quite easily.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+    /// Starts a "CaptivePortal" nettest.
+    /// \param settings contains the settings configuring this nettest.
+    /// \param router is the router used to route nettest events. If the
+    /// router is `nullptr`, we'll use an internal dummy router.
+    /// \return true on success, false on failure. There are two reasons why
+    /// this method could fail. Either the provided settings are not JSON
+    /// serializable because there was a string that was not valid UTF-8, or
+    /// you already used this class to start a nettest. In both cases, log
+    /// messages will be routed, so you will know what happened.
+    bool start_captive_portal(settings::CaptivePortalSettings settings, routers::Router *router) noexcept;
```
bassosimone: Why a different function for each nettest? Mainly because I need to initialise the internal JSON setting with specific variables (like the test name and other nettest specific settings) before running. An alternative design would have been to allow for serialisation of nettests to string and then to have a single start method receiving a string. Yet, I do not want to expose too much that we're using JSON internally. So, I decided to put all the logic in this class and to have different start methods. It could also be that we have a specific class for each nettest. I think this design is quite equivalent to the one here, but maybe I am missing some bits. Feedback appreciated.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+    /// Starts a "Dash" nettest.
+    /// \param settings contains the settings configuring this nettest.
+    /// \param router is the router used to route nettest events. If the
+    /// router is `nullptr`, we'll use an internal dummy router.
+    /// \return true on success, false on failure. There are two reasons why
+    /// this method could fail. Either the provided settings are not JSON
+    /// serializable because there was a string that was not valid UTF-8, or
+    /// you already used this class to start a nettest. In both cases, log
+    /// messages will be routed, so you will know what happened.
+    bool start_dash(settings::DashSettings settings, routers::Router *router) noexcept;
```
bassosimone: Also, `router` is an argument so that `log` errors caused by incorrect usage (e.g. starting more than one test with this class) are not only apparent by the `bool` return value but also by explicit log messages.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+    /// router is `nullptr`, we'll use an internal dummy router.
+    /// \return true on success, false on failure. There are two reasons why
+    /// this method could fail. Either the provided settings are not JSON
+    /// serializable because there was a string that was not valid UTF-8, or
+    /// you already used this class to start a nettest. In both cases, log
+    /// messages will be routed, so you will know what happened.
+    bool start_whatsapp(settings::WhatsappSettings settings, routers::Router *router) noexcept;
+
+    // Event processing methods
+    // ````````````````````````
```
bassosimone: This mimics very closely the FFI API. I figured that deviating for that would have added more complexity.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+
+    /// Returns true if nettest is done, false otherwise.
+    bool is_done() noexcept;
+
+    /// Interrupts the currently running nettest.
+    void interrupt() noexcept;
+
+  private:
+    // Private methods
+    // ```````````````
```
bassosimone: The public implementation should be such that we could possibly even migrate away from all the JSON dance below and there is not implementation coupling (I think) with such details. This is good because for most of our use cases I noticed that we can actually use SWIG on top of this API rather than using JSON. The advantage is that using this API there is much more certainty with respect to types.

(I am not suggesting that we do that, I am just suggesting that exposing more JSON details'd be bad.)

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+ * units to include the implementation into a single object.
+ */
+#if !defined MK_NETTEST_NO_INLINE_IMPL && !defined SWIG
+
+// Routers
+// -------
+
+namespace routers {
+
+void Router::on_failure_asn_lookup(const events::FailureAsnLookupEvent &) {}
```
bassosimone: Both the `Router` and `NoisyRouter` calls are auto-generated.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+} // namespace routers
+
+// Nettest
+// -------
+
+Nettest::Nettest() noexcept {}
+
+Nettest::~Nettest() noexcept {}
+
+bool Nettest::start_captive_portal(settings::CaptivePortalSettings settings, routers::Router *router) noexcept {
```
bassosimone: These `start_XXX` methods are auto-generated. (I need to map in more nettest specific settings. Until I do that, they will not appear in this implementation.)

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+ *  |____|     |__|  |__| \_/  (____  /__|  \___  >
+ *                                  \/          \/
+ *
+ * Definitions ends here. What follows is the inline implementation of this
+ * API, which you can exclude by setting MK_NETTEST_NO_INLINE_IMPL.
+ *
+ * This is handy if you want to include this code in many translation
+ * units to include the implementation into a single object.
+ */
+#if !defined MK_NETTEST_NO_INLINE_IMPL && !defined SWIG
```
bassosimone: I like this pattern of separating interface and implementation keeping all in the same header file - for relatively small files - since it seems to me this simplifies both integration and testing. I'll probably apply this pattern more in the future, to libndt and other small libs we want to integrate.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+void Nettest::route_next_event(routers::Router *router) noexcept {
+    router = (router) ? router : default_router();
+    UniqueEvent eventptr;
+    nlohmann::json ev;
+    {
+        std::unique_lock<std::mutex> _{mutex_};
+        eventptr.reset(mk_task_wait_for_next_event(task_.get()));
+    }
+    if (eventptr == nullptr) {
+        // TODO(bassosimone): route this error.
```
bassosimone: I will add code to generate `log` messages here before hitting "merge".

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+        interrupt();
+        return;
+    }
+    if (ev.count("key") <= 0 || !ev.at("key").is_string() || ev.count("value") <= 0 || !ev.at("value").is_object()) {
+        // TODO(bassosimone): route this error.
+        interrupt();
+        return;
+    }
+    if (ev.at("key") == events::FailureAsnLookupEvent::key) {
```
bassosimone: The code from here until the end of the method is auto-generated.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+    router = (router) ? router : default_router();
+    UniqueEvent eventptr;
+    nlohmann::json ev;
+    {
+        std::unique_lock<std::mutex> _{mutex_};
+        eventptr.reset(mk_task_wait_for_next_event(task_.get()));
+    }
+    if (eventptr == nullptr) {
+        // TODO(bassosimone): route this error.
+        interrupt();
```
bassosimone: The design decision here is that anytime we see something unexpected we log and interrupt the test.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,1922 @@
...
+            interrupt();
+            return;
+        }
+        event.idx = ev.at("value").at("idx");
+        if (ev.at("value").count("input") <= 0 || !ev.at("value").at("input").is_string()) {
+            // TODO(bassosimone): route this error.
+            interrupt();
+            return;
+        }
+        event.input = ev.at("value").a
```
bassosimone: This part is of course not autogenerated. It interrupts the test if we see an unmapped event.

bassosimone: This block of code to fill in common settings is autogenerated.

bassosimone: Code from here until the method end is not autogenerated.

include/measurement_kit/nettests/ffi.hpp
```diff
@@ -426,14 +436,97 @@ class MK_NETTESTS_DEPRECATED BaseTest {
...
 };
 
-#define MK_DECLARE_TEST(_name_,  _type_ignored_, _mandatory_ignored_)          \
-    class _name_##Test : public BaseTest {                                     \
-      public:                                                                  \
-        _name_##Test() : BaseTest() { impl_->settings["name"] = #_name_; }     \
-    };
-MK_ENUM_TASKS(MK_DECLARE_TEST)
-#undef MK_DECLARE_TEST // Tidy
+class CaptivePortalTest : public BaseTest {
```
bassosimone: Here I replaced `X-macros` code with explicit code generated using `jinja2`.

include/measurement_kit/swig.hpp
```diff
@@ -1,177 +0,0 @@
-// Part of Measurement Kit <https://measurement-kit.github.io/>.
```
bassosimone: This API has been removed. The original idea was to wrap this using SWIG. But then I realised that, as long as we use sync APIs (like `nettest.hpp`) and manage threading in the target language (e.g. Java), we can quite easily wrap a much more rich API. So the point of this "swig" API was vanished.

script/autoapi/autoapi
```diff
@@ -0,0 +1,189 @@
...
+
+class Attribute(object):
+    """ Attribute of a class. """
+
+    def __init__(self, cxx_type, key, value=None):
+        self.cxx_type = cxx_type
+        self.key = key
+        self.value = value if value is not None else "{}"
+        # Casting from/to JSON is required for booleans because the FFI
+        # implementation uses integers instead of booleans.
```
bassosimone: Note: this cannot be changed without heavy internals refactoring. So, it can be done, but now we're working on the external API, so it's quite not the time to start refactoring the internals.

script/autoapi/engine.cpp.j2
```diff
@@ -0,0 +1,582 @@
+// Part of Measurement Kit <https://measurement-kit.github.io/>.
+// Measurement Kit is free software under the BSD license. See AUTHORS
+// and LICENSE for more information on the copying conditions.
```
bassosimone: This is the template for `engine.cpp`. Is it readable? (I'll comment on specific choices in `engine.cpp`.)

script/autoapi/nettest.hpp.j2
```diff
@@ -0,0 +1,463 @@
+// Part of Measurement Kit <https://measurement-kit.github.io/>.
+// Measurement Kit is free software under the BSD license. See AUTHORS
+// and LICENSE for more information on the copying conditions.
+#ifndef MEASUREMENT_KIT_NETTEST_HPP
+#define MEASUREMENT_KIT_NETTEST_HPP
```
bassosimone: Is this file readable? (I have already commented on specific choices in `nettest.hpp`.)

script/autoapi/nettests-ffi.hpp.j2
```diff
@@ -0,0 +1,448 @@
+// Part of Measurement Kit <https://measurement-kit.github.io/>.
```
bassosimone: Same as above: is it readable? (My comments are in the `nettests/ffi.hpp` header diff.)

script/gen-ffi-macros
```diff
@@ -1,72 +0,0 @@
-#!/usr/bin/env python
```
bassosimone: Script gone as now we use `jinja2`.

src/libmeasurement_kit/common/libevent_reactor.hpp
```diff
@@ -101,12 +101,18 @@ class LibeventReactor : public Reactor, public NonCopyable, public NonMovable {
```
bassosimone: This is actually a bugfix for a bug noticed while working on this PR. See 0477c8c79bf1e8507a84fd35bfd42ef29369e8bd.

(Can become a separate PR, if needed.)

src/libmeasurement_kit/common/logger.cpp
```diff
@@ -210,41 +209,25 @@ class DefaultLogger : public Logger, public NonCopyable, public NonMovable {
 
     void emit_event_ex(
             const std::string &key, nlohmann::json &&value) override {
-        if (!value.is_object()) {
```
bassosimone: Removed event validation from here. Instead now I register a handler for unhandled events in `engine.cpp` and I make sure in there that events are correct. Removes some code duplication.

src/libmeasurement_kit/common/logger.cpp
```diff
@@ -210,41 +209,25 @@ class DefaultLogger : public Logger, public NonCopyable, public NonMovable {
...
         nlohmann::json event{
             {"key", key},
             {"value", std::move(value)}
         };
+        std::string real_key;
+        if (handlers_.count(key) <= 0) {
+            // Even if the event has not registered handler, we pass them up
+            // one layer to validate the event structure.
+            real_key = "__disabled";
+            assert(handlers_.count(key));
```
bassosimone: This `assert()` is here to make sure I have actually registered a handler for unhandled messages.

src/libmeasurement_kit/engine.cpp
```diff
@@ -153,22 +162,40 @@ Task::~Task() {
 // # Helpers
 
 static std::tuple<int, bool> log_level_atoi(const std::string &str) {
-#define ATOI(value, type_ignored, mandatory_ignored)                           \
-    if (str == #value) {                                                       \
-        return std::make_tuple(MK_LOG_##value, true);                          \
+    if (str == "ERR") {
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -153,22 +162,40 @@ Task::~Task() {
...
-    MK_ENUM_LOG_LEVELS(ATOI)
-#undef ATOI
     return std::make_tuple(0, false);
 }
 
 static std::tuple<std::string, bool> log_level_itoa(int n) {
-#define ITOA(value, type_ignored, mandatory_ignored)                           \
-    if (n == MK_LOG_##value) {                                                 \
-        return std::make_tuple(std::string{#value}, true);                     \
+    if (n == MK_LOG_ERR) {
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -190,63 +217,302 @@ static nlohmann::json make_failure_event(const Error &error) {
     return object;
 }
 
-static bool is_event_valid(const std::string &str) {
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -190,63 +217,302 @@ static nlohmann::json make_failure_event(const Error &error) {
...
 static nlohmann::json possibly_validate_event(nlohmann::json &&event) {
     // In debug mode, make sure that we're emitting an event that we know
     assert(event.is_object());
     assert(event.count("key") != 0);
     assert(event.at("key").is_string());
-    assert(is_event_valid(event.at("key").get<std::string>()));
+    assert(is_event_key_valid(event.at("key").get<std::string>()));
     assert(event.count("value") != 0);
     assert(event.at("value").is_object());
+    do {
```
bassosimone: Added autogenerated code that makes sure events have the desired structure.

src/libmeasurement_kit/engine.cpp
```diff
@@ -190,63 +217,302 @@ static nlohmann::json make_failure_event(const Error &error) {
...
+            assert(event.at("value").at("status").is_string());
+            break;
+        }
+    } while (0);
     return event;
 }
 
 static nlohmann::json known_events() {
     nlohmann::json json;
-#define ADD(name, type_ignored, mandatory_ignored) json.push_back(#name);
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -190,63 +217,302 @@ static nlohmann::json make_failure_event(const Error &error) {
...
+    json.push_back("status.started");
+    json.push_back("status.update.performance");
+    json.push_back("status.update.websites");
+    json.push_back("task_terminated");
     return json;
 }
 
 static std::string known_tasks() {
     nlohmann::json json;
-#define ADD(name, type_ignored, mandatory_ignored) json.push_back(#name);
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -190,63 +217,302 @@ static nlohmann::json make_failure_event(const Error &error) {
...
+    json.push_back("TcpConnect");
+    json.push_back("Telegram");
+    json.push_back("WebConnectivity");
+    json.push_back("Whatsapp");
     return json.dump();
 }
 
 static std::string known_log_level_levels() {
     nlohmann::json json;
-#define ADD(name, type_ignored, mandatory_ignored) json.push_back(#name);
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -190,63 +217,302 @@ static nlohmann::json make_failure_event(const Error &error) {
...
+    json.push_back("warning");
+    json.push_back("info");
+    json.push_back("debug");
+    json.push_back("debug2");
     return json.dump();
 }
 
 static std::unique_ptr<nettests::Runnable> make_runnable(const std::string &s) {
     std::unique_ptr<nettests::Runnable> runnable;
-#define ATOP(value, type_ignored, mandatory_ignored)                           \
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -263,28 +529,90 @@ static bool validate_known_settings_shallow(
         TaskImpl *pimpl, const nlohmann::json &settings) {
     auto rv = true;
 
-#define VALIDATE(name, type, mandatory)                                        \
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -294,15 +622,21 @@ static void remove_unknown_settings_and_warn(
         TaskImpl *pimpl, nlohmann::json &settings) {
     std::set<std::string> expected;
     std::set<std::string> unexpected;
-#define FILL(name, type, mandatory) expected.insert(#name);
```
bassosimone: Replaced `X-macros` with autogenerated code.

src/libmeasurement_kit/engine.cpp
```diff
@@ -349,12 +683,364 @@ static void task_run(TaskImpl *pimpl, nlohmann::json &settings) {
     // extract and process `options`
     if (settings.count("options") != 0) {
         auto &options = settings.at("options");
-        // TODO(bassosimone): enumerate all possible options. Make sure we
```
bassosimone: Replaced this TODO with some autogenerated code (in the following) that starts doing checks.

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -279,7 +279,8 @@ void Runnable::geoip_lookup(Callback<> cb) {
             logger->emit_event_ex("status.geoip_lookup", {
                 {"probe_asn", probe_asn},
                 {"probe_cc", probe_cc},
-                {"probe_ip", ip}
+                {"probe_ip", ip},
+                {"probe_network_name", ""} // TODO(bassosimone): gather!
```
bassosimone: Another bug fix. This key was missing. I need to improve the code to actually fill it (in another PR).

test/swig/initialize.cpp
```diff
@@ -1,43 +0,0 @@
-// Part of Measurement Kit <https://measurement-kit.github.io/>.
```
bassosimone: API gone. Unit test no longer relevant.

include/measurement_kit/common/logger.hpp
```diff
@@ -185,8 +185,7 @@ class Logger {
     virtual void set_logfile(std::string fpath) = 0;
 
     /// `emit_event_ex()` emits an event as a JSON.
-    virtual void emit_event_ex(const std::string &key,
-                               nlohmann::json &&value) = 0;
+    virtual void emit_event_ex(std::string key, nlohmann::json &&value) = 0;
```
bassosimone: Changed the API because we modify the key in this method.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,2589 @@
...
+constexpr const char *log_level_info = "INFO";
+
+/// String representation of the "debug" log level.
+constexpr const char *log_level_debug = "DEBUG";
+
+/// String representation of the "debug2" log level.
+constexpr const char *log_level_debug2 = "DEBUG2";
+
+/// Groups all events together.
+namespace events {
```
bassosimone: All in this namespace is autogenerated.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,2589 @@
...
+class TaskTerminatedEvent {
+  public:
+    /// The unique identifier of this event.
+    static constexpr const char *key = "task_terminated";
+    /* No attributes */
+};
+
+} // namespace events
+
+#if !defined SWIG && !defined DOXYGEN
```
bassosimone: Internal undocumented namespace containing stuff that must be here for the remainder of the code to work but that generally one should not look into. Maybe put this instead inside `ffi.h`?

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,2589 @@
...
+
+} // namespace detail
+#endif // !SWIG && !DOXYGEN
+
+/// Common generic code.
+namespace common {
+
+/// Settings common to all nettests.
+class Settings {
+  public:
```
bassosimone: All attributes are autogenerated. In the future we may want to have more explanatory docstrings.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,2589 @@
...
+
+    /// The "software_name" setting.
+    std::string software_name = "";
+
+    /// The "software_version" setting.
+    std::string software_version = "";
+
+#ifdef SWIG
+  private:
+#endif
```
bassosimone: SWIG must not generate wrappers for this.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,2589 @@
...
+    // `````````
+
+    /// Runs the nettest.
+    virtual bool run() noexcept { return false; }
+
+    /// Interrupts a running nettest.
+    void interrupt() noexcept;
+
+    // Event handlers
+    // ``````````````
```
bassosimone: Handlers are autogenerated

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,2589 @@
...
+    virtual void on_status_update_websites(events::StatusUpdateWebsitesEvent);
+
+  protected:
+    // Dispatch the JSON event to the proper handler
+    bool dispatch_event(nlohmann::json doc) noexcept override;
+};
+
+} // namespace common
+
+#ifndef MK_NETTEST_NO_CAPTIVE_PORTAL
```
bassosimone: I have been asked about omitting nettests in the past. This may be a first step in this direction.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,2589 @@
...
+
+/// Settings for CaptivePortal
+class CaptivePortalSettings : public common::Settings {
+  public:
+    /// Unique name of this nettest.
+    static constexpr const char *name = "CaptivePortal";
+    /* No nettest-specific settings */
```
bassosimone: I have not mapped settings for most nettests yet. In most cases there are nettest specific settings.

include/measurement_kit/nettest.hpp
```diff
@@ -0,0 +1,2589 @@
...
+    (*doc)["output_filepath"] = output_filepath;
+    {
+        auto &o = (*doc)["options"];
+        o["bouncer_base_url"] = bouncer_base_url;
+        o["collector_base_url"] = collector_base_url;
+        o["dns/nameserver"] = dns_nameserver;
+        o["dns/engine"] = dns_engine;
+        o["geoip_asn_path"] = geoip_asn_path;
+        o["geoip_country_path"] = geoip_country_path;
+        o["ignore_bouncer_error"] = (int64_t)ignore_bouncer_error;
```
bassosimone: We internally use integers. We cannot change this _easily_ now.

src/libmeasurement_kit/common/libevent_reactor.hpp
bassosimone: Is now a separate PR #1646

pboothe:
This is a lot of changes, but they look like they are moving in the right direction. LGTM.

bassosimone:
Thanks! (I do not understand what `coveralls` is waiting after several days and I am going to wait just a little more, after which I will merge the PR.)

PR #1646, opened 2018-07-16
fix: libevent_reactor.hpp: make reactor stoppable

bassosimone:
Found while thinkering with #1645. This _may_ have implications, especially on the way in which the thread pool is implemented. So it's important to do some thinking before moving forward.


bassosimone:
This is too difficult because it breaks many hidden assumptions in the code. My sense is that, by progressively deploying smaller building blocks, we will make the libevent reactor unnecessary (or we can confine it in smaller scopes). At that point, improving the worker pool will be easier.

PR #1647, opened 2018-07-29
build/cross/ios: disable thread local for i386

bassosimone:
Signed-off-by: Simone Basso <bassosimone@gmail.com>


bassosimone:
It is expected to see an AppVeyor failure because Windows support is not part of stable. Regarding coveralls, the coverage drop is unrelated to this commit. It is more of a consequence of a previous branch having higher coverage. So, we can merge when `travis-ci/pr` is happy.

PR #1649, opened 2018-08-01
Integration branch for v0.9.0-alpha.x with x >= 3

bassosimone:
Bless a new alpha and, now that #1645 is landed, resume working on the Android API.


bassosimone:
This branch is running for quite some time now. It could actually make sense to merge it now into master because it contains useful fixes that prevent other PRs from being merged cleanly (see e.g. #1658) and spending time back-porting stuff would be a waste of time at this point.

PR #1665, opened 2018-10-10
Release v0.8.4

bassosimone:
This branch contains some fixes that we need to land on stable for an upcoming point release of the OONI mobile applications, as planned with @lorenzoPrimi in the Mexico meeting.


bassosimone:
@lorenzoPrimi please, try using `v0.8.4` for iOS and `v0.8.4-3` for Android, and tell me whether they work for you and the functionality in there is consistent with what you needed :-).

PR #1667, opened 2018-10-27
Release v0.9.0-alpha.11

bassosimone:


bassosimone:
Travis seems to be stuck. Appveyor is building. A previous version of this diff successfully built yesterday. What changed in here is just a new version of a vendored dependency that builds on its own CI, therefore I am going to merge anyway.

PR #1674, opened 2018-11-06
Add annotations when geoip-lookup steps fail

### LINKED ISSUE #1435: Distinguish between privacy settings and failure to lookup geoip information
Currently we don't record in the measurement if the ASN is `AS0` because the user has opted out to submitting network information vs the geoip lookup failed.

I noticed this because I found that a old maxmind DB leads to the ASN not being discovered and as a result that is set as `AS0` in the measurement.

It would be useful to maybe write an annotation with the privacy settings so that we can actually distinguish these cases and if we start noticing many of these cases, then we can push an updated geoip DB.

bassosimone:
Closes #1435 

```JSON
{
    "annotations": {
        "engine_name": "libmeasurement_kit",
        "engine_version": "0.9.0-beta",
        "engine_version_full": "v0.9.0-beta",
        "failure_asn_lookup": "true",
        "failure_network_name_lookup": "true",
        "platform": "macos"
    }
}
```

### DIFF
--- a/src/libmeasurement_kit/nettests/runnable.cpp
+++ b/src/libmeasurement_kit/nettests/runnable.cpp
@@ -209,6 +209,7 @@ void Runnable::geoip_lookup(Callback<> cb) {
             logger->warn("=== BEGIN IP_LOOKUP LOGS ===");
             logger->warn("%s", logs.c_str());
             logger->warn("=== END IP_LOOKUP LOGS ===");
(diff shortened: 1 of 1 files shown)


bassosimone:
Will merge. The [AppVeyor failure](https://ci.appveyor.com/project/bassosimone/measurement-kit/builds/20079902) is a connect timeout of NDT. Nothing to worry about.

PR #1722, opened 2019-01-04
Add settings

elevenfive:
- Logs contained warnings such as:
  `"Found mlabns\/address_family option which is not mapped as a valid option. We will continue processing"`


bassosimone:
I noticed that I've removed the possibility of overriding `probe_asn` and `probe_cc` (i.e. you can set them, but nothing happens. I'm adding a fix also for this.)

bassosimone:
Thanks! I cherry-picked your commit into a larger working branch of mine where I also updated the documentation! 👍

PR #1725, opened 2019-01-14
Release/0.9.2

bassosimone:
This PR backports parts of #1722 that should facilitate @elevenfive's usage of MK and wraps these changes into a new stable release of MK, v0.9.2.


bassosimone:
So, the CI is failing on Travis but _not failing_ for me locally and on AppVeyor. 🤔

bassosimone:
So, [we now have insights](https://travis-ci.org/measurement-kit/measurement-kit/jobs/479642815#L1356). After I updated the container used for running tests, which in turn updated OpenSSL (I believe, but this is a reasonable guess), we see:

```
[D] ssl: handshake...
[D] ssl: handshake error: ssl_error: error:1414D172:SSL routines:tls12_check_peer_sigalg:wrong signature type
```

Conclusion. The test fails because the OpenSSL we're using in the container is such that the hushmail.com test fails. Didn't see it otherwise because only on Travis we're using that specific container. It may be worth checking whether updating the container again fixes.

bassosimone:
FWIW, I can see the same error in the same container with `mkcurl` and `curl`:

```
$ curl -v https://www.hushmail.com
*   Trying 72.51.60.140...
* TCP_NODELAY set
* Connected to www.hushmail.com (72.51.60.140) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (OUT), TLS alert, handshake failure (552):
* error:1414D172:SSL routines:tls12_check_peer_sigalg:wrong signature type
* Closing connection 0
curl: (35) error:1414D172:SSL routines:tls12_check_peer_sigalg:wrong signature type
```

bassosimone:
Before moving on with this PR, I need to address the issue of nonworking integration tests. Then, I possible, I'd like also to have #1727 fixed in here.

bassosimone:
We're all green now!

bassosimone:
When travis and appveyor complete, we can finally merge and bless v0.9.2

PR #1745, opened 2019-02-03
Continue refactoring task

bassosimone:
1. Use the default queue for running tests, so we enforce that orchestration and tests cannot actually run concurrently (as it ought to be)

2. Pass settings by move, rather than that by reference, to task_run_legacy


bassosimone:
I have added a test [that breaks because of the changes of this PR](https://travis-ci.org/measurement-kit/measurement-kit/builds/488207420). Now that Travis also shows that with at least one failed build, I can proceed with committing a diff that (hopefully) will fix the breakage.

bassosimone:
I am going to close this. It seems I rediscovered detached threads are not so easy to handle. I could probably fix this, but I should also avoid doing tricky stuff.

PR #1748, opened 2019-02-14
Start preparing for MK v0.10.0-beta

bassosimone:


bassosimone:
The [AppVeyor test that failed](https://ci.appveyor.com/project/bassosimone/measurement-kit/builds/22379638#L1546) looks like:

```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
runnable.exe is a Catch v2.6.1 host application.
Run with -? for options
test/nettests/runnable.cpp:238
test/nettests/runnable.cpp:239: FAILED:
with expansion:
  2 == 9
test cases:   7 |   6 passed | 1 failed
assertions: 696 | 695 passed | 1 failed
FAIL test/nettests/runnable.exe (exit status: 1)
```

This is an innocuous false positive that happens from time to time.

src/libmeasurement_kit/net/libssl.hpp
```diff
@@ -288,42 +288,36 @@ template <size_t max_cache_size = 64> class Cache {
...
-        logger->warn("ssl: got an invalid certificate");
-        return SslInvalidCertificateError(
-                X509_verify_cert_error_string(verify_err));
-    }
-    X509 *server_cert = SSL_get_peer_certificate(ssl);
-    if (server_cert == nullptr) {
-        logger->warn("ssl: got no certificate");
-        return SslNoCertificateError();
+template <MK_MOCK(SSL_get0_param), MK_MOCK(X509_VERIFY_PARAM_set1_host)>
+Error set_hostname_for_verification(
```
bassosimone: I believe the name of this function is a bit misleading. Plus, the code to really enable peer verification should probably be refactored inside here, for additional clarity. Plus a link to the OpenSSL Wiki is required.

src/libmeasurement_kit/nettests/runnable.cpp
```diff
@@ -184,39 +184,44 @@ void Runnable::geoip_lookup(Callback<> cb) {
...
+            }
             logger->warn("=== END IP_LOOKUP LOGS ===");
             annotations["failure_ip_lookup"] = "true";
         } else {
-            real_probe_ip = mkiplookup_response_get_probe_ip(res.get());
+            real_probe_ip = res.probe_ip;
             logger->debug("=== BEGIN IP_LOOKUP LOGS ===");
-            logger->debug("%s", logs.c_str());
+            for (auto &s : res.logs) {
+                logger->debug("%s", s.c_str());
```
bassosimone: It would be nice, for new code, to have a call that bypasses performing a snprintf() and that sanitizes the logs such that we're sure that non-UTF-8 codepoints are not going to be emitted as events.

include/measurement_kit/internal/report/resubmit.h
```diff
@@ -0,0 +1,95 @@
...
+
+#include <stdint.h>
+#include <stdlib.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif  /* __cplusplus */
+
+/** mk_report_resubmit_request_t is a request to resubmit a report. */
+typedef struct mk_report_resubmit_request mk_report_resubmit_request_t;
```
bassosimone: For testability, it would make lots of sense to be able to specify a custom collector base URL. This would be helpful to direct the requests to a custom collector rather than to the default collector. OTOH, using the bouncer discovery mechanism is clearly conducive to a more thorough integration test in this case, because that is the code that will actually run in the real apps. This leads me to think that one perhaps may also want to set more variables here, like for example, the bouncer base URL. However, since this is a C API, there is no real issue here, because we can always _add_a new function to the API without breaking the API itself. Conclusion: there is no need to create a function for setting a custom collector URL now. This may be considered for later. I will write an integration test that runs a real test, without submitting the results, and then submit the results after the test.

PR #1769, opened 2019-03-21
Release v0.10.0

bassosimone:
This PR contains the missing bits to bless v0.10.0. OONI for mobile will need it soon.


bassosimone:
I'm going to merge with-non-fully-passing-tests because [the only test that fails](https://travis-ci.org/measurement-kit/measurement-kit/jobs/509458219#L1207) is an HTTP website returning `503` rather than `200` in one specific run. I don't want to restart travis just because of this.

PR #1770, opened 2019-03-25
[ci skip] stop mentioning the stable branch everywhere

bassosimone:
In the interest of making the build process easier, this commit
removes the stable branch and only keeps the master branch.

If there's need to make an emergency stable release, then I will
create temporary branches. The workflow in which I also need to
maintain a stable branch has too much toil and too little advantages
known to me. I can easily use tags instead.


bassosimone:
Merging immediately since this is a documentation only PR that I created to more clearly mark a point in history where this specific change has occurred.

Issue #1767, opened 2019-03-20
Stuck travis build


bassosimone:
A recent travis build was stuck with this output:

```
[...]
make[2]: Entering directory '/mk'
PASS: test/common/error
PASS: test/common/delegate
PASS: test/common/error_or
PASS: test/main
PASS: test/common/encoding
PASS: test/common/enable_shared_from_this
PASS: test/common/fapply
PASS: test/common/fcar
PASS: test/common/fcdr
PASS: test/common/fmap
PASS: test/common/freverse
PASS: test/common/libevent_reactor
PASS: test/common/logger
PASS: test/common/locked
PASS: test/common/maybe
PASS: test/common/range
PASS: test/common/scalar
PASS: test/common/fcompose
PASS: test/common/shared_ptr
PASS: test/common/settings
PASS: test/common/unique_ptr
PASS: test/common/version
PASS: test/common/utils
PASS: test/dns/libevent_error
PASS: test/common/continuation
PASS: test/dns/query_class
PASS: test/dns/query_type
PASS: test/common/parallel
PASS: test/ffi/mk_event_create
PASS: test/dns/query
PASS: test/ffi/start
PASS: test/dns/system_resolver
PASS: test/http/headers
PASS: test/http/parse_url
PASS: test/common/every
PASS: test/http/response_parser
PASS: test/dns/ping
PASS: test/mlabns/mlabns
PASS: test/ndt/measure_speed
PASS: test/ndt/messages
PASS: test/ndt/protocol
PASS: test/mkapi/orchestra
PASS: test/ndt/test_c2s
PASS: test/ndt/test_meta
PASS: test/ndt/test_s2c
PASS: test/net/buffer
PASS: test/ffi/web_connectivity_example
PASS: test/net/emitter
PASS: test/net/evbuffer
PASS: test/net/libevent_emitter
PASS: test/net/libssl
PASS: test/net/socks5
PASS: test/net/transport
PASS: test/net/utils
PASS: test/nettests/captive_portal
PASS: test/net/connect
PASS: test/ndt/run
PASS: test/ffi/ndt_example
PASS: test/nettests/dns_injection
PASS: test/nettests/facebook_messenger
PASS: test/nettests/http_header_field_manipulation
PASS: test/http/request
PASS: test/nettests/meek_fronted_requests
PASS: test/nettests/http_invalid_request_line
PASS: test/nettests/dash
PASS: test/common/worker
PASS: test/nettests/utils
[...]

PR #1773, opened 2019-03-27
Release 0.10.1

bassosimone:


bassosimone:
The failed travis build is interesting because related to https://github.com/measurement-kit/measurement-kit/issues/1767#issuecomment-477383591. I'm therefore not going to restart the build but I'm keeping it failed for future reference. The failure is that the `runnable` test enters into some sort of deadlock and never completes.

bassosimone:
I'm going to merge with the failed test without restarting. This is a known issue (see #1767) and I suspect this has more to do with the tests themselves that with MK.

PR #1785, opened 2019-04-24
Fix compilation without deprecated OpenSSL APIs

neheb:
Initialization is deprecated and causes compile errors with OpenSSL 1.1 when
deprecated APIs are disabled.


neheb:
The patch breaks on LibreSSL as it targets the 1.0.1 API and because they decided to be smart and mess with the macro.

bassosimone:
> The patch breaks on LibreSSL as it targets the 1.0.1 API and because they decided to be smart and mess with the macro.

I'm quite sure there is a LibreSSL specific macro that can be used to check whether you're using OpenSSL or instead you are using LibreSSL.

neheb:
Yeah there is. Will update the pull request in a few hours.

neheb:
Updated.

bassosimone:
Thanks for working on this! I'll merge this PR after I've repaired tests. It'll be released along with MK v0.11.0, which is quite close in time.

bassosimone:
Since we're making v0.10.2, it probably makes sense to put this in there.

bassosimone:
CI is failing with this error:

```
+ ./autogen.sh
* Generating include.am
* Fetching CA from cURL website
curl: (28) Operation timed out after 300694 milliseconds with 0 out of 0 bytes received
The command "./.ci/travis/script/travis $BUILD_TYPE" exited with 28.
```

That's a transient network error. Since 4/5 of the checks worked, and that was the `vanilla` build, I am going to merge without bothering with restarting the build.

Thank you!

PR #1794, opened 2019-04-30
collector_client: always use production collector

### LINKED ISSUE #1787: travis: repair broken regress tests
See #1785. I know that `collector-sandbox.ooni.io` has been retired. I know we're using it in several tests. @hellais had no objections to using production services. It seems fine to run tests for client code using production servers. Also, having less servers to worry about is also conducive to free more of our time to focus on measurements and apps.

bassosimone:
As agreed with @hellais, the sandbox collector is gone.

Closes #1787


bassosimone:
Travis is now all green. AppVeyory has serious issues that will be handled in #1796.

Merging now.

PR #1793, opened 2019-04-30
Upgrade all vendored dependencies for v0.10.2

bassosimone:
Each dependency is in its own commit where I explain why I'm upgrading.

Closes #1792

PR #1798, opened 2019-05-02
 Allow users to specify software_{name,version} for reuploading measurements

### LINKED ISSUE #1797: Allow users to specify software_{name,version} for reuploading measurements

bassosimone:
Closes #1797 

To be merged after #1793


src/libmeasurement_kit/collector/collector.cpp
```diff
@@ -395,7 +407,7 @@ static mk_collector_resubmit_response_t *mk_collector_resubmit_(
     new mk_collector_resubmit_response_t
   };
   auto load_result = mk::collector::open_request_from_measurement(
-      request->content, "measurement-kit", MK_VERSION);
+      request->content, request->software_name, request->software_version);
```
hellais: I guess this is the key element.

src/libmeasurement_kit/collector/collector.cpp
```diff
@@ -291,6 +291,8 @@ mk_collector_close_response_t *mk_collector_close(
 
 struct mk_collector_resubmit_request {
   std::string content;
+  std::string software_name = "libmeasurement_kit";
```
hellais: It's probably worth noting somewhere that the default software_name changed from `measurement-kit` to `libmeasurement_kit`. I don't see there being any problems with that, but it's probably useful to document it at least in the changelog.

hellais:
🐳 looks good to me

src/libmeasurement_kit/collector/collector.cpp
bassosimone: I use `libmeasurement_kit` to indicate the library and `measurement-kit` to indicate the binary executable. Here the difference is subtle and is basically telling anyone checking the logs etc that we didn't change the default. It's basically a way to distinguish between the library and the executable that by default is using the library.

bassosimone: (Ah, and I used `measurement-kit` in the previous diff as a placeholder name.)

Issue #1800, opened 2019-05-02
Compile with `-Werror` on travis.


bassosimone:
Judging from #1799, it's not a good idea to tolerate warnings when compiling on Travis.

This would allow warnings to slip through and incidentally not doing that would most likely also have averted issue #1785.

bassosimone:
Done in #1805

PR #1805, opened 2019-05-02
.ci/travis: enforce -Werror for builds

bassosimone:
See #1800

(I expect this to fail initially because of #1798, which is also failing and which is the reason why I started working to make `-Werror` an always on warning.)


bassosimone:
Sigh. The way in which this fails is interesting:

```
[...]
checking whether clang++ supports C++14 features with -std=c++14... no
checking whether clang++ supports C++14 features with +std=c++14... no
checking whether clang++ supports C++14 features with -h std=c++14... no
checking whether clang++ supports C++14 features with -std=c++1y... no
checking whether clang++ supports C++14 features with +std=c++1y... no
checking whether clang++ supports C++14 features with -h std=c++1y... no
configure: error: *** A compiler with support for C++14 language features is required.
The command "./.ci/travis/script/travis $BUILD_TYPE" exited with 1.
```

I did not see this coming (damn!). It seems the correct behaviour is to implement such functionality such that `-Werror` is _only_ applied when compiling MK, not when configuring.

bassosimone:
Success! As in, now [the build fails as expected](https://travis-ci.org/measurement-kit/measurement-kit/jobs/527476375#L706):

```
./include/measurement_kit/internal/vendor/json.hpp:13087:5: error: use of the 'nodiscard' attribute is a C++17 extension [-Werror,-Wc++17-extensions]
    JSON_NODISCARD
```

bassosimone:
Success, as in: we've suppressed the warning with json.hpp. Still very bummer-y that the build fails with AppVeyor, but we're moving towards having also that fixed.

PR #1843, opened 2019-05-28
Release v0.10.4

bassosimone:
This PR contains some commits that are meant to be included in v0.10.4.

### DIFF
--- a/include/measurement_kit/README.md
+++ b/include/measurement_kit/README.md
@@ -185,6 +185,7 @@ The nettest task settings object is a JSON like:
     "no_bouncer": false,
     "no_collector": false,
     "no_file_report": false,
+    "no_geoip": false,
     "port": 1234,
     "probe_ip": "1.2.3.4",
     "probe_asn": "AS30722",
@@ -359,6 +360,9 @@ These are the available options:
 
 - `"no_file_report"`: (boolean) whether to write a report (i.e. measurement
   result) file on disk. By default set to `false`, meaning that we'll try;
+
+- `"no_geoip"`: (boolean) whether to perform GeoIP lookup. By default set to
+  `false`, meaning that we'll try;
 
 - `"probe_asn"`: (string) sets the `probe_asn` to be included into the
   report, thus skipping the ASN resolution;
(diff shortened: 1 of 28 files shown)


bassosimone:
I've created a tag off this branch: https://github.com/measurement-kit/measurement-kit/releases/tag/v0.10.4.

Will soon merge and build all the binaries.

include/measurement_kit/README.md
```diff
@@ -919,7 +946,7 @@ by setting the appropriate settings.
   let probe_cc = "ZZ"
   if (settings.options.probe_cc != "") {
     probe_cc = settings.options.probe_cc
-  } else if (settings.options.geoip_country_path != "") {
+  } else if (!settings.options.no_geoip) {
```
lorenzoPrimi: I like it!

include/measurement_kit/internal/vendor/mkcollector.hpp
```diff
@@ -21,8 +21,9 @@ class Settings {
   /// ca_bundle_path is the path to the CA bundle (required on mobile)
   std::string ca_bundle_path;
 
-  /// timeout is the whole operation timeout (in seconds)
-  int64_t timeout = 30;
+  /// timeout is the whole operation timeout (in seconds). Zero indicatest
+  /// that there actually is no timeout.
+  int64_t timeout = 0;
```
lorenzoPrimi: so we leave the timeout managment entirely to the app?

bassosimone: Yes. In all the cases where a default timeout was a bad idea, I zeroed the timeout. Which means that, if the app sets no timeout, there's no timeout!

lorenzoPrimi: What is the implication of a no timeout? the operation can run forever?

bassosimone: Yes, it's up to the app to set a timeout. I think, if the app does not set one, the user would eventually be bored and kill the app. I'm designing a better scheme in Go where the user would be able to interrupt a pending operation. Theoretically, I could also implement it in C++, but I think it's probably best to focus on Go at this point, as it's definitely easier.

lorenzoPrimi: Ok I'm gonna check that actually there is a timeout everywhere in the app and approve this PR

PR #1854, opened 2019-05-29
README.md: review and make sure it's up to date

### LINKED ISSUE #1853: Make sure the documentation is current
Re-reading the README after #1852, it occurred to me that there were fake news sprinkled here and there, mainly because the status of the code and the relevant repositories changed since last time we updated the README.

Therefore, another diff is in order to make sure the README is more current. As part of this diff, it would also be nice to stop mentioning the Go wrappers as an official stable way of using MK since it's clearly not and since we're not going to invest further effort in it.

bassosimone:
While there, also stop mentioning the Go API, which is not an
official and stable product.

Closes #1853

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,20 +59,22 @@ for Homebrew.
 
 ## Mingw-w64 instructions
 
-For information on how to compile Measurement Kit using the [Mingw-w64](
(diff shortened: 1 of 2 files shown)


bassosimone:
(No need to wait for CI, since this was a documentation-only change.)

PR #1855, opened 2019-06-03
Document WARNING as the default log level

bassosimone:
It has been like this since forever, but it wasn't documented.


bassosimone:
Cancelled all builds. Should have used `[ci skip]`. This is just a documentation PR. Therefore, there is no point in rebuilding and rerunning tests.

Issue #1866, opened 2019-06-18
Make (sure that) MK work(s) during network outages


bassosimone:
This is not about magically sending packets, but rather make sure that we can perform a test without failing even if the internet is having a bad day. The first issue that comes to mind is that we should see what happens if we cannot contact the bouncer, use test helpers etc.

The initial idea is to understand what is the current situation and the we can open more specific issues for the problems identified and decide how to proceed. It may be that some tasks are more easy to be implemented in the Go engine. But, still, let's measure MK first.

bassosimone:
Moved to https://github.com/ooni/probe-engine/issues/814

PR #1879, opened 2019-07-31
More work towards MK v0.10.5

bassosimone:
This PR addresses many outstanding open issues for MK v0.10.5: https://github.com/measurement-kit/measurement-kit/milestone/20.

You should review every commit independently and refer to the related issue.

### DIFF
--- a/include/measurement_kit/README.md
+++ b/include/measurement_kit/README.md
@@ -184,6 +184,7 @@ The nettest task settings object is a JSON like:
     "no_collector": false,
     "no_file_report": false,
     "no_geoip": false,
+    "no_resolver_lookup": false,
     "port": 1234,
     "probe_ip": "1.2.3.4",
     "probe_asn": "AS30722",
@@ -364,6 +365,10 @@ These are the available options:
 
 - `"no_geoip"`: (boolean) whether to perform GeoIP lookup. By default set to
   `false`, meaning that we'll try;
+
+- `"no_resolver_lookup"`: (boolean) whether to lookup the IP address of
+  the resolver. By default `false`, meaning that we'll try. When true we
+  will set the resolver IP address to `127.0.0.1`;
 
 - `"probe_asn"`: (string) sets the `probe_asn` to be included into the
   report, thus skipping the ASN resolution;
@@ -968,9 +973,8 @@ by setting the appropriate settings.
     emitProgress(0.2, "geoip lookup")
   }
 
-  // TODO(bassosimone): take decision wrt null vs. ""
-  let resolver_ip = null
-  {
+  let resolver_ip = "127.0.0.1"
+  if (!settings.options.no_resolver_lookup) {
     let error
     [resolver_ip, error] = lookupResolver(settings)
     if (error) {
--- a/src/libmeasurement_kit/ooni/web_connectivity.cpp
+++ b/src/libmeasurement_kit/ooni/web_connectivity.cpp
@@ -399,16 +399,17 @@ static void control_request(http::Headers headers_to_pass_along,
     http::request(settings, headers, body,
                   [=](Error error, SharedPtr<http::Response> response) {
                       if (!error) {
+                          nlohmann::json doc;
                           try {
-                              (*entry)["control"] =
-                                  nlohmann::json::parse(response->body);
-                              callback(NoError());
-                              return;
+                              doc = nlohmann::json::parse(response->body);
                           } catch (const std::exception &) {
                               (*entry)["control_failure"] = "json_parse_error";
                               callback(JsonProcessingError());
                               return;
                           }
+                          (*entry)["control"] = std::move(doc);
+                          callback(NoError());
+                          return;
                       }
                       (*entry)["control_failure"] = error.reason;
                       callback(error);
--- a/src/libmeasurement_kit/ooni/web_connectivity.cpp
+++ b/src/libmeasurement_kit/ooni/web_connectivity.cpp
@@ -389,6 +389,12 @@ static void control_request(http::Headers headers_to_pass_along,
     request["http_request_headers"] = true_headers;
     std::string body = request.dump();
 
+    /* We want a larger timeout when communicating with the OONI backend
+     * because it may take time for the backend to retrieve a web page and
+     * we don't want to recv() to timeout too soon.
+     *
+     * See <https://github.com/measurement-kit/measurement-kit/issues/1864>. */
+    settings["net/timeout"] = 30.0;
     settings["http/url"] = settings["backend"];
     settings["http/method"] = "POST";
     headers_push_back(headers, "Content-Type", "application/json");
--- a/src/libmeasurement_kit/ooni/web_connectivity.cpp
+++ b/src/libmeasurement_kit/ooni/web_connectivity.cpp
@@ -139,9 +139,14 @@ static void compare_http_requests(SharedPtr<nlohmann::json> entry,
     }
 }
 
-static void compare_dns_queries(SharedPtr<nlohmann::json> entry,
+static void compare_dns_queries(mk::http::Url url, SharedPtr<nlohmann::json> entry,
                                 std::vector<std::string> experiment_addresses,
                                 nlohmann::json control, Settings options) {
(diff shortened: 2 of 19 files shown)


bassosimone:
This branch is currently broken because of NDT issues documented in #1880.

src/libmeasurement_kit/ooni/web_connectivity.cpp
```diff
@@ -583,12 +583,6 @@ void web_connectivity(std::string input, Settings options,
     (*entry)["tcp_connect"] = nlohmann::json::array();
     (*entry)["control"] = nlohmann::json({});
 
-    if (!mk::startswith(input, "http://") &&
-        !mk::startswith(input, "https://")) {
-        // Similarly to ooni-probe also accept a list of endpoints
-        input = "http://" + input;
-    }
-
```
hellais: Let's remember to mention this somewhere in the changelog or in a prominent location so we don't forget.

src/libmeasurement_kit/ooni/web_connectivity.cpp
```diff
@@ -524,8 +530,14 @@ static void experiment_dns_query(
...
-        // Don't perform DNS resolutions if it's an IP address
-        // XXX This means we are not filling the entry
+        // Don't perform DNS resolutions if it's an IP address. The spec of
+        // web connectivity does not say what to do in case we are not doing
+        // the DNS step. A good value to indicate we've not done this step
+        // seems to be `null`. That is also Go-friendly because you can use
+        // a pointer in the structure definition.
+        //
+        // See <https://github.com/ooni/spec/issues/148>.
+        (*entry)["queries"] = nullptr;
```
hellais: Can we initialise this to an empty list instead of a null value?

src/libmeasurement_kit/ooni/web_connectivity.cpp
```diff
@@ -139,9 +139,14 @@ static void compare_http_requests(SharedPtr<nlohmann::json> entry,
     }
 }
 
-static void compare_dns_queries(SharedPtr<nlohmann::json> entry,
+static void compare_dns_queries(mk::http::Url url, SharedPtr<nlohmann::json> entry,
                                 std::vector<std::string> experiment_addresses,
                                 nlohmann::json control, Settings options) {
+    // Short circuit for the case where we're using an IP address.
+    if (mk::net::is_ip_addr(url.address)) {
+        (*entry)["dns_consistency"] = "consistent";
```
hellais: I guess this is a reasonable value, however we should mention it in the spec. I added a note to: https://github.com/ooni/spec/issues/148#issuecomment-520373921

test/ooni/templates.cpp
```diff
@@ -97,10 +97,10 @@ TEST_CASE("dns query template works as expected with system engine") {
                     REQUIRE(answers.is_array());
                     REQUIRE(answers.size() == 2);
                     REQUIRE((answers[0]["ttl"].is_null()));
-                    REQUIRE((answers[0]["hostname"] == "server-nexa.polito.it"));
+                    REQUIRE((answers[0]["hostname"] == "vm-nexa.polito.it"));
```
hellais: Do we maybe want to use some other host that we control for these tests?

include/measurement_kit/README.md
```diff
@@ -365,6 +366,10 @@ These are the available options:
 - `"no_geoip"`: (boolean) whether to perform GeoIP lookup. By default set to
   `false`, meaning that we'll try;
 
+- `"no_resolver_lookup"`: (boolean) whether to lookup the IP address of
+  the resolver. By default `false`, meaning that we'll try. When true we
+  will set the resolver IP address to `127.0.0.1`;
+
```
hellais: Is there a way, when this option is enabled, to pass to MK the IP of the resolver of the client obtained through some other means (similarly to how you can pass `probe_cc`, `probe_ip`, etc.)?

hellais: Also should we maybe set the IP of the resolver to `127.0.0.2` so as to distinguish this scrubbed IP from the scrubbed IP of the probe (this can be useful to do a search and replace if there is lack of a proper function to pass it into MK).

test/ooni/templates.cpp
bassosimone: Yes, using Nexa is becoming unreliable.

include/measurement_kit/README.md
bassosimone: No, there is not. The reason why there is not is that the approach that I am following for probe-engine is that we disable everything and we unconditionally overwrite such fields with the values we know to be good at the end of a test.

src/libmeasurement_kit/ooni/web_connectivity.cpp
bassosimone: https://github.com/measurement-kit/measurement-kit/issues/1882

src/libmeasurement_kit/ooni/web_connectivity.cpp
bassosimone: We can. Why do you think an empty list is better than a `null` value? My reasoning for using `null` is that it indicates the field has not / could not been set. This fits more nicely with the semantic of an optional field than leaving the list empty, I believe.

src/libmeasurement_kit/ooni/web_connectivity.cpp
bassosimone: Thanks!

src/libmeasurement_kit/ooni/web_connectivity.cpp
hellais: I was thinking of it from the perspective of a consumer of the data as it's easier to have it be of the same type than having to check for null values.

I see your point though and what you propose is possibly more consistent with the approach we currently following WRT to data formats.

I am now convinced that using `null` here is the right thing to do.

include/measurement_kit/README.md
hellais: Ok I see. Do you see some value in using a different magic number for the resolver_ip in this case then?

bassosimone: Yes. What could be a better value to use? Can we leave it empty?

src/libmeasurement_kit/ooni/web_connectivity.cpp
bassosimone: Awesome, so I don't think I need to do anything special here because that's what we already have written in https://github.com/measurement-kit/measurement-kit/issues/1866

include/measurement_kit/README.md
bassosimone: I am going to leave it empty.

bassosimone:
So, today I'll proceed with making the CI pass and merge.

src/libmeasurement_kit/ooni/web_connectivity.cpp
bassosimone: Well, except perhaps opening a PR to address https://github.com/ooni/spec/issues/148

include/measurement_kit/README.md
bassosimone: Well, actually I don't know. After this PR has landed, I will address this issue as one of the few remaining issues before blessing v0.10.5. See https://github.com/measurement-kit/measurement-kit/issues/1884.

test/ooni/templates.cpp
bassosimone: Created issue at https://github.com/measurement-kit/measurement-kit/issues/1885

bassosimone:
# TODO list before merging

- [ ] rewrite history to consolidate changes to fix Travis and AppVeyor into two commits (it would be nice to rebase on top of ec26c0f, to avoid messing up too much with history)

- [ ] create PR addressing https://github.com/ooni/spec/issues/148

bassosimone:
I've rewritten the history consolidating all the CI attempts into a single commit. See https://github.com/measurement-kit/measurement-kit/compare/bad85334dda4019309f978849dc1f6f52f21546e..2236e3b6979e81ed1c372eac7ba3e2cc59f7fa82.

bassosimone:
Merging! (It's all green but GH hasn't noticed that all Travis have completed, weirdly.)

Issue #1877, opened 2019-07-27
JNI DETECTED ERROR IN APPLICATION: input is not valid Modified UTF-8: illegal start byte 0xf0


bassosimone:
```
[...] A/art: art/runtime/check_jni.cc:65]
JNI DETECTED ERROR IN APPLICATION: input is not valid Modified UTF-8:
illegal start byte 0xf0

[...] A/art: art/runtime/check_jni.cc:65]
string: '{"key":"measurement","value":{"idx":0,"json_str":"{\"annotations\":{\"engine_name\":\"libmeasurement_kit\",\"engine_version\":\"0.10.4\",\"engine_version_full\":\"v0.10.4\",\"flavor\":\"fdroid\",\"network_type\":\"mobile\",\"platform\":\"android\"},\"data_format_version\":\"0.2.0\",\"id\":\"6d2f6c96-561b-4607-bc94-a46f72081077\",\"input\":\"https://www.fanfiction.net/\",\"input_hashes\":[],\"measurement_start_time\":\"2019-07-27 08:20:02\",\"options\":[],\"probe_asn\":\"AS1267\",\"probe_cc\":\"IT\",\"probe_city\":null,\"probe_ip\":\"127.0.0.1\",\"report_id\":\"\",\"software_name\":\"ooniprobe-android\",\"software_version\":\"2.1.0\",\"test_helpers\":{\"backend\":{\"address\":\"https://b.web-connectivity.th.ooni.io:443\",\"type\":\"https\"}},\"test_keys\":{\"accessible\":true,\"agent\":\"redirect\",\"blocking\":false,\"body_length_match\":true,\"body_proportion\":1.0,\"client_resolver\":\"173.194.170.4\",\"control\":{\"dns\":{\"addrs\":[\"fanfiction.net\",\"173.205.18

[...] A/art: art/runtime/check_jni.cc:65]
in call to NewStringUTF

[...] A/art: art/runtime/check_jni.cc:65]
from java.lang.String io.ooni.mk.MKEvent.Serialize(long)
```

This happens reliably by forcing `https://www.fanfiction.net/` as input to `web_connectivity`.

bassosimone:
Given what I discovered so far, it seems that the most reasonable thing to do is to extract the offending JSON encoded as base64, and then understand why it is offending. Luckily, we have a reproducible test case that allows us to dig in. (I am not sure whether this matter but my test case is using Android 5.0 Lollipop.)

[...]

PR #1889, opened 2019-08-26
engine, ffi: emit UTF-7 without U+0000

bassosimone:
JNI's `NewStringUTF` requires modified UTF-8. That is, actually, incompatible UTF-8. In most cases UTF-8 streams are accepted by JNI. See #1877 for exceptions to this rule.

Address this issues with two changes:

1. export `engine::Task` as a public, internal API, such that we can, as a last resort mechanism, modify OONI code interfacing with MK to completely by pass C strings and C code. From C++ it seems easier to get an array of bytes and a size, and to convert that to a Java `byteArray` that then can be converted to a string using native Java code and an encoding. This approach has been documented, e.g., by [Scott Banachowski](http://banachowski.com/deprogramming/2012/02/working-around-jni-utf-8-strings/).

2. modify the way in which we returns serialised JSONs such that we only use UTF-7 bytes and make sure that `U+0000` is properly encoded with unit tests. This approach should be [compliant with what JNI code is documented to always be able to accept](https://developer.android.com/training/articles/perf-jni#utf-8-and-utf-16-strings).

The engine and ffi modules are the two only places where we need (1) to exercise care in the way in which we encode strings and (2) to have more flexibility. The rest of the API is now internal, C++ header only bits that we can easily convert to follow a `byteArray` compatible approach.

This set of changes will land as MK v0.10.6. I will use this release, as opposed to v0.10.5, for Android because I'd like to have this annoying JNI issue fixed as soon as possible.

### DIFF
--- a/include/measurement_kit/internal/engine/task.hpp
+++ b/include/measurement_kit/internal/engine/task.hpp
@@ -1,0 +1,74 @@
+// Part of Measurement Kit <https://measurement-kit.github.io/>.
+// Measurement Kit is free software under the BSD license. See AUTHORS
+// and LICENSE for more information on the copying conditions.
+#ifndef INCLUDE_MEASUREMENT_KIT_INTERNAL_ENGINE_TASK_HPP
+#define INCLUDE_MEASUREMENT_KIT_INTERNAL_ENGINE_TASK_HPP
+
+#include <memory>
+
+#include <measurement_kit/common/nlohmann/json.hpp>
+
+namespace mk {
+namespace engine {
+
+/// TaskImpl is the opaque implementation of Task.
+class TaskImpl;
+
+/// Task is a task that Measurement Kit can run. You create a task with Task()
+/// by passing it the desired settings as a nlohmann::json. The minimal settings
+/// JSON must include the "name" key indicating the task name.
+///
+/// Creating a Task also creates the thread that will run it. Altough you can
+/// construct more than one Task at a time, Measurement Kit will make sure that
+/// tasks do not run concurrently, even though there is no guarantee that they
+/// will actually run sequentially.
+///
+/// A Task will emit events while running, which you can retrieve using the
+/// wait_for_next_event() call, which blocks until next event occurs. You can
+/// configure a Task to disable some or all events.
+///
+/// To know whether a task has finished running, use is_done(). The method will
+/// return true when the task thread has exited and there are no unread events
+/// in the queue drained by wait_for_next_event().
+///
+/// You can also interrupt a running task using interrupt().
+///
+/// The destructor, ~Task(), will join on the Task thread. That is, it will
+/// wait for the Task to complete before destroying all the resources.
+class Task {
+  public:
+    /// Task creates and starts a Task configured according to settings.
+    explicit Task(nlohmann::json &&settings);
+
+    /// wait_for_next_event blocks until the next event occurs. When the
+    /// task is terminated, it returns the "task_terminated" dummy event.
+    nlohmann::json wait_for_next_event();
+
+    /// emit emits a specific event by appending it to the queue that
+    /// is drained by user code via wait_for_next_event.
+    void emit(nlohmann::json event);
+
+    /// is_done returns true when the task has finished running and there
+    /// are no unread events in wait_for_next_event's queue.
+    bool is_done() const;
+
+    /// interrupt forces the Task to stop running ASAP.
+    void interrupt();
+
+    /// ~Task waits for the task to finish and deallocates resources.
+    ~Task();
+
+    // Implementation note: this class is _explicitly_ non copyable and non
+    // movable so we don't have to worry about pimpl's validity.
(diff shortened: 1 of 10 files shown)


include/measurement_kit/internal/engine/task.hpp
```diff
@@ -0,0 +1,74 @@
+// Part of Measurement Kit <https://measurement-kit.github.io/>.
+// Measurement Kit is free software under the BSD license. See AUTHORS
+// and LICENSE for more information on the copying conditions.
+#ifndef INCLUDE_MEASUREMENT_KIT_INTERNAL_ENGINE_TASK_HPP
+#define INCLUDE_MEASUREMENT_KIT_INTERNAL_ENGINE_TASK_HPP
```
bassosimone: This code has been extracted verbatim from `src/libmeasurement_kit/engine/task.hpp`.

src/libmeasurement_kit/engine/task.hpp
```diff
@@ -4,81 +4,21 @@
 #ifndef SRC_LIBMEASUREMENT_KIT_ENGINE_TASK_HPP
 #define SRC_LIBMEASUREMENT_KIT_ENGINE_TASK_HPP
```
bassosimone: The code removed from here was actually moved to `include/measurement_kit/internal/engine/task.hpp`

src/libmeasurement_kit/ffi.cpp
```diff
@@ -21,7 +21,16 @@
...
+        // The following encodes the JSON in the most compact representation
+        // (i.e. indent=-1, indent_char=' '), and make sure that we emit in
+        // ouput only ASCII characters (i.e. UTF-7 rather than UTF-8). We're
+        // doing this to increase the likelihood that the code that will be
+        // parsing this JSON will accept it. For example, Java's modified
+        // UTF-8 SHOULD be able to deal with UTF-7 where `U+0000` is properly
+        // encoded but MAY NOT deal correctly with an UTF-8 stream.
+        //
+        // See https://developer.android.com/training/articles/perf-jni#utf-8-and-utf-16-strings
+        ev = json.dump(-1, ' ', true);
```
bassosimone: This fix should guarantee that we emit UTF-7 rather than UTF-8, as explained in the comment.

test/dns/query.cpp
```diff
@@ -389,26 +389,26 @@ TEST_CASE("The libevent resolver works as expected") {
     });
 
     reactor->run_with_initial_event([=]() {
```
bassosimone: Changes in this file are because a host that we used to test has changed IPv6 address.

test/ffi/mk_event_create.cpp
```diff
@@ -42,3 +42,32 @@ TEST_CASE("mk_event_create_() deals with invalid JSON event key") {
   REQUIRE(ev != nullptr);
   REQUIRE(mk_event_serialization(ev.get()) != nullptr);
 }
+
```
bassosimone: These changes are to make sure that we correctly deal with U+0000 and UTF-8 characters.

test/ffi/mk_event_create.cpp
```diff
@@ -42,3 +42,32 @@ TEST_CASE("mk_event_create_() deals with invalid JSON event key") {
...
+  auto seriop = mk_event_serialization(ev.get());
+  REQUIRE(seriop != nullptr);
+  auto serios = std::string{seriop};
+  REQUIRE(serios == R"({"sbs\u0000sbs":"sbs\u0000sbs"})");
+}
+
+TEST_CASE("mk_event_create_() correctly guarantees UTF-7 output") {
+  // (Cultural note: despite being born in Sanremo I'm actually opposed to
+  // the festival, which I consider low quality music, and I fancy much more
+  // Rock in the Casbah, a summer festival in Sanremo's casbah.)
```
bassosimone: @lorenzoPrimi you probably have some opinion on this comment too :-P

lorenzoPrimi: The quality of the festival went up drastically in the last 2/3 years. More rock and fresh songs.

test/ffi/mk_event_create.cpp
```diff
@@ -42,3 +42,32 @@ TEST_CASE("mk_event_create_() deals with invalid JSON event key") {
...
+  const std::string sanremo{"Perché Sanremo è Sanremo"};
```
lorenzoPrimi: Don't you want to test a more weird char than è or é? Or is not necessary?

bassosimone: Good question. I've updated the test to include a character above U+FFFF, [which is another example of character that Modified UTF-8 shouldn't be able to deal with](https://stackoverflow.com/a/7921064/4354461).

test/ffi/mk_event_create.cpp
bassosimone: https://www.dailymotion.com/video/x8ft2p

PR #1912, opened 2020-02-14
autoapi: add backend to the valid options

bassosimone:
Closes https://github.com/ooni/probe-engine/issues/315


bassosimone:
It turns out `backend` was already present?!

PR #1932, opened 2020-09-17
all: use ams-pg.ooni.org as testing probe service

bassosimone:
As indicated at https://github.com/ooni/sysadmin/pull/452, we wanna use ams-pg.ooni.org as our testing probe service for some time while ps-test.ooni.io is down and while we are working to refactor APIs at ams-pg.ooni.org.


bassosimone:
Actually the reference issue is https://github.com/ooni/backend/issues/446

PR #1936, opened 2021-01-19
MK: bless final release and archive the repository

bassosimone:


bassosimone:
The only change since 0.10.13 is that we updated Catch2, a library we use for testing. It feels needless to bless a new release now. Let us wait some more time and see if more dependencies updates arise.

